PR #7, opened 2017-10-05
Refactor ack frame parsing, fix ack parsing, add frame parsing for logging code

igorlord:
Refactor ack frame parsing, fix ack parsing, add frame parsing for logging code.

Signed-off-by: Igor Lubashev <ilubashe@akamai.com>

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -344,9 +344,9 @@ int picoquic_decode_stream_reset_frame(picoquic_cnx_t * cnx, uint8_t * bytes,
  */
 static const int picoquic_offset_length_code[4] = { 0, 2, 4, 8 };
 
-static int picoquic_parse_stream_header(picoquic_cnx_t * cnx, const uint8_t * bytes, size_t bytes_max,
-                                        uint32_t * stream_id, uint64_t * offset, size_t * data_length,
-                                        size_t * consumed)
+int picoquic_parse_stream_header(const uint8_t * bytes, size_t bytes_max,
+								 uint32_t * stream_id, uint64_t * offset, size_t * data_length,
+								 size_t * consumed)
 {
     int     ret = 0;
     size_t  byte_index = 1;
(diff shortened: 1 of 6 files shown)


picoquic/frames.c
```diff
@@ -1179,167 +1195,83 @@ static picoquic_packet * picoquic_process_ack_range(
...
-				}
+			/* Skip the gap */
+			uint64_t block_to_block = range + bytes[byte_index++];
+			if (largest < block_to_block)
+			{
+				DBG_PRINTF("ack gap error: largest=%" PRIx64 ", range=%" PRIx64 ", gap=%" PRIu64,
+						   largest, range, block_to_block-range);
+				ret = -1;
+				break;
 			}
```
huitema: I am not sure about that code. The gaps are normally in decreasing order. The first range is "largest-range...largest", then second largest = largest - first range - first gap, etc. I have the impression that your code is doing the opposite.

picoquic/frames.c
```diff
@@ -940,108 +1020,44 @@ static picoquic_packet * picoquic_update_rtt(picoquic_cnx_t * cnx, uint64_t larg
...
-		sack_count = 0;
+		picoquic_sack_item_t * sack          = &cnx->first_sack_item;
+		picoquic_sack_item_t * previous_sack = NULL;
+		unsigned               sack_count    = 0;
 
 		do {
+			/* TODO: Clarify logic here and simplify */
 			if (sack->start_of_sack_range < largest &&
 				sack->end_of_sack_range < largest)
 			{
```
huitema: I think you changed the logic there. Largest is the "largest seen". It does not mean that all packets before that were acked. The acknowledged ranges are "before largest". An ACK is acknowledged if it is present in one of these ranges, not just if it's sequence number is below largest.

igorlord: Are you referring to this the code below this comment?  I specifically did NOT change that code -- it is the original.  I added the comment, *because* it seemed wrong.

picoquic/frames.c
igorlord: I think my code is fine.  It is doing exactly what you described:

            /* Skip the gap */
            uint64_t block_to_block = range + bytes[byte_index++];
            ...
            largest -= block_to_block;
 
So the next largest is largest - (range+gap)

picoquic/frames.c
huitema: You are right, I should have looked at the whole set.

picoquic/frames.c
huitema: Sorry, I was reading this too quickly. Yes, you are right.

PR #28, opened 2017-11-08
Connection ID modification test/demo

dbavatar:
This feature is intended to show how connection ID's could be changed.
For example: to encode instance specific information in the server
connection id for routing purposes.


dbavatar:
> I can certainly check that in, it will not break anything in our demo or in interop. But I am a bit skeptical. You want to have clear text bits identifying the load balancing context. Doesn't that enable some directed denial of service attacks?

Effectively the idea is that bits of the connection ID could be stolen for load balancing, without enforcing an algorithm for the contents of those bits - they do not need to be cleartext, as you can set some cryptographically hashed value making the pattern indistinguishable from random. That design falls onto the load balancer. We think it's important for people to understand this, as well as allowing it to be tested.

huitema:
My concern is that there are two ways to achieve what you desire. You describe the simple "mask" solution, for separating the bits seen by the load balancer and the bits set by the individual server. The downside is that for a given server the "load balancer" bits will be constant. They can then be observed by third parties, and might be used in attacks.

An alternative approach would be to use some 64 bit encryption algorithm, maybe Blowfish, such that connection_ID = encrypt(server_ID|nonce). This would require that the load balancer decrypts the connection ID to find the server_ID. It would effectively hide the connection_id to server relation from third parties.

From a code point of view, I would not want to program the specific encryption algorithm in QUIC. What I would do instead is program a callback. If the callback is set, the server would call it instead of calling crypto-random to set the connection ID. It would also pas a context parameter to the call back.

The callback design is more flexible than just setting a mask. Indeed, it would be possible to implement your mask solution with a callback -- pass struct { mask, val } as context parameter. But it would allow for experimentation with more complex connection ID generation as well.

Of course for me all that is theoretical. You are the ones actually developing load balancers. So if the simple solution is good enough for you, I am happy to check it in. But I want to confirm that this is OK.

huitema:
My concern is that there are two ways to achieve what you desire. You describe the simple "mask" solution, for separating the bits seen by the load balancer and the bits set by the individual server. The downside is that for a given server the "load balancer" bits will be constant. They can then be observed by third parties, and might be used in attacks.

An alternative approach would be to use some 64 bit encryption algorithm, maybe Blowfish, such that connection_ID = encrypt(server_ID|nonce). This would require that the load balancer decrypts the connection ID to find the server_ID. It would effectively hide the connection_id to server relation from third parties.

From a code point of view, I would not want to program the specific encryption algorithm in QUIC. What I would do instead is program a callback. If the callback is set, the server would call it instead of calling crypto-random to set the connection ID. It would also pas a context parameter to the call back.

The callback design is more flexible than just setting a mask. Indeed, it would be possible to implement your mask solution with a callback -- pass struct { mask, val } as context parameter. But it would allow for experimentation with more complex connection ID generation as well.

Of course for me all that is theoretical. You are the ones actually developing load balancers. So if the simple solution is good enough for you, I am happy to check it in. But I want to confirm that this is OK.

dbavatar:
The algorithm we're toying with internally is effectively a callback, but it's simple for us to have a 2 liner patch versus a callback mechanism - we're using picoquic at the moment as a reference implementation and test traffic provider for our load balancer.

I agree callback is superior, although we were struggling with how to share this meaningfully with picoquic. I could implement this PR to do exactly the same thing, except the simple mask/value demo would be the default callback registered from picoquicdemo. Does that sound better?

huitema:
It does sound better, a bit more general purpose. Can you do that easily?

huitema:
One more thing. Your PR assumes that the stateless reset secret is set per server. Is that OK in your deployment?

dbavatar:
I believe you would need to have stateless reset secret will have to be shared per load balancing group, is that your concern? So I should add the ability to set that from the command line as well.

PR #32, opened 2017-11-15
Allow linux to use a listen socket per protocol

dbavatar:
This allows win32 and linux to work in the same way, fixing ip addr logging


huitema:
Sorry, I am traveling and my laptop is broken. I will only be able to review this PR on Wednesday 22.

PR #34, opened 2017-11-21
Force server cnx_id xmit when modifying cnx_id

dbavatar:
Modifying cnx_id implies you are doing some sort of load balancing based on
cnx_id, therefore a server should always xmit cnx_id such that you can find
that packet's origin later, for example if it comes back in an icmp error.

### DIFF
--- a/picoquic/picoquic_internal.h
+++ b/picoquic/picoquic_internal.h
@@ -77,7 +77,8 @@ extern "C" {
 	 */
 	typedef enum {
 		picoquic_context_server = 1,
-        picoquic_context_check_cookie = 2
+		picoquic_context_check_cookie = 2,
+		picoquic_context_unconditional_cnx_id = 4
 	} picoquic_context_flags;
 
 	/*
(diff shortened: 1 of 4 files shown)


huitema:
OK. I will do a small change to make sure that we don't set the callback if the -i option is not used.

huitema:
Actually, I had to also fix a bug in your PR. You had a test written as:

							if (cnx->quic->flags & picoquic_context_unconditional_cnx_id == 0)
								cnx->remote_parameters.omit_connection_id = 1;

The problem there is the precedence between the "&" and "==" operators. With the Windows compiler, the "==" is evaluated first, and your condition was always true. A number of unit tests were failing. I had to change it to:

							if ((cnx->quic->flags & picoquic_context_unconditional_cnx_id) == 0)
								cnx->remote_parameters.omit_connection_id = 1;

Also, I had to fix the transport parameter unit test. But that's my fault. On the other hand, you should have run the unit tests before pushing the fix...

PR #62, opened 2018-01-27
Adds support for using a quic context as server/client

bkchr:
Before these changes, a context could either be a serer or a client.
With these changes, a context can be both.

All tests are working.

### DIFF
--- a/picoquic/packet.c
+++ b/picoquic/packet.c
@@ -43,7 +43,6 @@ int picoquic_parse_packet_header(
     uint8_t * bytes,
     uint32_t length,
     struct sockaddr * addr_from,
-    int to_server,
     picoquic_packet_header * ph,
     picoquic_cnx_t ** pcnx)
 {
@@ -71,7 +70,7 @@ int picoquic_parse_packet_header(
                 ph->pnmask = 0;
                 ph->version_index = -1;
 
-                if (to_server == 0 && *pcnx == NULL)
+                if (*pcnx == NULL)
                 {
                     *pcnx = picoquic_cnx_by_id(quic, ph->cnx_id);
 
@@ -93,6 +92,23 @@ int picoquic_parse_packet_header(
                 }
                 else
                 {
+                    /* Is the context found by using the `addr_from`? */
+                    char context_by_addr = 0;
+                    /* Retrieve the connection context */
+                    if (*pcnx == NULL)
+                    {
+                        *pcnx = picoquic_cnx_by_id(quic, ph->cnx_id);
+
+                        /* TODO: something for the case of client initial, e.g. source IP + initial CNX_ID */
+                        if (*pcnx == NULL)
+                        {
+                            *pcnx = picoquic_cnx_by_net(quic, addr_from);
+                            context_by_addr = 1;
+                        }
+                    }
+
+                    char is_client = *pcnx && (*pcnx)->client_mode;
+
                     /* If the version is supported now, the format field in the version table
                      * describes the encoding. */
                     switch (picoquic_supported_versions[ph->version_index].version_header_encoding)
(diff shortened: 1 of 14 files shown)


picoquic/packet.c
```diff
@@ -93,6 +92,23 @@ int picoquic_parse_packet_header(
...
+
+                        /* TODO: something for the case of client initial, e.g. source IP + initial CNX_ID */
+                        if (*pcnx == NULL)
+                        {
+                            *pcnx = picoquic_cnx_by_net(quic, addr_from);
+                            context_by_addr = 1;
+                        }
+                    }
+
+                    char is_client = *pcnx && (*pcnx)->client_mode;
```
huitema: I always try to place declarations at the beginning of the block. Is it because I am too old school? That used to be mandated by the old C compilers...

picoquic/packet.c
```diff
@@ -109,7 +125,7 @@ int picoquic_parse_packet_header(
                             ph->ptype = picoquic_packet_server_stateless;
                             break;
                         case 0xFD:
-                            ph->ptype = (to_server == 0) ? picoquic_packet_server_cleartext : picoquic_packet_client_cleartext;
+                            ph->ptype = is_client ? picoquic_packet_server_cleartext : picoquic_packet_client_cleartext;
                             break;
```
huitema: The two types "server clear text" and "client clear text" were defined in the early drafts. They have been unified in a single "handshake" type in draft 08. We might be able to simplify the code by just handling a single packet type instead on insisting on 2. I opened issue #63 for that.

picoquic/quicctx.c
```diff
@@ -150,11 +150,7 @@ picoquic_quic_t * picoquic_create(uint32_t nb_connections,
 			quic->flags |= picoquic_context_unconditional_cnx_id;
 		}
 
-        if (cert_file_name != NULL)
-        {
-            quic->flags |= picoquic_context_server;
-        }
-        else if (ticket_file_name != NULL)
+        if (ticket_file_name != NULL && cert_file_name == NULL)
         {
```
huitema: Not sure that this is the right test. The beauty of this whole PR is that it allows a single context to be used for accepting connections as server and establishing connections as client. If the context is created with a non NULL cert file name, that implies it will be able to be used as server. But if it could also be used as client, then it will need a ticket store in order to resume connections and use 0-RTT. Isn't that what the programmer indicates by setting the ticket_file_name to a non NULL value?

picoquic/quicctx.c
```diff
@@ -504,16 +501,16 @@ picoquic_cnx_t * picoquic_create_cnx(picoquic_quic_t * quic,
             sizeof(struct sockaddr_in) : sizeof(struct sockaddr_in6));
         memcpy(&cnx->peer_addr, addr, cnx->peer_addr_len);
 
-		picoquic_init_transport_parameters(&cnx->local_parameters, quic->flags&picoquic_context_server);
+        picoquic_init_transport_parameters(&cnx->local_parameters, !cnx->client_mode);
         /* Special provision for test -- create a deliberate transport parameters error */
```
huitema: If we are going to have a client mode flag, then maybe we should update the init transport parameter implementation. In fact, we may be able to just use a single "connection context" parameter for that function. There are probably other examples like that. Tracking that in issue #64.

picoquic/packet.c
bkchr: Yeah the old days ^^ I will try to do so in future commits :)

picoquic/packet.c
bkchr: Ahh that is nice :)

picoquic/quicctx.c
bkchr: Yeah I was not quite sure on what to test here.

PR #66, opened 2018-01-27
Formats source code with clang-format and adds clangformat target

bkchr:
Run `make clangformat` to format all the source code again.

I used the "Webkit" style to format the source code, because it seems to fit best the current formatting. I can also change that, if you want. The following styles are supported: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#configurable-format-style-options

If none of these styles is okay for you, we could also create a custom one.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,3 +95,15 @@ TARGET_LINK_LIBRARIES(picoquic_ct picoquic-core
 )
 
 SET(TEST_EXES picoquic_ct)
+
+# get all project files for formatting
+file(GLOB_RECURSE CLANG_FORMAT_SOURCE_FILES *.c *.h)
+
+# Adds clangformat as target that formats all source files
+add_custom_target(
+    clangformat
+    COMMAND clang-format
+    -style=Webkit
+    -i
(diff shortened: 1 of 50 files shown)


CMakeLists.txt
```diff
@@ -95,3 +95,15 @@ TARGET_LINK_LIBRARIES(picoquic_ct picoquic-core
...
+file(GLOB_RECURSE CLANG_FORMAT_SOURCE_FILES *.c *.h)
+
+# Adds clangformat as target that formats all source files
+add_custom_target(
+    clangformat
+    COMMAND clang-format
+    -style=Webkit
+    -i
+    ${CLANG_FORMAT_SOURCE_FILES}
+)
```
huitema: That's cool. So "make clangformat" would redo the indentations, correct?

bkchr: If you have `clang-format` installed, it will always format the source code in the Webkit style. So, you can write as you want and before committing, you just need to execute that once and everything is formatted. :)

bkchr:
Yeah, braces are important ;) Would be nice it the compiler could warn on that, but I did not found anything useful.

PR #96, opened 2018-02-27
Adds support for keep alive

bkchr:


huitema:
I am looking at the code and it looks OK, but I am concerned with the reliance on the PONG frame. This frame is stated for removal from the spec in the next draft, and is already removed from the editor's copy. It was introduced as a mechanism to prove reachability after connection migration, but it is being replaced by a dedicated set of challenge/response frames.

The key difference between Ping/Pong and Challenge/Response is that the Pong frame is explicitly acknowledged, while challenge and response do not generate ACK -- the response is the implicit ACK of the challenge. This means 3 messages for Ping/Pong/ACK, and just 2 for Challenge/Response. 

How hard would it be to eliminate the dependency on the Pong response? For example, could you just send a Ping with zero data, which does not elicit a Pong in the current code. Add some padding if you think that's useful. If the Ping is acked, the connection is alive. If it is not, retransmission will kick in, and if there is no response after 4 retransmissions the connection will be terminated. That way, you will  not need a keep alive call back. If applications do need a callback, they can implement the functionality at the application level.

We also need to understand the difference between the keep alive timer and the "idle timeout" transport parameter. The keep alive will not be effective if the interval is longer than the remote idle timeout, because the peer will close the connection. In fact, the interval should also be shorter than the local timeout. Of course, the current code in picoquic does not heed the parameters' values. I entered issue #97 to track that.

bkchr:
My code does not rely on the `PONG` frames, I just use the `PONG` frames in the added test, but changing that should not be that complicated :)

Yeah that is right, the implicit `ACK` should be sufficient. I first tried to create the `PING` frame without any data, but that did not generate any `PONG` frame. We could easily revert back to `PING` frames without any data. 

Yeah keep alive will not be effective, if the idle timeout is smaller. But I think, the application layer should know what it is doing. The spec also says, that the application layer decides on which side the keep alive is handled and then it should also be possible for the application layer to deliver the right keep alive interval. 

So what should I do? Should I stop relying on `PONG` in the test and send empty `PING` frames?

picoquic/sender.c
```diff
@@ -1469,9 +1474,23 @@ int picoquic_prepare_packet_ready(picoquic_cnx_t* cnx, picoquic_packet* packet,
...
+                    ret = picoquic_prepare_misc_frame(cnx->keep_alive->frame, &bytes[length],
+                                                      cnx->send_mtu - checksum_overhead - length, &data_bytes);
+                    if (ret == 0) {
+                        length += data_bytes;
+                        cnx->keep_alive->next_frame_time = current_time + cnx->keep_alive->interval;
+                    } else {
+                        if (ret == PICOQUIC_ERROR_FRAME_BUFFER_TOO_SMALL) {
+                            ret = 0;
+                        }
+                    }
```
huitema: This could be simplified if you go for the "simple ping" approach. You would just need something like:
~~~
if (cnx->keep_alive != NULL && cnx->keep_alive->next_frame_time <= current_time && length  + checksum_overhead + 1< cnx->send_mtu) {
    bytes[length++] = picoquic_frame_type_ping;
    bytes[length++] = 0;
    cnx->keep_alive->next_frame_time = current_time + cnx->keep_alive->interval;
}
~~~
No need to involve the "misc frame type" path -- it is really meant for frames directly formatted by the application.

bkchr: Yeah, got catch!

huitema:
Just looked at the code again. Switching to simple ping does make the code quite simpler. No need to worry about misc frames, or to store a copy of the keep alive frame.

There is another issue. In principle, there is no need to send a keep alive if any other packet was sent that required an acknowledgement. So the keep-alive timer should not just be reset when sending an actual keep alive, but also whenever a STREAM frame is sent, or in fact any kind of control frame that requires an acknowledgement, e.g. MAX_DATA, MAX_STREAM or MAX_STREAM_DATA.

huitema:
Then there is a little matter of philosophy regarding who sends what. Should the keep alive timer be reset if the peer *receives* a data frame?

There already is a variable that tracks activity: cnx->latest_progress_time. I wonder whether we need a separate keep-alive timer, or whether we could just test for something like "cnx->latest_progress_time + cnx->keep_alive_timer", and consider that keep alive is enabled if keep_alive_timer != 0.

huitema:
The test could just be to start a connection, enable keep alive on one of the ends of the connection, and verify that the the connection stays up for a long virtual time. And conversely, do the same test without enabling keep alive and verify that the connection goes down.

bkchr:
@huitema I think that I worked in all your comments. I hope that I did not miss anything :) I hope that it is now mergable.

picoquic/frames.c
```diff
@@ -1795,11 +1795,24 @@ int picoquic_decode_max_stream_id_frame(picoquic_cnx_t* cnx, uint8_t* bytes,
...
+int picoquic_prepare_first_misc_frame(picoquic_cnx_t* cnx, uint8_t* bytes,
+                                      size_t bytes_max, size_t* consumed)
+{
+    int ret = picoquic_prepare_misc_frame(cnx->first_misc_frame, bytes, bytes_max, consumed);
+
+    if (ret == 0) {
+        picoquic_misc_frame_header_t* misc_frame = cnx->first_misc_frame;
+        cnx->first_misc_frame = misc_frame->next_misc_frame;
+        free(misc_frame);
+    }
```
huitema: Maybe we need to start write documentation for these API. Starting on the Wiki would be nice. I am not sure that I understand the difference between picoquic_prepare_first_misc_frame() and picoquic_prepare_misc_frame(). The original function was supposed to take the first frame in the queue, copy it to the output buffer, and then remove it from the queue and delete the corresponding buffer.

This matches the behavior of the posting function, which takes the input from the application, allocates a buffer, copies the frame in the buffer, and chains the buffer to the misc queue.

picoquic/frames.c
```diff
@@ -1808,8 +1821,6 @@ int picoquic_prepare_misc_frame(picoquic_cnx_t* cnx, uint8_t* bytes,
         uint8_t* frame = ((uint8_t*)misc_frame) + sizeof(picoquic_misc_frame_header_t);
         memcpy(bytes, frame, misc_frame->length);
         *consumed = misc_frame->length;
-        cnx->first_misc_frame = misc_frame->next_misc_frame;
-        free(misc_frame);
     }
```
huitema: Removing the free(frame) statement will create a memory leak if we keep the definition of the original API. Do we really need this?

picoquic/quicctx.c
```diff
@@ -1188,3 +1199,17 @@ void picoquic_set_congestion_algorithm(picoquic_cnx_t* cnx, picoquic_congestion_
         cnx->congestion_alg->alg_init(cnx);
     }
 }
+
+void picoquic_enable_keep_alive(picoquic_cnx_t* cnx, uint64_t interval)
+{
+    if (interval == 0) {
+        cnx->keep_alive_interval = PICOQUIC_MICROSEC_SILENCE_MAX / 2;
+    } else {
```
huitema: OK. Simple enough.

picoquic/sender.c
```diff
@@ -853,6 +853,11 @@ void picoquic_cnx_set_next_wake_time(picoquic_cnx_t* cnx, uint64_t current_time)
                 }
             }
         }
+
+        /* Consider keep alive */
+        if (cnx->keep_alive_interval) {
+            next_time = next_time < cnx->keep_alive_interval ? next_time : cnx->keep_alive_interval;
+        }
```
huitema: Should this be a direct comparison to the interval, or to "current_time + cnx->keep_alive_interval" ?

picoquic/sender.c
```diff
@@ -1511,6 +1516,16 @@ int picoquic_prepare_packet_ready(picoquic_cnx_t* cnx, picoquic_packet* packet,
...
+
+                /* If necessary, encode and send the keep alive packet!
+                 * We only send keep alive packages, when no other data is send!
+                 */
+                if (cnx->keep_alive_interval != 0
+                    && cnx->latest_progress_time + cnx->keep_alive_interval <= current_time
+                    && length == 0) {
+                    bytes[length++] = picoquic_frame_type_ping;
+                    bytes[length++] = 0;
+                }
```
huitema: Yes.

picoquictest/tls_api_test.c
```diff
@@ -1454,6 +1454,76 @@ int tls_api_multiple_versions_test()
...
+        picoquic_enable_keep_alive(test_ctx->cnx_client, keep_alive_interval);
+    }
+
+    /* Perform a couple rounds of sending data */
+    for (int i = 0; ret == 0 && i < 512 && test_ctx->cnx_client->cnx_state != picoquic_state_disconnected; i++) {
+        was_active = 0;
+
+        ret = tls_api_one_sim_round(test_ctx, &simulated_time, &was_active);
+    }
+
```
huitema: Shouldn't that be, "a couple rounds until the virtual time is larger than the idle timeout?" Doing 512 iterations probably does the trick, but it is a bit of a side effect.

huitema:
After merging that, I changed the unit test to ensure that the loop lasted for 2*PICOQUIC_MICROSEC_HANDSHAKE_MAX. It showed a couple of issues. First, the test for sending the keep alive was in an inner block that in some conditions was not executed. I moved it outside these blocks, and added a complete packet formatting sequence. Second, the "progress" timer was not progressing when data or important frames were sent -- my bug I guess, so I fixed it. Third, the progress timer was also not progressing on the server side when non-data important frames were received. Fixed that too. Sorry for sending you through that kind of maze...

bkchr:
Thanks for merging and fixes all the remaining bugs! :)

PR #102, opened 2018-03-06
Implements verify certificate as plug-able callback

bkchr:
The verify certificate function is now set-able. If the callback is NULL, the
default behavior is to use the standard OpenSSL implementation of the verify
certificate function from picotls.

Requires: https://github.com/h2o/picotls/pull/123

### DIFF
--- a/picoquic/picoquic.h
+++ b/picoquic/picoquic.h
@@ -35,6 +35,8 @@
 #include <unistd.h>
 #endif
 
+#include "picotls.h"
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -190,6 +192,22 @@ typedef void (*picoquic_stream_data_cb_fn)(picoquic_cnx_t* cnx,
 typedef void (*cnx_id_cb_fn)(picoquic_connection_id_t cnx_id_local,
     picoquic_connection_id_t cnx_id_remote, void* cnx_id_cb_data, picoquic_connection_id_t * cnx_id_returned);
 
+/* Will be called to verify that the given data corresponds to the given signature.
+ * This callback and the `verify_ctx` will be set by the `verify_certificate_cb_fn`.
+ * If `data` and `sign` are empty buffers, an error occurred and `verify_ctx` should be freed.
+ * Expect `0` as return value, when the data matches the signature.
+ */
+typedef int (*picoquic_verify_sign_cb_fn)(void* verify_ctx, ptls_iovec_t data, ptls_iovec_t sign);
+/* Will be called to verify a certificate of a connection.
+ * The arguments `verify_sign` and `verify_sign_ctx` are expected to be set, when the function returns `0`.
+ * See `verify_sign_cb_fn` for more information about these arguments.
+ */
+typedef int (*picoquic_verify_certificate_cb_fn)(void* ctx, picoquic_cnx_t* cnx, ptls_iovec_t* certs, size_t num_certs,
+                                                 picoquic_verify_sign_cb_fn* verify_sign, void** verify_sign_ctx);
+
+/* Is called to free the verify certificate ctx */
+typedef void (*picoquic_free_verify_certificate_ctx)(void* ctx);
+
 /* QUIC context create and dispose */
 picoquic_quic_t* picoquic_create(uint32_t nb_connections,
     char const* cert_file_name, char const* key_file_name,
@@ -209,6 +227,10 @@ void picoquic_free(picoquic_quic_t* quic);
 
 /* Set cookie mode on QUIC context when under stress */
 void picoquic_set_cookie_mode(picoquic_quic_t* quic, int cookie_mode);
+
+/* Set the verify certificate callback and context. */
+int picoquic_set_verify_certificate_callback(picoquic_quic_t* quic, picoquic_verify_certificate_cb_fn cb, void* ctx,
+                                             picoquic_free_verify_certificate_ctx free_fn);
 
 /* Connection context creation and registration */
 picoquic_cnx_t* picoquic_create_cnx(picoquic_quic_t* quic,
--- a/picoquic/tls_api.c
+++ b/picoquic/tls_api.c
@@ -144,14 +144,14 @@ uint64_t picoquic_crypto_uniform_random(picoquic_quic_t* quic, uint64_t rnd_max)
  *
  * The following is an implementation of xorshift1024* suggested by Sebastiano Vigna,
  * following the general xorshift design by George Marsaglia.
- * The state must be seeded so that it is not everywhere zero. 
+ * The state must be seeded so that it is not everywhere zero.
  *
  * The seed operation gets 64 bits from the crypto random generator. We then run the
  * generator 16 times to mix that input into the 1024 bits of seed[16].
  *
  * If we were really paranoid, we would want to break possible discovery by passing
- * the seeding bits from the crypto random generator through SHA256 or something 
- * similar, so there would be really no way to get at the state of crypto random 
+ * the seeding bits from the crypto random generator through SHA256 or something
+ * similar, so there would be really no way to get at the state of crypto random
  * generator. The 16 rounds of the xorshift process give a pretty good hash, but
  * that can probably be broken by linear analysis. Or at least we have no proof
  * that it cannot be broken.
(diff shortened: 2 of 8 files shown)


bkchr:
Ahh, I forgot that the callback needs a context argument that is set-able. I will add this tomorrow.

picoquic/tls_api.c
```diff
@@ -429,6 +429,55 @@ uint64_t picoquic_get_simulated_time_cb(ptls_get_time_t* self)
...
+#define container_of(ptr, type, member) ((type *)((char *)(ptr) - offsetof(type, member)))
+
+static int verify_certificate_callback(ptls_verify_certificate_t* _self, void* ctx, ptls_t *tls,
+                                       int (**verify_sign)(void *verify_ctx, ptls_iovec_t data, ptls_iovec_t sign),
+                                       void **verify_data,
+                                       ptls_iovec_t *certs,
+                                       size_t num_certs)
+{
+    picoquic_verify_certificate_t *self = container_of(_self, picoquic_verify_certificate_t, cb);
+    picoquic_cnx_t* cnx = (picoquic_cnx_t*)ctx;
```
huitema: The macro "container of" is Linux specific, and will not compile in Windows. For portability, you want something like:
picoquic_verify_certificate_t *self = (picoquic_verify_certificate_t *)((char*)_self - offsetof(struct st_picoquic_verify_certificate_t , cb));

picoquictest/cnx_creation_test.c
```diff
@@ -63,7 +63,7 @@ int cnxcreation_test()
     };
 
     /* Create QUIC context */
-    quic = picoquic_create(8, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, 0, NULL, NULL, NULL, 0);
+    quic = picoquic_create(8, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, 0, NULL, NULL, NULL, 0, NULL);
```
huitema: I am very much guilty of that, but we may want to find an alternative to passing this long list of parameters. Something like "create with default value" followed by a series of "set_this_specific_parameter(quic, param_value)" would probably be less error prone. Not really an issue with your PR, just something we may want to do soon...

huitema:
Any idea for a unit test? Maybe write a connection test with a specific certificate, and a call back that checks that this is the expected cert?

picoquic/tls_api.c
bkchr: I defined the macro in line 455 :D

picoquictest/cnx_creation_test.c
bkchr: I removed the additional argument and created a new function `picoquic_set_verify_certificate_callback` to set the callback.

bkchr:
I added a unit test. The test just ensures that the functions are called and nothing more. I think that should be enough?

picoquic/picoquic.h
```diff
@@ -190,6 +192,22 @@ typedef void (*picoquic_stream_data_cb_fn)(picoquic_cnx_t* cnx,
...
+ * Expect `0` as return value, when the data matches the signature.
+ */
+typedef int (*verify_sign_cb_fn)(void* verify_ctx, ptls_iovec_t data, ptls_iovec_t sign);
+/* Will be called to verify a certificate of a connection.
+ * The arguments `verify_sign` and `verify_sign_ctx` are expected to be set, when the function returns `0`.
+ * See `verify_sign_cb_fn` for more information about these arguments.
+ */
+typedef int (*verify_certificate_cb_fn)(void* ctx, picoquic_cnx_t* cnx, ptls_iovec_t* certs, size_t num_certs,
+                                        verify_sign_cb_fn* verify_sign, void** verify_sign_ctx);
+
```
huitema: Why void**?

picoquic/picoquic.h
```diff
@@ -190,6 +192,22 @@ typedef void (*picoquic_stream_data_cb_fn)(picoquic_cnx_t* cnx,
...
+/* Is called to free the verify certificate ctx */
+typedef void (*free_verify_certificate_ctx)(void* ctx);
+
```
huitema: Maybe the names verify_sign_cb_fn, verify_certificate_cb_fn, or free_verify_certificate_ctx  should have a "picoquic_" prefix to avoid collision with other packages.

Do we actually need a free_verify_certificate_ctx call back? What is the advantage of managing the memory from within the stack, versus letting the application deal with it?

picoquic/tls_api.c
huitema: Sorry, missed that.

picoquic/tls_api.c
```diff
@@ -597,7 +668,7 @@ int picoquic_tlscontext_create(picoquic_quic_t* quic, picoquic_cnx_t* cnx, uint6
         ctx->client_mode = cnx->client_mode;
 
         ctx->tls = ptls_new((ptls_context_t*)quic->tls_master_ctx,
-            (ctx->client_mode) ? 0 : 1);
+            (ctx->client_mode) ? 0 : 1, cnx);
```
huitema: Do you have a dependency on a specific version of Picotls? This change causes the Travis build to fail, yet the build pulls the latest version of h2o/picotls:
~~~
/home/travis/build/private-octopus/picoquic/picoquic/tls_api.c:671:41: error: 
      too many arguments to function call, expected 2, have 3
            (ctx->client_mode) ? 0 : 1, cnx);
~~~

picoquic/tls_api.c
```diff
@@ -429,6 +429,77 @@ uint64_t picoquic_get_simulated_time_cb(ptls_get_time_t* self)
...
+        return PICOQUIC_ERROR_MEMORY;
+    } else {
+        if (ctx->verify_certificate) {
+            free(ctx->verify_certificate);
+        }
+
+        verifier->quic = quic;
+        verifier->cb.cb = verify_certificate_callback;
+        ctx->verify_certificate = &verifier->cb;
+
```
huitema: Why do I get this warning in the Travis build?
~~~
/home/travis/build/private-octopus/picoquic/picoquic/tls_api.c:496:25: warning: 
      incompatible pointer types assigning to 'int (*)(struct
      st_ptls_verify_certificate_t *, ptls_t *, int (**)(void *, ptls_iovec_t,
      ptls_iovec_t), void **, ptls_iovec_t *, size_t)' (aka 'int (*)(struct
      st_ptls_verify_certificate_t *, struct st_ptls_t *, int (**)(void *,
      struct st_ptls_iovec_t, struct st_ptls_iovec_t), void **, struct
      st_ptls_iovec_t *, unsigned long)') from 'int (ptls_verify_certificate_t
      *, void *, ptls_t *, int (**)(void *, ptls_iovec_t, ptls_iovec_t), void
      **, ptls_iovec_t *, size_t)' (aka 'int (struct
      st_ptls_verify_certificate_t *, void *, struct st_ptls_t *, int (**)(void
      *, struct st_ptls_iovec_t, struct st_ptls_iovec_t), void **, struct
      st_ptls_iovec_t *, unsigned long)') [-Wincompatible-pointer-types]
        verifier->cb.cb = verify_certificate_callback;
~~~

picoquic/tls_api.c
```diff
@@ -683,7 +754,7 @@ void picoquic_tlscontext_free(void* vctx)
 {
     picoquic_tls_ctx_t* ctx = (picoquic_tls_ctx_t*)vctx;
     if (ctx->tls != NULL) {
-        ptls_free((ptls_t*)ctx->tls);
+        ptls_free((ptls_t*)ctx->tls, NULL);
```
huitema: Do you have a dependency on a new version of picotls?
~~~
/home/travis/build/private-octopus/picoquic/picoquic/tls_api.c:757:38: error: 
      too many arguments to function call, expected single argument 'tls', have
      2 arguments
        ptls_free((ptls_t*)ctx->tls, NULL);
~~~

bkchr:
Sorry, I thought that it would be clear that this pull request has a dependency on the linked picotls pull request.

picoquic/picoquic.h
bkchr: I can do the prefixing.

Hmm, in my opinion it is better to specify a free function that is called when picoquic is freed. Otherwise, I would need to store the ctx a second time in the application layer and would need to free this, after picoquic was freed. Do you understand my point here?

huitema: OK. But we don't do that anywhere else, so there is a small issue of consistency. Maybe I should be adopting the same approach as Kazuho. Instead of specifying the cb function directly in picotls, he asks you to provide a 
~~~
    struct { function_type cb } function_cb;
~~~
Then the application can allocate a larger struct, and use offset_off or container_off to get the corresponding data. I don't like it too much, but it does make the stack a bit simpler.

picoquic/picoquic.h
bkchr: Because this is call by reference and we require the implementation to set a pointer.

picoquic/picoquic.h
huitema: Just to be clear, your code is fine for now. If we change something, it will have to be done globally.

bkchr: Hmm. But that would not solve the problem with deallocating the context. With picotls that works, because the structs are stored in the global context and you can access them from the outside.

We could define certain values for the callback which mean that the context should be freed? Like picotls is doing it for the verify_sign callback.

bkchr:
@huitema picotls contains the required functions for this pull request. I hope that you now can merge this pull request :)

huitema:
Yes, merging since all tests are passing now.

huitema:
Something I missed in the review was the added dependency on picotls.h in the picoquic.h header. I would rather not have that, as it breaks the build in windows. Also, the philosophy is to have minimal exposure to implementation details in the main header, so we can easily evolve the code without breaking applications. See PR #112 for the quick fix that I did.

PR #112, opened 2018-03-08
Remove dependency on the picotls.h header in picoquic.h

huitema:
The dependency breaks the compile on windows, unless we also include a compatibility header to declare various Linux specific types used by Picotls. I prefer to only add this dependency where strictly needed, i.e. in the module that directly interacts with Picotls.

### DIFF
--- a/picoquic/picoquic.h
+++ b/picoquic/picoquic.h
@@ -35,8 +35,6 @@
 #include <unistd.h>
 #endif
 
-#include "picotls.h"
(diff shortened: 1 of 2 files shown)


bkchr:
Ohh, did not know that this breaks the windows build. I also did not like to expose the picotls types in the public api, but I required it for the certificates parameter. Maybe a forward declaration is a good solution for the beginning.

PR #114, opened 2018-03-09
Adds `picoquic_is_client` function to check if a connection is a client

bkchr:


huitema:
Thanks for this update.

PR #115, opened 2018-03-10
Fix bug which manifests when determining stream is locally initiated

tatsuhiro-t:
If I understand it correctly, the code is checking whether stream is
initiated by local endpoint, or not.  If so, it is enough to check
that the least significant bit has desired value.

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -603,7 +603,7 @@ picoquic_stream_head* picoquic_find_ready_stream(picoquic_cnx_t* cnx, int restri
                     /* Check parity */
                     int parity = cnx->client_mode ? 0 : 1;
 
-                    if (((stream->stream_id & 1) ^ parity) == 1) {
+                    if ((stream->stream_id & 1) == parity) {
                         if (stream->stream_id < cnx->max_stream_id_bidir_remote) {
                             break;
                         }


tatsuhiro-t:
@huitema I'm not sure you read my comment in quicdev slack, ngtcp2 client could not get response from picoquic endpoint in 3rd and 4th interop events.
After debugging your code, I confirmed that with this patch, picoquic server sends response to ngtcp2 client.
Please take a look.

huitema:
Thank you. I think there is a real problem, but I don't think this is the correct fix. Take the example of the first stream issued by the client. The spec specifies that client initiated streams should be even-numbered:
~~~
   The least significant bit (0x1) of the Stream ID identifies the
   initiator of the stream.  Clients initiate even-numbered streams
   (those with the least significant bit set to 0); servers initiate
   odd-numbered streams (with the bit set to 1).  Separation of the
   stream identifiers ensures that client and server are able to open
   streams without the latency imposed by negotiating for an identifier.
~~~
The current test is a bit convoluted, and it could be rewritten. Look at what is happening at the server side:
~~~
                    /* Check parity */
                    int parity = cnx->client_mode ? 0 : 1;

                    if (((stream->stream_id & 1) ^ parity) == 1) {
                        if (stream->stream_id < cnx->max_stream_id_bidir_remote) {
                            break;
                        }
                    } else {
                        break;
                    }
~~~
On the server, client mode is 0, so parity is 1. The next check will return true if the stream_id is even, which it should be. But this code is messy, with use of breaks. Give me a chance to look at it in details.

huitema:
OK. I understand what is happening now. This block of code is meant to check whether it is OK to write on the stream. We are looking at case where a stream was open by the application, and it is possible that the remote party is not yet ready to accept the stream, due to the stream limit restriction. The test:

1) If the stream was opened by the other party:
    * If the stream is bidir, the test should return true (break).
    * if the stream is unidir, it should return false. But there is no need to test for that here, the test is already done in the API.
    * bottom line, if the stream is open by the peer, it should be OK.
2) If the stream is open locally, it is only OK to send if the stream is under the remote party's "max stream" limit.

So the code has to test whether this is opened by a local party. If a server, that means the stream is odd. If a client, it means even. So yes, you are right and the code is wrong.

I understand why we don't hit the issue in unit tests: the tests routinely set the number of streams that the peer can start to a high value. I will merge your PR, and then I will also add a unit test in which the client sets the remote bidir limit to zero. 

Thanks for the fix.

PR #117, opened 2018-03-14
Adds Cmake switch `DISABLE_DEBUG_PRINTF` for disabling debug printf m…

bkchr:
…acro


huitema:
Also, why have a disable flag in the compilation, and an opposite enable flag in the CMake? Given the prototype nature of QUIC, I would rather have debug available by default...

bkchr:
With this approach, the debug output is enabled by default :) If you do not change anything, it will work as before.

Hmm, I find it better to be able to completely disable the debugging output. This also decreases the executable size ^^

PR #118, opened 2018-03-15
Fixes segmentation fault in `picoquic_incoming_encrypted`

bkchr:
When the certificate of a client can not be verified, but the client already
send an encrypted package, the server dies with a segmentation. In the
`picoquic_incoming_encrypted` function, the state is checked for `>=
client_almost_ready` and this returned true. With this reordering, the check
does not pass anymore and the server does not crash.

I could not write any tests that could reproduce the segmentation fault in your library. In my rust library I have some integration tests and they revealed this segmentation fault.


huitema:
Thanks for the fix. We need to have a couple of better tests. I will open an issue for that. For the server certificate, we have a test where the certificate is grossly wrong, but we need another test where it is just slightly wrong, e.g. a good cert for a different key than used by the server. And we need the same thing for client auth.

I would also be interested in learning what actually caused the crash. Testing the system state is an indirect way of stating that all objects have been allocated. For robustness, we may want to add explicit  tests that the encryption or decryption is ready.

bkchr:
The problem was, that the server could not finish the handshake and thus the aaed parameter was null, but as the check was wrong, the server tried to decrypt. But as the pointer was null, it crashed.

Maybe we can finde some good c test framework, the current tests are not that good in my opinion.

I also tried to modify the bad certificate test to crash at the same segmentation fault, but I could not reproduce it in the c tests.

huitema:
Entered issue #120 to track test improvements.

PR #121, opened 2018-03-15
Switching definition of enable/disable

huitema:
So we don't have to define enable switch on all windows projects. Also, a few code changes to make sure we don't have compile warnings when some variables are only used if debug printf is enabled


CMakeLists.txt
```diff
@@ -9,8 +9,8 @@ SET(CMAKE_C_STANDARD 11)
 
 SET(CMAKE_C_FLAGS "-std=c99 -Wall -O2 -g ${CC_WARNING_FLAGS} ${CMAKE_C_FLAGS}")
 
-if((NOT ${DISABLE_DEBUG_PRINTF}) OR (NOT DEFINED DISABLE_DEBUG_PRINTF))
-    SET(CMAKE_C_FLAGS "-DENABLE_DEBUG_PRINTF ${CMAKE_C_FLAGS}")
+if((${DISABLE_DEBUG_PRINTF}) AND (DEFINED DISABLE_DEBUG_PRINTF))
```
bkchr: I think that can be simplified to just `$DISABLE_DEBUG_PRINTF`.

huitema: As in "if ($DISABLE_DEBUG_PRINTF)  SET(...." ?

PR #126, opened 2018-03-16
Reset `next_wake_time` to `0`, if data is added to a stream

bkchr:
This ensures that the connection is scheduled to be woken up in the next call to `picoquic_get_earliest_cnx_to_wake`.

### DIFF
--- a/picoquic/sender.c
+++ b/picoquic/sender.c
@@ -130,6 +130,8 @@ int picoquic_add_to_stream(picoquic_cnx_t* cnx, uint64_t stream_id,
                 *pprevious = stream_data;
             }
         }
+
+        picoquic_cnx_set_next_wake_time(cnx, 0);
     }
 
     return ret;


bkchr:
@huitema okay, updated. Now better?

bkchr:
Yeah, the function requires to get passed the current time, but I looked into the function. When zero is passed, it sets the `next_wake_time` to `current_time`, if the stream is not blocked.

And yes, my code path does not have access to the current time and I think it should not be required at this point. Maybe we need a new flag, `wants_to_send` that is set to `1` if the stream is not blocked.

huitema:
You are unto something with the "want to send" flag. This wake time management is critical for performance, and I think we will need to revisit it as we do performance work. And I much prdfer an explicit flag than banking on a side effect of the code.

I probably need to isolate the wake management in its own module, with specific entries for "app data", "ack", "flow control", "send packet", etc.

I think the simplest fix is to add a current time in the context, instead of passing it through function arguments. Set it up in the top most call, e.g. incoming packet or prepare to send, so it is available everywhere.

But that's a massive change...

bkchr:
Okay, that sounds like a lot of work :D I could also close this pull request for now. I'm maintaining my own branch for picoquic in rust. I need this fix there and can use it until we have real fix in picoquic :)

huitema:
OK. I will work on that first thing tomorrow.

huitema:
I merged this as it provides a basis, but I will check the "current time" issue.

bkchr:
Okay. I yesterday switched to the "get next connection based on wake time" in picoquic-rs. I thought that it would return only connections which wake up time is lower than the current time. However, in some cases it still crested an infinite loop. I think that happens when I create a lot of connections, that will not be able to connect to the endpoint. I needed to revert my changes to get it back working :/

huitema:
See PR #127 for the additional work.

huitema:
How hard would it be to write a "lots of connections" unit test that reproduces your issue?

bkchr:
@huitema I can try to reproduce it, should not be that hard, as it does not requires to have servers that answer on packages.

PR #127, opened 2018-03-17
Provide access to current time through context

huitema:
This builds on previous work to enable "virtual time" management in the TLS stack: either a simulated time for tests, or the actual current time.

### DIFF
--- a/picoquic/tls_api.c
+++ b/picoquic/tls_api.c
@@ -660,6 +660,14 @@ void picoquic_master_tlscontext_free(picoquic_quic_t* quic)
     }
 }
 
+uint64_t picoquic_get_tls_time(picoquic_quic_t* quic)
+{
+    ptls_context_t* ctx = (ptls_context_t*)quic->tls_master_ctx;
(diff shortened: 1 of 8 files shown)


picoquic/tls_api.c
```diff
@@ -660,6 +660,22 @@ void picoquic_master_tlscontext_free(picoquic_quic_t* quic)
...
 
+/*
+ * Get the same time simulation as used for TLS 
+ */
+
+uint64_t picoquic_get_tls_time(picoquic_quic_t* quic)
+{
+    uint64_t now;
+    if (quic->p_simulated_time == NULL) {
+        now = picoquic_current_time();
```
bkchr: Hmm, in picoquic-rs I do not use the `picoquic_current_time` function. My timer starts when I create the initial context and counts upwards. So, I don't have all the milliseconds since 1970. If we start using the timer function internally, I would propose to remove giving the current time into the public interface.

huitema:
I did a lot of rework after this PR, and come up with 2 API. See PR #132

PR #132, opened 2018-03-18
Expose current time and virtual time functions.

huitema:


picoquic/picoquic.h
```diff
@@ -182,6 +182,35 @@ typedef enum {
...
+* as picotls. The default socket code in "picosock.[ch]" uses that time function, and returns the time
+* at which messages arrived. 
+*
+* The function "picoquic_get_virtual_time()" returns the "virtual time" used by the specified quic
+* context, which can be either the current wall time or the simulated time, depending on how the
+* quic context was initialized.
+*/
+
+uint64_t picoquic_current_time(); /* wall time */
+uint64_t picoquic_get_virtual_time(picoquic_quic_t* quic); /* connection time, compatible with simulations */
```
bkchr: I don't like that it is called virtual time, but it can be anything. Either the simulated time, if set or the wall time otherwise.

huitema: How about "picoquic_get_quic_time" ?

huitema:
It should indeed be possible to remove the "current_time" parameter from the "user" function calls. It should be sufficient to pass the "p_simulated_time" pointer when creating the quic context. Once the context is initialized, functions like connection creation, packet prepare or incoming packet could just read the current value of the "virtual time". Of course that's a high impact change, because these functions are used in lots of places.

huitema:
The big constraint is that the concept of time in picoquic should be aligned with that of picotls. The pointer to the simulated time achieves that. We could replace that with a single flag, "use simulated time", so that the default usage would not have to specify anything.

I thought that you also liked the idea of not passing time in the public API functions. I think I can do that.

huitema:
I think I will start by doing it halfway. Leave the function signature as is, but allow the current_time to be set to zero, in which case the code will just use "picoquic_get_quic_time()". That way I don't get a huge list of changes, but can still get to test the simple API.

PR #141, opened 2018-03-19
Add test for wrong TLS version and fix handshake bug

huitema:


picoquic/tls_api.c
```diff
@@ -1362,7 +1362,8 @@ int picoquic_tlsinput_stream_zero(picoquic_cnx_t* cnx)
         }
         ret = 0;
     } else {
-        ret = picoquic_connection_error(cnx, PICOQUIC_TLS_HANDSHAKE_FAILED);
+        (void)picoquic_connection_error(cnx, PICOQUIC_TLS_HANDSHAKE_FAILED);
```
bkchr: Hmm this line makes on of my tests in `picoquic-rs` flakly. Sometimes the handshake error is does not reach the other side.

bkchr:
Using the following patch:
```
diff --git a/picoquic/tls_api.c b/picoquic/tls_api.c
index 97ce855..e7dc4ed 100644
--- a/picoquic/tls_api.c
+++ b/picoquic/tls_api.c
@@ -1370,8 +1370,7 @@ int picoquic_tlsinput_stream_zero(picoquic_cnx_t* cnx)
         }
         ret = 0;
     } else {
-        (void)picoquic_connection_error(cnx, PICOQUIC_TLS_HANDSHAKE_FAILED);
-        ret = 0;
+        ret = picoquic_connection_error(cnx, PICOQUIC_TLS_HANDSHAKE_FAILED);
     }
 
     ptls_buffer_dispose(&sendbuf);
diff --git a/picoquictest/tls_api_test.c b/picoquictest/tls_api_test.c
index 6464263..6fb02cb 100644
--- a/picoquictest/tls_api_test.c
+++ b/picoquictest/tls_api_test.c
@@ -2334,13 +2334,15 @@ int wrong_tls_version_test()
             /* We do expect that the server will be ready to send an HRR */
             ret = picoquic_tlsinput_stream_zero(cnx);
 
-            if (ret != 0) {
+            if (ret != PICOQUIC_ERROR_DETECTED) {
                 DBG_PRINTF("Wrong TLS Hello process return code %x\n", ret);
                 ret = -1;
             } else if (cnx->cnx_state != picoquic_state_handshake_failure) {
                 DBG_PRINTF("Unexpected state %d after wrong TLS Hello process return code %x\n", cnx->cnx_state);
                 ret = -1;
-            } 
+            } else {
+               ret = 0;
+           }
         }
     }
```

Your test also works. Isn't that also a better version?

huitema:
My constraint is that in the case of a TLS protocol mismatch, the "incoming" function should return a valid context instead of a NULL pointer, and that the state variable in the context should be set to picoquic_state_handshake_failure. This will trigger sending a failure message in the next packet. 

Yes, this is a change in behavior -- previous version was returning a NULL pointer in the same circumstances. Is there a way to adapt your test to the new behavior?

With the change that you propose, the incoming function will detect the return error, believe it is fatal, and destroy the connection context. That's not what is needed.

bkchr:
Hmm, without your patch I always received the error of a picoquic handshake failure. I will look into this, maybe I'm doing something wrong on my side.

huitema:
The way the code is written today, the application submits the incoming data to "picoquic_incoming_packet", and receives a call back if something interesting happens, e.g. a new connection was created or some data. The "incoming packet" function should only return an error if some internal error happened, e.g. a failure to allocate memory. For example, it does not return an error if the checksum was wrong, or if the packet was a duplicate. Logically, it should also not return an error if the handshake packet was erroneous.

On the other hand, the current treatment of handshake errors has a weakness. The application will receive a callback when the connection is closed, without having been ever notified that the connection was created. We may want to do better.

PR #148, opened 2018-03-29
Quick label no hash

huitema:


picoquic/logger.c
```diff
@@ -1347,9 +1347,9 @@ void picoquic_log_tls_ticket(FILE* F, picoquic_connection_id_t cnx_id,
     uint32_t lifetime = 0;
     uint32_t age_add = 0;
     uint8_t nonce_length = 0;
-    uint8_t* nonce_ptr = NULL;
+    /* uint8_t* nonce_ptr = NULL; */
```
bkchr: Why do you not delete this commented out code? I prefer to delete code that is not used and if you still require it some later time, you have git to get it back :)

picoquic/quicctx.c
```diff
@@ -160,7 +149,7 @@ picoquic_quic_t* picoquic_create(uint32_t nb_connections,
 
         if (ticket_file_name != NULL) {
             quic->ticket_file_name = ticket_file_name;
-            ret = picoquic_load_tickets(&quic->p_first_ticket, current_time, ticket_file_name);
+            (void) picoquic_load_tickets(&quic->p_first_ticket, current_time, ticket_file_name);
```
bkchr: Why do you want to ignore the return value?

picoquic/tls_api.c
```diff
@@ -22,6 +22,14 @@
 #ifdef _WINDOWS
 #include "wincompat.h"
 #endif
+#ifndef NULL
+#ifdef __cplusplus
+#define NULL 0
```
bkchr: Instead of declaring `NULL`, you could include `<stddef.h>` to get it.

picoquic/tls_api.c
```diff
@@ -956,6 +964,69 @@ Decrypt returns size_t_max (-1) if decryption fails, number of bytes in output o
...
+    ptls_buffer_t hkdf_label;
+    uint8_t hkdf_label_buf[512];
+    int ret;
+
+    ptls_buffer_init(&hkdf_label, hkdf_label_buf, sizeof(hkdf_label_buf));
+
+    ptls_buffer_push16(&hkdf_label, (uint16_t)outlen);
+    ptls_buffer_push_block(&hkdf_label, 1, {
+        if (base_label == NULL)
+        base_label = "tls13 ";
```
bkchr: If the `base_label` is not given, it should be "QUIC". However, as the function now exists in our own code, we do not need to give the `base_label` as argument as it is fixed in quic :) So, I would remove this parameter.

picoquic/tls_api.c
```diff
@@ -956,6 +964,69 @@ Decrypt returns size_t_max (-1) if decryption fails, number of bytes in output o
...
+    ptls_buffer_pushv(&hkdf_label, base_label, strlen(base_label));
+    ptls_buffer_pushv(&hkdf_label, label, strlen(label));
+        });
+    /*
+     * removing the reference to the hash value, as it is not used in QUIC
```
bkchr: Same applies as above, I would remove this commented out code.

picoquic/tls_api.c
```diff
@@ -956,6 +964,69 @@ Decrypt returns size_t_max (-1) if decryption fails, number of bytes in output o
...
+    uint8_t hkdf_label_buf[512];
+    int ret;
+
+    ptls_buffer_init(&hkdf_label, hkdf_label_buf, sizeof(hkdf_label_buf));
+
+    ptls_buffer_push16(&hkdf_label, (uint16_t)outlen);
+    ptls_buffer_push_block(&hkdf_label, 1, {
+        if (base_label == NULL)
+        base_label = "tls13 ";
+    ptls_buffer_pushv(&hkdf_label, base_label, strlen(base_label));
```
bkchr: These two lines need to be intended.

picoquic/tls_api.c
```diff
@@ -956,6 +964,69 @@ Decrypt returns size_t_max (-1) if decryption fails, number of bytes in output o
...
+{
+    ptls_aead_context_t *ctx;
+    uint8_t key[PTLS_MAX_SECRET_SIZE];
+    int ret;
+
+    if ((ctx = (ptls_aead_context_t *)malloc(aead->context_size)) == NULL)
+        return NULL;
+
+    *ctx = (ptls_aead_context_t) { aead };
+    if ((ret = picoquic_get_traffic_key(hash, key, aead->key_size, 0, secret, base_label)) != 0)
```
bkchr: Maybe add `{}` to not force the good old apple `goto Exit` bug :D

picoquic/logger.c
huitema: Yes, you are right. It is just my lack of proficiency with git...

picoquic/tls_api.c
huitema: I thought about that, but I would rather wait until the current discussion of https://github.com/quicwg/base-drafts/issues/1255. If the format reverts to being TLS compatible, we will have
to revert this change. If it doesn't, there are other things that can be cleaned up, e.g. the construction of label for plaintext AEAD, and the generation of PN encryption keys. I added issue #149 to track that.

PR #153, opened 2018-04-07
Align to quic draft 11 and TLS draft 28

huitema:
@bkchr this brings changes to the API, because there are now 3 connection identifiers:
 * The initial CID, used for the very first packet and also used to compute the cleartext encryption key;
 * The server CID, which the client will copy in packets sent to the server;
 * The client CID, which the server will copy in packets sent to the client.
The CID are now variable length {0 | [4..18]}. The "local" length is set as a parameter of the quic context. It is expected to be the same for all connections.

@dbavatar I have kept the call back logic for building up the connection ID as is, expecting a CID size of 8 bytes. You may want to look at using the API building CID as  <nonce | server-id | oracle> encrypted with a load-balancer specified key.


picoquic/frames.c
```diff
@@ -277,6 +277,45 @@ int picoquic_decode_stream_reset_frame(picoquic_cnx_t* cnx, uint8_t* bytes,
...
+        *consumed = bytes_max;
+        ret = -1;
+    }
+
+    return ret;
+}
+
+int picoquic_decode_connection_id_frame(picoquic_cnx_t* cnx, uint8_t* bytes, size_t bytes_max, size_t* consumed)
+{
+    int ret = picoquic_skip_connection_id_frame(bytes, bytes_max, consumed);
```
bkchr: Why are you skipping the frame? Should that just be a first implementation? Maybe I would add a `TODO: Store new ID `

picoquic/frames.c
```diff
@@ -2248,8 +2282,8 @@ int picoquic_skip_frame(uint8_t* bytes, size_t bytes_max, size_t* consumed,
             break;
         case picoquic_frame_type_new_connection_id:
             /* Skip the sequence */
-            byte_index += picoquic_varint_skip(&bytes[byte_index]);
```
bkchr: Couldn't somehow `consumed == 0`? And why at all `-1`?

picoquic/packet.c
```diff
@@ -50,221 +50,183 @@ int picoquic_parse_packet_header(
...
-                ph->ptype = picoquic_packet_version_negotiation;
-                ph->pn = 0;
-                ph->pnmask = 0;
-                ph->pn_offset = 0;
-                ph->version_index = -1;
-
-                if (*pcnx == NULL) {
-                    *pcnx = picoquic_cnx_by_id(quic, ph->cnx_id);
+            uint8_t l_dest_id, l_srce_id;
+            /* The bytes at position 1..3 describe the version */
```
bkchr: The version is 4byte long, so shouldn't that byte 1-4?

picoquic/packet.c
```diff
@@ -50,221 +50,183 @@ int picoquic_parse_packet_header(
...
+            ph->vn = PICOPARSE_32(bytes + 1);
+            /* Obtain the connection ID lengths from the byte following the version */
+            picoquic_parse_packet_header_cnxid_lengths(bytes[5], &l_dest_id, &l_srce_id);
+            if (6 + l_dest_id + l_srce_id + 4 > (int) length) {
```
bkchr: Maybe add a comment what each number is. 6 = version + cnx id lengths + type, 4= packet number.

picoquic/packet.c
huitema: Yes.

picoquic/packet.c
huitema: OK, doing that.

picoquic/frames.c
huitema: Consumed always includes the type byte, already included in byte index, hence -1. Rewriting for clarity, and adding comment.
~~~
ret = picoquic_skip_connection_id_frame(&bytes[byte_index-1], bytes_max, consumed);
~~~
Consumed is set to bytes_max in case of error, thus not zero.

picoquic/frames.c
huitema: Yes.

huitema: Yes.

PR #166, opened 2018-04-17
Add client auth in the simulated migration branch

huitema:


huitema:
Merging so the issue can be fixed in the simulated migration branch.

PR #171, opened 2018-04-21
logger: Fix crashes, caught by Address Sanitizer

Lekensteyn:
While the STREAM_BLOCKED frame has a Stream ID and offset field, the
BLOCKED frame only has a offset field. Fix another bounds issue as well.


huitema:
Thanks for the fix, @Lekensteyn. I also added a new issue: Add logger fuzz test #172.

huitema:
And also added this issue: varint skip is unsafe #173.

Issue #90, opened 2018-02-24
Socket code appears to not work from within a container


huitema:
This was reported by @alagoutte during the installation of the Travis build verifier. Need to check.

Lekensteyn:
I have hit this issue locally on Arch Linux as well. It does not always happen, perhaps it is a race condition.

```
$ ./picoquic_ct sockets
Starting test number 45, sockets
    Success.
$ ./picoquic_ct sockets
Starting test number 45, sockets
ic/picoquic/picosocks.c:561 [picoquic_select]: Could not receive packet on UDP socket[0]= 5!

    Fails, error: -1.
```

PR #174, opened 2018-04-21
More crash/test fixes

Lekensteyn:
Another memory safety fix and improvements for the test suite. After this, `./picoquic_ct` runs passes tests without crashing. (There are lots of memory leaks, but I ignored those for now.)

Also figured out why #90 happened and added a workaround for it in the travis config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,8 +25,9 @@ before_install:
 before_script:
   # First build external lib
   - ./ci/build_picotls.sh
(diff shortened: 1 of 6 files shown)


Lekensteyn:
In the travis container, IPv6 is apparently disabled by default. This resulted in failure to connect to `::1`. The solution was to enable IPv6 (see the commit message for references).

PR #198, opened 2018-06-05
Implement the VEC extension to the Spin Bit (0K110SVV)

ferrieux:
Hello. This implements the Valid Edge Counter extension to the Spin Bit, as per section 4 of draft-trammell-quic-spin . To exercise it and validate the result, take a pcap and submit it to our online spibit+VEC analysis tool at http://193.252.113.227/cgi-bin/quicspin.cgi

Welcoming any comments/questions: alexandre.ferrieux@orange.com

### DIFF
--- a/picoquic/packet.c
+++ b/picoquic/packet.c
@@ -77,6 +77,7 @@ int picoquic_parse_packet_header(
                 ph->offset += picoquic_parse_connection_id(bytes + ph->offset, l_srce_id, &ph->srce_cnx_id);
 
                 /* Not applicable for long packets. */
+		ph->hasspinbit = 0;
                 ph->spin = 0;
 
                 if (ph->vn == 0) {
@@ -253,7 +254,9 @@ int picoquic_parse_packet_header(
                      ph->ptype = picoquic_packet_1rtt_protected_phi1;
                  }
 
+		 ph->hasspinbit = 1;
                  ph->spin = (bytes[0] >> 2) & 1;
+		 ph->VEC = bytes[0] & 0x03 ;
 
                  ph->pn_offset = ph->offset;
                  ph->pn = 0;
@@ -1139,6 +1142,14 @@ int picoquic_incoming_encrypted(
         /* Packet is correct */
         if (ph->pn64 > cnx->first_sack_item.end_of_sack_range) {
             cnx->current_spin = ph->spin ^ cnx->client_mode;
+	    if (ph->hasspinbit && cnx->current_spin != cnx->prev_spin) {
+	      // got an edge 
+	      cnx->prev_spin = cnx->current_spin;
+	      cnx->Edge = 1;
+	      cnx->VEC = ph->VEC + 1;
+	      if (cnx->VEC > 3) cnx->VEC = 3;
+	      cnx->lastTrigger =  picoquic_get_quic_time(cnx->quic);
+	    }
         }
 
         /* Do not process data in closing or draining modes */
--- a/picoquic/picoquic_internal.h
+++ b/picoquic/picoquic_internal.h
@@ -54,6 +54,8 @@ extern "C" {
 #define PICOQUIC_CWIN_INITIAL (10 * PICOQUIC_MAX_PACKET_SIZE)
 #define PICOQUIC_CWIN_MINIMUM (2 * PICOQUIC_MAX_PACKET_SIZE)
 
+#define PICOQUIC_VEC_LATE 1000 /* microseconds */
+
 /*
  * Types of frames
  */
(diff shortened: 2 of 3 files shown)


picoquic/packet.c
```diff
@@ -77,6 +77,7 @@ int picoquic_parse_packet_header(
                 ph->offset += picoquic_parse_connection_id(bytes + ph->offset, l_srce_id, &ph->srce_cnx_id);
 
                 /* Not applicable for long packets. */
+		ph->hasspinbit = 0;
                 ph->spin = 0;
```
huitema: I am not too picky about variable names, but I would have used "has_spin_bit" instead of "hasspinbit"

picoquic/packet.c
```diff
@@ -253,7 +254,9 @@ int picoquic_parse_packet_header(
                      ph->ptype = picoquic_packet_1rtt_protected_phi1;
                  }
 
+		 ph->hasspinbit = 1;
                  ph->spin = (bytes[0] >> 2) & 1;
+		 ph->VEC = bytes[0] & 0x03 ;
```
huitema: Careful there. This branch of code supports both draft-11 and draft-12. You cannot use the VEC functionality in draft-11, the two bits are defined for another usage. So you need to test the version,  set VEC to zero if draft 11, and to the 2 bits if draft-12 and higher.

picoquic/packet.c
```diff
@@ -1139,6 +1142,13 @@ int picoquic_incoming_encrypted(
         /* Packet is correct */
         if (ph->pn64 > cnx->first_sack_item.end_of_sack_range) {
             cnx->current_spin = ph->spin ^ cnx->client_mode;
+	    if (ph->hasspinbit && cnx->current_spin != cnx->prev_spin) {
+	      // got an edge 
+	      cnx->prev_spin = cnx->current_spin;
+	      cnx->Edge = 1;
+	      cnx->VEC = (ph->VEC==3) ? 3 : (ph->VEC + 1);
+	      cnx->lastTrigger =  picoquic_get_quic_time(cnx->quic);
+	    }
```
huitema: Can you find some more expressive variable names than "Edge" or "VEC" or "LastTrigger"? Maybe something like "spin_edge", "spin_vec" and "spin_last_trigger"? My concern is that when someone later reads the code they do not understand the purpose of the variables.

picoquic/picoquic_internal.h
```diff
@@ -376,6 +378,11 @@ typedef struct st_picoquic_cnx_t {
     unsigned int ack_needed : 1;
     unsigned int current_spin : 1; /* Current value of the spin bit */             
     unsigned int client_mode : 1; /* Is this connection the client side? */
+    unsigned int prev_spin : 1;  /* previous Spin bit */
+    unsigned int VEC : 2; 
+    unsigned int Edge : 1;
+    uint64_t lastTrigger;
+
```
huitema: See previous comment on variable names.

picoquic/sender.c
```diff
@@ -319,10 +319,21 @@ uint32_t picoquic_create_packet_header_12(
         /* Create a short packet -- using 32 bit sequence numbers for now */
         uint8_t K = (packet_type == picoquic_packet_1rtt_protected_phi0) ? 0 : 0x40;
         const uint8_t C = 0x30;
+	uint8_t VEC = (cnx->VEC );
```
huitema: Same remark about version dependency. We cannot implement VEC with draft-11. We have to either write some version dependent test, or decide to just forget about draft 11 compatibility. But if we do that, I need to make a snapshot of the code and archive it before we merge this PR.

picoquic/packet.c
ferrieux: The idea was that these internal state variables are harmless in draft-11, since the serialization code ignores them, and  that indeed support for earlier versions would soon disappear. That's why I felt it was OK to spare the extra version-switch complexity. OK to do it if you prefer.

ferrieux:
Travis CI spurious failure (keyserver). Sh*t happens.

private-octopus:
Just restarted the CLANG build.

picoquic/packet.c
ferrieux: Please comment on these considerations ; I'll comply.

ferrieux:
Thanks, worked. Please look at the comment in the review too.

picoquic/picoquic_internal.h
```diff
@@ -376,6 +378,11 @@ typedef struct st_picoquic_cnx_t {
     unsigned int ack_needed : 1;
     unsigned int current_spin : 1; /* Current value of the spin bit */             
     unsigned int client_mode : 1; /* Is this connection the client side? */
+    unsigned int prev_spin : 1;  /* previous Spin bit */
+    unsigned int spin_vec : 2; 
+    unsigned int spin_edge : 1;
+    uint64_t lastTrigger;
+
```
huitema: What is the definition of last trigger?

huitema: How are these variables initialized? Do you rely on default init to zero?

ferrieux: It is the timestamp of the incoming packet that triggered the last spin bit change, to be aired on next outgoing packet.  Necessary to detect when the reaction is too slow, and advertise it through VEC==1. Now documented in comments.

ferrieux: I rely on a memset to zero in cnx creation.

PR #199, opened 2018-06-06
Big frame refactoring. Phase 1. Decoder function.

igorlord:
This is a very large commit.

# Purpose

I am trying to clean up and simplify a lot of code to make the semantics of frame operations easier to read and understand and, consequently, easier to keep bug-free.

This is just the first step in this direction.  I've only changed piciquic_decode_* functions.

Each function and its helpers check their own byte constraints.

I have also changed how frame accounting works (but only for piciquic_decode_* functions so far).
Before:
- keeping three variables responsible for each frame (buffer ptr, buffer size, and consumed bytes as an output)
- requiring the function to keep updating both the current index and the remaining buffer size.
- requiring the caller to update its own the current index and the remaining buffer size, based on the consumed bytes output of the callee.

Now:
- keeping two variables responsible for each frame (address of buffer ptr, max buffer pointer)
- requiring the function to keep updating only a single buffer ptr as it consumes data
- caller does not need to do anything at all

This allowed me to save a lot of code and improve readability by removing extraneous bookkeeping.

Also, early return in many places simplified code and improved readability to placing logically sequential operations in a single control flow of the function, instead of the normal flow control continuing into a branch of some conditional.

# Testing

Since I work on Linux, I was only able to test via the demo client/server app.
I have had to make a number of changes to unit tests -- both to get them to work with the new API at times and to correct semantic differences.  I am worried about the latter ones.  It is important to run Windows tests before establishing a complete trust in this update.

# PS

There was a few places where I had to add code for "impedance matching" between the new-style and old-style accounting functions.  I expect these will go away and get more streamlined as more of the codebase gets converted.

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -170,6 +170,54 @@ int picoquic_flow_control_check_stream_offset(picoquic_cnx_t* cnx, picoquic_stre
     return ret;
 }
 
+
+/* ********* 
+ *
+ *  Frames
+ *
+ * ******** */
+
+typedef int (*picoquic_decode_frame_t)(picoquic_cnx_t* cnx, const uint8_t** pbytes, const uint8_t* bytes_max, int restricted, uint64_t current_time);
+
+static int picoquic_varint_skip_check(picoquic_cnx_t* cnx, const uint8_t** pbytes, const uint8_t* bytes_max)
+{
+    if (*pbytes < bytes_max) {
+        size_t length = picoquic_varint_skip(*pbytes);
+        if (length <= bytes_max - *pbytes) {
+            *pbytes += length;
+            return 0;
+        }
+    }
+    return picoquic_connection_error(cnx, PICOQUIC_TRANSPORT_FRAME_FORMAT_ERROR);
+}
+
+static int picoquic_varint_decode_check(picoquic_cnx_t* cnx, const uint8_t** pbytes, const uint8_t* bytes_max, uint64_t* n64)
+{
+    if (*pbytes < bytes_max) {
+        size_t length = picoquic_varint_decode(*pbytes, bytes_max - *pbytes, n64);
+        if (length > 0) {
+            *pbytes += length;
+            return 0;
+        }
+    }
+    return picoquic_connection_error(cnx, PICOQUIC_TRANSPORT_FRAME_FORMAT_ERROR);
+}
+
+#define DECL_PICOQUIC_FIXED_DECODE_CHECK(n)                                                                                        \
+static int picoquic_fixed##n##_decode_check(picoquic_cnx_t* cnx, const uint8_t** pbytes, const uint8_t* bytes_max, uint##n##_t* v) \
+{                                                                                                                                  \
+    if (bytes_max - *pbytes < sizeof(*v)) {                                                                                        \
+        return picoquic_connection_error(cnx, PICOQUIC_TRANSPORT_FRAME_FORMAT_ERROR);                                              \
+    }                                                                                                                              \
+    *v = PICOPARSE_##n(*pbytes);                                                                                                   \
+    *pbytes += sizeof(*v);                                                                                                         \
+    return 0;                                                                                                                      \
+}
+
+DECL_PICOQUIC_FIXED_DECODE_CHECK(16)
+DECL_PICOQUIC_FIXED_DECODE_CHECK(64)
+
+
 /*
  * RST_STREAM Frame
  * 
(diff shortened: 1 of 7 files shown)


igorlord:
One more change: picoquic_decode_* functions no longer receive their own frame_type byte in the buffer.  I did this to avoid just having to skip it in every single decoder, except for one -- stream frame decoder.  That one reaches one octet back to gets its frame type -- a bit of a hack, indeed.

huitema:
Give me some time to review this, please. I just did a cursory review. I understand what you want to do -- change the parsing logic to use pointer to use pointer comparisons, refactor with tables of functions, etc. I have two broad requests:

1) Do it piecemeal, with one type of change at a time. At each stage, verify that all the tests are passing. Only do small mechanical changes in the tests.

2) Respect the coding style. One thing I insist on is having no goto, and a single exit point per function -- one single return statement at the end, not several return statements a multiple places in the code.

I will try to load your branch on a windows computer to check it.

By the way, the tests on frame parsing are failing. We should fix that!

picoquic/frames.c
```diff
@@ -1888,211 +1896,140 @@ int picoquic_prepare_path_challenge_frame(uint8_t* bytes,
...
+  DUP16({0}),                                            /* 0x90 - 0x9f */
+  DUP16({0}),                                            /* 0xa0 - 0xaf */
+  DUP16({0}),                                            /* 0xb0 - 0xbf */
+  DUP16({0}),                                            /* 0xc0 - 0xbf */
+  DUP16({0}),                                            /* 0xd0 - 0xdf */
+  DUP16({0}),                                            /* 0xe0 - 0xef */
+  {0}, {0}, {0}, {0}, {0}, {0}, {0}, {0},
+  {0}, {0}, {0}, {0}, {0}, {0}, {0}, {0}
+};
+
```
huitema: Can I suggest that we do not implement the function table in the current PR? I would rather break this current PR in two:

1) Change the calling logic for some frame types to use the pointer/pointer max template, intead of the byte/length/consumed template, but do not change the signature of the picoquic_decode_frames function. Just keep the same switch statement, calling the new functions instead of the old ones.

2) Once the first PR is verified, change from the switch statement to a function table resolution.

The advantage is that the first changes will be almost entirely internal to "frames.c", with no need to change any of the functions that consumes "picoquic_decode_frames". In particular, there will not be a need to change any of the test cases. We will be able to verify that the decoding works.

In fact, we could consider doing the same thing with the "skip" functions, again before moving to the table structure. That will also allow us to use the commonality between "decode with no side effect" and "skip".

picoquic/picoquic_internal.h
```diff
@@ -79,7 +79,8 @@ typedef enum {
     picoquic_frame_type_ack_range_min_old = 0xa0,
     picoquic_frame_type_ack_range_max_old = 0xbf,
     picoquic_frame_type_stream_range_min_old = 0xc0,
-    picoquic_frame_type_stream_range_max_old = 0xcf
+    picoquic_frame_type_stream_range_max_old = 0xcf,
+    picoquic_frame_type_expired_stream_data = 0xf0
```
huitema: Is picoquic_frame_type_expired_stream_data the expansion type?

picoquic/picoquic_internal.h
```diff
@@ -528,7 +529,7 @@ void picoformat_64(uint8_t* bytes, uint64_t n64);
 size_t picoquic_varint_encode(uint8_t* bytes, size_t max_bytes, uint64_t n64);
 void picoquic_varint_encode_16(uint8_t* bytes, uint16_t n16);
 size_t picoquic_varint_decode(const uint8_t* bytes, size_t max_bytes, uint64_t* n64);
-size_t picoquic_varint_skip(uint8_t* bytes);
+size_t picoquic_varint_skip(const uint8_t* bytes);
```
huitema: I get that adding const is good hygiene, but do we really need that change in the first package? This PR is too big. Anything that reduces its size would be nice.

picoquic/picoquic_internal.h
```diff
@@ -683,7 +684,7 @@ int picoquic_prepare_misc_frame(picoquic_misc_frame_header_t* misc_frame, uint8_
 
 /* send/receive */
 
-int picoquic_decode_frames(picoquic_cnx_t* cnx, uint8_t* bytes,
+int picoquic_decode_frames(picoquic_cnx_t* cnx, const uint8_t* bytes,
     size_t bytes_max, int restricted, uint64_t current_time);
```
huitema: See remark above. I would leave that signature unchanged in the first PR, to contain the size and scope of that PR.

picoquic/quicctx.c
```diff
@@ -1236,12 +1236,12 @@ int picoquic_connection_error(picoquic_cnx_t* cnx, uint32_t local_error)
...
+        DBG_PRINTF("Protocol error (0x%x)", local_error);
     } else if (cnx->cnx_state < picoquic_state_client_ready) {
         cnx->local_error = local_error;
         cnx->cnx_state = picoquic_state_handshake_failure;
 
-        DBG_PRINTF("Protocol error %x", local_error);
+        DBG_PRINTF("Protocol error 0x%x", local_error);
     }
 
     return PICOQUIC_ERROR_DETECTED;
```
huitema: Good fixes, but unrelated. We should break them in a separate PR.

picoquicfirst/picoquicdemo.c
```diff
@@ -760,33 +766,35 @@ static void first_client_callback(picoquic_cnx_t* cnx,
...
+        stream_ctx->F = NULL;
+        ctx->nb_open_streams--;
+        fin_stream_id = stream_id;
 
-            fprintf(stdout, "Received file %s, after %d bytes, closing stream %d\n",
+        fprintf(stdout, "Received file %s, after %d bytes, closing stream %d\n",
                 strip_endofline(buf, sizeof(buf), (char*)&stream_ctx->command[4]),
                 (int)stream_ctx->received_length, stream_ctx->stream_id);
-        }
     }
```
huitema: Again, these are probably good fixes, but unrelated. Should be moved to a separate PR.

picoquictest/stream0_frame_test.c
```diff
@@ -132,12 +132,12 @@ static int StreamZeroFrameOneTest(struct test_case_st* test)
...
-    size_t consumed = 0;
     uint64_t current_time = 0;
 
     for (size_t i = 0; ret == 0 && i < test->list_size; i++) {
-        if (0 != picoquic_decode_stream_frame(&cnx, test->list[i].packet,
-                     test->list[i].packet_length, 1, &consumed, current_time)) {
+        const uint8_t* bytes = test->list[i].packet + 1;
+        if (0 != picoquic_decode_stream_frame(&cnx, &bytes, test->list[i].packet + test->list[i].packet_length,
+                                              1, current_time)) {
             FAIL(test, "packet %" PRIst, i);
```
huitema: I would much rather leave the tests completely unchanged in the first PR. That would minimize the risks of an inadvertent issue. And it will also make the PR smaller.

picoquictest/tls_api_test.c
```diff
@@ -3236,4 +3236,4 @@ int client_error_test()
     }
```
huitema: This is a weird test, dating back to the definition of ping/pong frames in draft 7 or 8. You should leave it alone in the first PR. We probably want to just get rid of it in a separate PR.

igorlord:
It would be nice to be able to build and run the tests on Linux.

Yes, this PR is very large.  Sometimes I get carried away with refactoring things.  I'll see when I will get a chance to break this down.

picoquic/frames.c
```diff
@@ -170,6 +170,54 @@ int picoquic_flow_control_check_stream_offset(picoquic_cnx_t* cnx, picoquic_stre
...
+    if (*pbytes < bytes_max) {
+        size_t length = picoquic_varint_skip(*pbytes);
+        if (length <= bytes_max - *pbytes) {
+            *pbytes += length;
+            return 0;
+        }
+    }
+    return picoquic_connection_error(cnx, PICOQUIC_TRANSPORT_FRAME_FORMAT_ERROR);
+}
+
```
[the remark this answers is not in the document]
igorlord: Yes, this can get refactored into single-return.  For functions that can fail in different ways along the way, it may mean adding some extra code and moving the common-case flow from the "top-level" into separate "then" blocks.  But it can be done systematically.

picoquic/frames.c
```diff
@@ -170,6 +170,54 @@ int picoquic_flow_control_check_stream_offset(picoquic_cnx_t* cnx, picoquic_stre
...
+        return picoquic_connection_error(cnx, PICOQUIC_TRANSPORT_FRAME_FORMAT_ERROR);                                              \
+    }                                                                                                                              \
+    *v = PICOPARSE_##n(*pbytes);                                                                                                   \
+    *pbytes += sizeof(*v);                                                                                                         \
+    return 0;                                                                                                                      \
+}
+
+DECL_PICOQUIC_FIXED_DECODE_CHECK(16)
+DECL_PICOQUIC_FIXED_DECODE_CHECK(64)
+
```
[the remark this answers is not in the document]
igorlord: Sure.  Away with templates.

picoquic/frames.c
```diff
@@ -220,56 +268,89 @@ int picoquic_prepare_stream_reset_frame(picoquic_stream_head* stream,
...
+static int picoquic_decode_padding_or_ping_frame(picoquic_cnx_t* cnx, const uint8_t** pbytes, const uint8_t* bytes_max,
+                                                    picoquic_frame_type_enum_t frame_type)
+{
+    /* As an optimization, consume multiple subsequent padding in one go */
+    while (*pbytes < bytes_max && **pbytes == frame_type) {
+        ++*pbytes;
+    }
+    return 0;
+}
+
```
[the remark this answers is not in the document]
igorlord: Yes, it was an entry in the frame description table -- whether the frame required an ack (the other is whether the frame can be used in the 'restricted' context).

picoquic/frames.c
```diff
@@ -220,56 +268,89 @@ int picoquic_prepare_stream_reset_frame(picoquic_stream_head* stream,
...
+{
+    /* TODO: check that the stream number is valid */
+    int ret = picoquic_varint_skip_check(cnx, pbytes, bytes_max);
+    if (ret == 0) {
+        /* Skip the max data offset */
+        ret = picoquic_varint_skip_check(cnx, pbytes, bytes_max);
+    }
+    return ret;
+}
+
```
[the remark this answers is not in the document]
igorlord: I do not think there is duplication here.  I tried to remove duplication.

picoquic/frames.c
```diff
@@ -220,56 +268,89 @@ int picoquic_prepare_stream_reset_frame(picoquic_stream_head* stream,
...
+int picoquic_decode_stream_id_needed_frame(picoquic_cnx_t* cnx, const uint8_t** pbytes, const uint8_t* bytes_max, int restricted, uint64_t current_time)
+{
+    return picoquic_varint_skip_check(cnx, pbytes, bytes_max);
+}
+
```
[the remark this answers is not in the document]
igorlord: I was trying to optimize human-readable code, not machine code.  The human-readable code does get smaller (while the extra indirection will likely be optimized away by compiler and HW).

It is possible to switch to
   next_pointer = (cnx, current_pointer, max_pointer, &ret, ...)
It is a bit unusual (usually return is a status is the return of the function), but it can work.  It does add an extra argument to the function.

> Also, I am dead set against the multiple return structure. It is a form of goto, and I insist on using structured code instead.

Your position regarding returns (and, implicitly, goto) is clear.  What about "breaks" and "continues"?

(My preference is to hold having the code follow the "natural" logic of the process encoded as much as possible.  I.e. if someone asked me to do some multi-step task, and I find early on that I cannot finish, I would quit right away and tell them what happened right there and then.  I'd rather bend the coding style to align with thought flow than the other way around.  But, yes, one coding style can be converted into the other.)

picoquic/frames.c
```diff
@@ -1258,110 +1332,76 @@ static picoquic_packet* picoquic_process_ack_range(
...
-                    PICOQUIC_TRANSPORT_FRAME_ERROR(bytes[0]));
-                break;
-            }
+    /* parse routes are expecting first byte to be the frame type for now */
+    if (picoquic_parse_ack_header((*pbytes)-1, bytes_max - *pbytes + 1,
+                                  &num_block, &largest, &ack_delay, &consumed,
+                                  cnx->remote_parameters.ack_delay_exponent) != 0) {
+        goto conn_error;
+    }
+    *pbytes += consumed - 1;
```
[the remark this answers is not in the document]
igorlord: I agree.  That's why I like C++ references.  Avoids the need to keep doing this indirection business.

picoquic/frames.c
igorlord: This is a reasonable suggestion.

I was going to move the skip functions into the table as well.  Just did not get to them and decided to focus on decode functions first.

picoquic/picoquic_internal.h
igorlord: I should had left this out of the PR.

picoquic/picoquic_internal.h
igorlord: Right.  I just hated to write new functions lacking const and go back and redo them later.  But you are correct -- too many unrelated themes in a single PR.

picoquic/quicctx.c
igorlord: Yep

picoquic/frames.c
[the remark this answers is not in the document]
huitema: I saw that. It works.

picoquictest/stream0_frame_test.c
igorlord: The test is reaching deep into the frame decoder and calling its function that needs an interface change.  It is the test violating the abstraction layer.

picoquic/frames.c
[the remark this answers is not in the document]
huitema: We will see that when we add the "skip" functions for these frame types.

igorlord:
Cool.  Thanks!  I'll look there to see what Travis does.

picoquic/frames.c
huitema: OK, do what is easiest. The problem is not so much the table as the need to keep the existing external API for now, so we don't have to change the protocol implementation or the tests. As long as we achieve that, I am fine.

igorlord: The "external API" is not super-well defined.  APIs for callbacks is one thing (reasonably clear -- external API).  But one test is calling picoquic_decode_stream_frame directly.  That's a layer violation, so it may need to change.  (I had almost marked all picoquic_decode_*_frame functions static.)

picoquictest/stream0_frame_test.c
huitema: This is a unit test for the stream 0 frame handling. So yes, it does call specific low level functions -- after all, the purpose is precisely to test them.

PR #200, opened 2018-06-10
Basic stress test

huitema:
Create a stress test. The test is parameterized by duration, number of clients, and number of queries in a client session. The default stress run, part of the unit test, runs for 2 minutes with 4 clients and 16 streams. The plan is to use additional "-s" parameters to the picoquic_ct test exe to change these parameters, but we need to keep it simple initially. These parameters are not yet implemented.

We can also think of adding some variability in the client profiles. For example, we could simulate clients that go away without completing the connection, or without closing it. This is not implemented yet. We could also add a fuzzing option, in which clients sometimes send bogus messages. That, too, is not implemented.

The simple stress test found many "hot loops" bugs, which are also fixed in this PR.

### DIFF
--- a/picoquic/picosocks.c
+++ b/picoquic/picosocks.c
@@ -476,7 +476,6 @@ int picoquic_sendmsg(SOCKET_TYPE fd,
                 cmsg_2->cmsg_level = IPPROTO_IPV6;
                 cmsg_2->cmsg_type = IPV6_DONTFRAG;
                 cmsg_2->cmsg_len = CMSG_LEN(sizeof(int));
-//                *((int *)CMSG_DATA(cmsg_2)) = val;
                 memcpy(CMSG_DATA(cmsg_2), &val, sizeof(int));
                 control_length += CMSG_SPACE(sizeof(int));
             }
--- a/picoquic/picosocks.c
+++ b/picoquic/picosocks.c
@@ -476,7 +476,8 @@ int picoquic_sendmsg(SOCKET_TYPE fd,
                 cmsg_2->cmsg_level = IPPROTO_IPV6;
                 cmsg_2->cmsg_type = IPV6_DONTFRAG;
                 cmsg_2->cmsg_len = CMSG_LEN(sizeof(int));
-                *((int *)CMSG_DATA(cmsg_2)) = val;
+//                *((int *)CMSG_DATA(cmsg_2)) = val;
+                memcpy(CMSG_DATA(cmsg_2), &val, sizeof(int));
                 control_length += CMSG_SPACE(sizeof(int));
             }
         }
--- a/picoquic/sender.c
+++ b/picoquic/sender.c
@@ -63,21 +63,7 @@ int picoquic_add_to_stream(picoquic_cnx_t* cnx, uint64_t stream_id,
             int parity = cnx->client_mode ? 0 : 1;
             if ((stream_id & 1) != parity) {
                 ret = PICOQUIC_ERROR_INVALID_STREAM_ID;
-            } else {
-#if 0
-                if ((stream_id & 2) == 0) {
-                    if (stream_id > cnx->max_stream_id_bidir_remote) {
-                        ret = PICOQUIC_ERROR_INVALID_STREAM_ID;
-                    }
-                } else {
-                    is_unidir = 1;
-
-                    if (stream_id > cnx->max_stream_id_unidir_remote) {
-                        ret = PICOQUIC_ERROR_INVALID_STREAM_ID;
-                    }
-                }
-#endif
-            }
+            } 
 
             if (ret == 0) {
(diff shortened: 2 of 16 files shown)


huitema:
I am excluding the stress test from the travis run, because there is an execution error in Linux that I have a hard time debugging. To repro, just run:
```
./picoquic_ct stress
```

picoquic/sender.c
```diff
@@ -64,6 +64,7 @@ int picoquic_add_to_stream(picoquic_cnx_t* cnx, uint64_t stream_id,
             if ((stream_id & 1) != parity) {
                 ret = PICOQUIC_ERROR_INVALID_STREAM_ID;
             } else {
+#if 0
```
igorlord: Why is this #ifdef-ed out?

bkchr:
Ohh, it took me almost the full day to find this little bug :( But now the stresstest works for me on Linux!

picoquictest/stresstest.c
```diff
@@ -0,0 +1,875 @@
...
+
+int stress_test()
+{
+    int ret = 0;
+    picoquic_stress_ctx_t stress_ctx;
+
+    /* Initialization */
+    memset(&stress_ctx, 0, sizeof(picoquic_stress_ctx_t));
+    stress_set_ip_address_from_index(&stress_ctx.server_addr, -1);
+    if (stress_ctx.nb_clients > PICOQUIC_MAX_STRESS_CLIENTS) {
```
bkchr: The context is initialized to `0` two lines above, so this check is not required.

picoquic/picosocks.c
```diff
@@ -476,7 +476,8 @@ int picoquic_sendmsg(SOCKET_TYPE fd,
                 cmsg_2->cmsg_level = IPPROTO_IPV6;
                 cmsg_2->cmsg_type = IPV6_DONTFRAG;
                 cmsg_2->cmsg_len = CMSG_LEN(sizeof(int));
-                *((int *)CMSG_DATA(cmsg_2)) = val;
+//                *((int *)CMSG_DATA(cmsg_2)) = val;
```
bkchr: This line can be removed?

picoquictest/stresstest.c
```diff
@@ -0,0 +1,875 @@
...
+                        size_t received = ctx->data_received_on_stream[bidir_id] + length;
+                        if (ctx->data_received_on_stream[bidir_id] < PICOQUIC_STRESS_MINIMAL_QUERY_SIZE) {
+                            int processed = (int) length;
+                            if (received >= PICOQUIC_STRESS_MINIMAL_QUERY_SIZE) {
+                                processed = (int) received - PICOQUIC_STRESS_MINIMAL_QUERY_SIZE;
+                            }
+
+                            for (int i = 0; i < processed; i++) {
+                                ctx->data_sum_of_stream[bidir_id] =
+                                    ctx->data_sum_of_stream[bidir_id] * 101 + bytes[i];
```
bkchr: I don't know what you are doing here, why this multiplication with `101`? Could you maybe add a comment what all this is doing?

picoquic/sender.c
huitema: That test was checking against the sending limits imposed by the peer. But that is unnecessary, and in fact it was blocking the stress test. It should be OK to just queue data on streams, and then let the transport dequeue the data when the peer opens the flow control limits.

Removing the ifdef.

picoquic/picosocks.c
huitema: Yes. It was causing a weird message about pointer mappings in  gcc, so I replaced the mapping by a memcpy.

picoquictest/stresstest.c
huitema: Or initialize the field before testing it... Thanks for finding that.

picoquictest/stresstest.c
```diff
@@ -825,7 +825,7 @@ int stress_test()
 #else
             "certs/cert.pem", "certs/key.pem",
 #endif
-            PICOQUIC_TEST_ALPN, stress_server_callback, (void*)&stress_ctx, NULL, NULL, NULL,
+            PICOQUIC_TEST_ALPN, stress_server_callback, NULL, NULL, NULL, NULL,
```
huitema: Wow. Yes, you are right, it does not belong there. I wonder why the test was actually working on Windows...

huitema: Wow. Yes, you are right, it does not belong there. I wonder why the test was actually working on Windows...

huitema:
Just sent a commit addressing the review comments.

picoquictest/stresstest.c
huitema: I did add a comment. Is that OK?

huitema:
Checking this in, as all comments appear addressed + the bugs fixed here are causing the test server to crash.

PR #208, opened 2018-06-22
Remove extra initialization?

martinduke:
I believe this initialization does nothing, and possibly leaks memory.

### DIFF
--- a/picoquic/tls_api.c
+++ b/picoquic/tls_api.c
@@ -1058,7 +1058,7 @@ void * picoquic_pn_enc_create(
 {
     int ret = 0;
     uint8_t key[256]; 
-    ptls_cipher_context_t *pn_enc = ptls_cipher_new(aead->ctr_cipher, 1, key);
+    ptls_cipher_context_t *pn_enc;
 
     /*
      * Derive the key by extending the secret for PN encryption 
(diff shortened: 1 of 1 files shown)


bkchr:
The tests do not work anymore, with your patch. I'm unsure why, but that should not happen. I'm still not quite sure what the removed functions does.

martinduke:
OK, If there's a purpose to the double-call that I don't see, please drop the PR. By inspection, it seemed like a bug.

huitema:
Please hold on for a few days. This is a good patch, but I would rather not make any change until the tests are working. The reason they are breaking is probably unrelated (issue #209), but then the existing issue may be masking whatever problem there might be in the patch.

picoquic/tls_api.c
```diff
@@ -1058,7 +1058,7 @@ void * picoquic_pn_enc_create(
 {
     int ret = 0;
     uint8_t key[256]; 
-    ptls_cipher_context_t *pn_enc = ptls_cipher_new(aead->ctr_cipher, 1, key);
+    ptls_cipher_context_t *pn_enc;
```
huitema: You need to set pn_sec = NULL. Otherwise, if picoquic_hkdf_expand_label fails, the function returns an unitialized pointer.

huitema:
Checking this after updating the code.

PR #215, opened 2018-07-05
Draft 13

huitema:
Major changes coming with the implementation of draft 13. Biggest changes are in the TLS API (tls_api.c), with big impact on sender logic (sender.c) and receiver logic (packet.c), as well as several new frame types (frames.c).


picoquic/logger.c
```diff
@@ -242,64 +245,64 @@ char const* picoquic_log_ptype_name(picoquic_packet_type_enum ptype)
...
-        case picoquic_frame_type_path_challenge:
-            frame_name = "path_challenge";
-            break;
-        case picoquic_frame_type_path_response:
-            frame_name = "path_response";
-            break;
-        default: 
-            break;
-        }
+
```
igorlord: What happened to naming stream frames?

picoquic/logger.c
```diff
@@ -577,60 +580,44 @@ size_t picoquic_log_connection_close_frame(FILE* F, uint8_t* bytes, size_t bytes
...
-        error_code, (uint16_t)string_length, (uint16_t)string_length);
-    if (byte_index + string_length > bytes_max) {
-        fprintf(F, "    Malformed APPLICATION CLOSE, requires %d bytes out of %d\n",
-            (int)(byte_index + string_length), (int)bytes_max);
-        byte_index = bytes_max;
-    } else {
-        /* TODO: print the UTF8 string */
-        byte_index += (size_t)string_length;
-    }
+size_t picoquic_log_application_close_frame(FILE* F, uint8_t* bytes, size_t bytes_max)
```
igorlord: You may want to leave a comment that this "generic" close frame is going away (not there in Editor draft anymore).

huitema:
This is not ready yet. Still missing:

* separate sequence numbers, ACK, retransmit for initial/handshake/application packets
* new definition of RETRY packet and handling of cookies
* debug client authentication in the new logic
* debug HRR in the new logic
* repeat 0-RTT packets during handshake if they contain crypto stream frames
* coalesce packets
* new transport parameters
* ACK-ECN

We may want to check-in with only some of these fixes.

picoquic/frames.c
```diff
@@ -1506,25 +1669,27 @@ int picoquic_prepare_connection_close_frame(picoquic_cnx_t* cnx,
...
         *consumed = 0;
         ret = PICOQUIC_ERROR_FRAME_BUFFER_TOO_SMALL;
     }
 
     return ret;
 }
 
-int picoquic_decode_connection_close_frame(picoquic_cnx_t* cnx, uint8_t* bytes,
-    size_t bytes_max, size_t* consumed)
+int picoquic_decode_generic_close_frame(picoquic_cnx_t* cnx, uint8_t* bytes,
```
igorlord: Same comment as in logging -- the generic thing is short-lived.

(rathole: seems weird that we replicate frame decoding logic twice -- once here and once in logging.)

picoquic/frames.c
```diff
@@ -1533,21 +1698,37 @@ int picoquic_decode_connection_close_frame(picoquic_cnx_t* cnx, uint8_t* bytes,
 
     if (l1 == 0 || byte_index > bytes_max || bytes_max < 4) {
         ret = picoquic_connection_error(cnx,
-            PICOQUIC_TRANSPORT_FRAME_ERROR(picoquic_frame_type_connection_close));
+            PICOQUIC_TRANSPORT_FRAME_ERROR(ftype));
```
igorlord: Is the error PICOQUIC_TRANSPORT_FRAME_ERROR returns a valid QUIC error per spec?

picoquic/frames.c
```diff
@@ -1975,19 +2135,26 @@ int picoquic_decode_frames(picoquic_cnx_t* cnx, uint8_t* bytes,
...
+                byte_index += consumed;
+            }
         } else if (first_byte == picoquic_frame_type_ack) {
             ret = picoquic_decode_ack_frame(cnx, bytes + byte_index,
-                bytes_max - byte_index, &consumed, current_time, restricted);
+                bytes_max - byte_index, &consumed, current_time, epoch);
             byte_index += consumed;
         } else {
-            if (restricted && first_byte != picoquic_frame_type_padding 
+            if (epoch != 1 && epoch != 3 && first_byte != picoquic_frame_type_padding
```
igorlord: Would an enum be better than magic numbers for epoch?

picoquic/logger.c
huitema: The "generic" name is used for the closing code common to application, connection and crypto close messages.

picoquic/frames.c
huitema: Yes. Should find a way to unify that. And "generic" is not a frame type, just a name for common code.

picoquic/frames.c
huitema: It is defined in the quic spec as a range, with the final 8 bits encoding the frame type. Or at least it was...

picoquic/frames.c
huitema: Yes, probably, more mnemonic. I was lazy.

picoquic/frames.c
igorlord: Right.  In draft-13, both frames are identical.  However, in Editor's draft they have different fields already. That was my comment.  Nothing to do here for now, of course, since this is draft-13 support.

Issue #212, opened 2018-06-26
Refactoring frame skip and decode function interfaces


igorlord:
As stated in #199, the current interface to frame skip and decode functions makes the code too verbose.  We have options.

The proposed interface #199 (see below) was deemed too complex.
```
err <- skip_fn(uint8_t** pbytes, uint8_t* max_bytes)
err <- decode_fn(picoquic_cnx_t* cnx, uint8_t** pbytes, uint8_t* max_bytes)
```

Here is another proposal that is also concise.  Here, we return new bytes.  Errors are indicated by return of NULL (no separate error code).  Besides being concise, it ensures that processing will NOT continue after a failure.
```
uint8_t* <- skip_fn(uint8_t* bytes, uint8_t* max_bytes)
uint8_t* <- decode_fn(picoquic_cnx_t* cnx, uint8_t* bytes, uint8_t* max_bytes)
```

I'd appreciate feedback before I go ahead and create a PR to change a bunch of code.

huitema:
I am just boarding a plane right now, will come back to you after landing tonight

igorlord:
Bon voyage!

huitema:
I like the proposal -- if ((x == decode(x, xmax, &err_detail,...))==NULL) { process error}.
I think it is a reasonable compromise between speed and reliability.

igorlord:
Thanks.

One thing, I looked at all decode() and skip() functions, and I could not find the need for err_detail. Everywhere decode() sees an error, it just calls protocol error and returns the same error code. Skip just returns -1. There seems to be no informational content in err. That's why my proposal did not have err or err_detail.

(I am driving to NY this weekend, but next week I should have some time for this.)

igorlord:
Closing via #222

PR #222, opened 2018-07-07
Refactor decode and skip

igorlord:
This is a large PR.

# Purpose

I am trying to clean up and simplify a lot of code to make the semantics of frame operations easier to read and understand and, consequently, easier to keep bug-free.

This is just the first step in this direction.  I've only changed piciquic_decode_* and piciquic_skip_* functions.

Each function and its helpers check their own byte constraints.

I have also changed the internal frame decoding and skipping interface.  See #212 .
This allowed me to save a lot of code and improve readability by removing extraneous bookkeeping.

For now, all changes are limited to frame.c (and a few tests that reach into frame.c directly).

# Testing

The current tests, including CI tests work.  (Actually "request_client_authentication" also works.)

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -24,6 +24,129 @@
 #include <stdlib.h>
 #include <string.h>
 
+/* ****************************************************
+ * Frames private declarations
+ * ****************************************************/
+
+static const size_t challenge_length = 8;
+
+
+/**
+ * Frame decoder function
+ * Inputs:
+ *   cnx       - [in/out] picoquic Context
+ *   bytes     - [in]     pointer to the beginning of the frame (frame type)
+ *   bytes_max - [in]     pointer to the end of the packet (one past the last byte)
+ * Returns:
+ *   Pointer to the data following the end of this frame, if the frame has been decoded successfully;
+ *   or NULL if, decoding failed (in which case, picoquic_connection_error has been called).
+ */
+typedef uint8_t* (*decode_frame_fn)(picoquic_cnx_t* cnx, uint8_t* bytes, const uint8_t* bytes_max);
+
+/**
+ * Frame skip function
+ * Inputs:
+ *   bytes     - [in]     pointer to the beginning of the frame (frame type)
+ *   bytes_max - [in]     pointer to the end of the packet (one past the last byte)
+ * Returns:
+ *   Pointer to the data following the end of this frame, if the frame has been skipped successfully;
+ *   or NULL if, skipping failed.
+ */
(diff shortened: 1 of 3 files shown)


picoquic/frames.c
```diff
@@ -24,6 +24,129 @@
...
+        }
+
+        *n64 = v;
+    } else {
+        bytes = NULL;
+    }
+
+    return bytes;
+}
+
```
huitema: TODO: consider updating the varint unit test to also verify this "varint decode" function. Maybe this requires moving this to "intformat.c", or at least publishing the function signature so it is available for tests.

picoquic/frames.c
```diff
@@ -24,6 +24,129 @@
...
+{
+    if (bytes + sizeof(*n) <= bytes_max) {
+        *n = PICOPARSE_16(bytes);
+        bytes += sizeof(*n);
+    } else {
+        bytes = NULL;
+    }
+    return bytes;
+}
+
```
huitema: Same remark as varint -- less important as this is trivial, but having an unit tests would be nice.

picoquic/frames.c
```diff
@@ -24,6 +24,129 @@
...
+{
+    if (bytes + sizeof(*n) <= bytes_max) {
+        *n = PICOPARSE_64(bytes);
+        bytes += sizeof(*n);
+    } else {
+        bytes = NULL;
+    }
+    return bytes;
+}
+
```
huitema: I know this is trivial, but still, would be nice if it was published in .h for general availability and maybe unit test.

picoquic/frames.c
```diff
@@ -598,25 +665,23 @@ int picoquic_stream_network_input(picoquic_cnx_t* cnx, uint64_t stream_id,
...
+uint8_t* picoquic_decode_stream_frame(picoquic_cnx_t* cnx, uint8_t* bytes, const uint8_t* bytes_max, uint64_t current_time)
 {
-    int ret;
-    int fin = 0;
     uint64_t stream_id;
-    size_t data_length;
+    size_t   data_length;
     uint64_t offset;
+    int      fin;
+    size_t   consumed;
```
huitema: For code formatting, we have a specific cmake target. See PR #66 for details, but the idea is to use clang-format to automatically reset indents, comments, block ends, etc. we should consider running that after the current series of updates.

huitema:
I merged this code quickly because it somewhat conflicts with the work on multiple packet number sequences, part of draft 13. As part of that, some of the decoding functions need an extra parameter for the current packet context -- e.g., acks are only affecting a specific context. Doing the manual merge with my code now.

PR #224, opened 2018-07-07
Three seqs

huitema:
Implement the split into three packet number spaces, per draft 13. Fairly invasive.
Still need to implement other parts of draft-13, in particular ACK-ECN, retry cookies, and new transport parameters.

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -998,9 +998,10 @@ int picoquic_parse_ack_header(uint8_t const* bytes, size_t bytes_max,
 }
 
 void picoquic_check_spurious_retransmission(picoquic_cnx_t* cnx,
-    uint64_t start_of_range, uint64_t end_of_range, uint64_t current_time)
+    uint64_t start_of_range, uint64_t end_of_range, uint64_t current_time,
+    picoquic_packet_context_enum pc)
 {
-    picoquic_packet* p = cnx->retransmitted_newest;
+    picoquic_packet* p = cnx->pkt_ctx[pc].retransmitted_newest;
 
     while (p != NULL) {
         picoquic_packet* should_delete = NULL;
@@ -1008,8 +1009,8 @@ void picoquic_check_spurious_retransmission(picoquic_cnx_t* cnx,
         if (p->sequence_number >= start_of_range && p->sequence_number <= end_of_range) {
 
             uint64_t max_spurious_rtt = current_time - p->send_time;
-            uint64_t max_reorder_delay = cnx->latest_time_acknowledged - p->send_time;
-            uint64_t max_reorder_gap = cnx->highest_acknowledged - p->sequence_number;
+            uint64_t max_reorder_delay = cnx->pkt_ctx[pc].latest_time_acknowledged - p->send_time;
+            uint64_t max_reorder_gap = cnx->pkt_ctx[pc].highest_acknowledged - p->sequence_number;
             picoquic_path_t * old_path = p->send_path;
 
             if (p->length + p->checksum_overhead > old_path->send_mtu) {
@@ -1034,7 +1035,7 @@ void picoquic_check_spurious_retransmission(picoquic_cnx_t* cnx,
 
             cnx->nb_spurious++;
             should_delete = p;
-        } else if (p->send_time + PICOQUIC_SPURIOUS_RETRANSMIT_DELAY_MAX < cnx->latest_time_acknowledged) {
+        } else if (p->send_time + PICOQUIC_SPURIOUS_RETRANSMIT_DELAY_MAX < cnx->pkt_ctx[pc].latest_time_acknowledged) {
             should_delete = p;
         }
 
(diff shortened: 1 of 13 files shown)


picoquic/sender.c
```diff
@@ -1011,36 +1026,39 @@ static void picoquic_cnx_set_next_wake_time_init(picoquic_cnx_t* cnx, uint64_t c
...
+                        }
                     }
                 }
             }
+        }
+    }
 
+    /* Consider path challenges */
+    if (path_x->challenge_verified == 0) {
+        uint64_t next_challenge_time = path_x->challenge_time + path_x->retransmit_timer;
```
igorlord: This is perfectly fine, though I find nesting levels slow me down, when I am reading code.  I would had prefered the same as below.  This makes code shorter, fewer nesting levels, comparisons are always "left less than right".  Being more straightforward, I bet that if there were a logical mistake in this, it would become more immediately apparent.

```
if (next_challenge_time <= current_time) {
    next_time = current_time;
} else if (next_challenge_time < next_time) {
    next_time = next_challenge_time;
}
```

picoquic/sender.c
```diff
@@ -1230,7 +1267,51 @@ picoquic_packet_type_enum picoquic_packet_type_from_epoch(int epoch)
...
+
+    *header_length = 0;
+
+    length = picoquic_retransmit_needed(cnx, pc, path_x, current_time, packet, 
+        &is_cleartext_mode, header_length);
+    
+    if (length == 0 && cnx->pkt_ctx[pc].ack_needed != 0) {
+        packet->ptype =
+            (pc == picoquic_packet_context_initial) ? picoquic_packet_initial :
+            ((pc == picoquic_packet_context_handshake) ? picoquic_packet_handshake :
```
igorlord: The extra () about (x?y:z) Seem unnecessary and actually detract from the idea that this is basically a switch statement.
This construct is very idiomatic and therefore clear:
```
x=a?x:
x=b?y:
    z;
```

picoquic/picoquic_internal.h
```diff
@@ -376,6 +378,35 @@ typedef struct st_picoquic_crypto_context_t {
...
+    picoquic_sack_item_t first_sack_item;
+    uint64_t time_stamp_largest_received;
+    uint64_t highest_ack_sent;
+    uint64_t highest_ack_time;
+    uint64_t ack_delay_local;
+
+    uint64_t nb_retransmit;
+    uint64_t latest_retransmit_time;
+    uint64_t highest_acknowledged;
+    uint64_t latest_time_acknowledged; /* time at which the highest acknowledged was sent */
```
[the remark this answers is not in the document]
huitema: It is the time at which the latest ack was received. I kept the name that was already in use in the single context mode, to make comparison between the old and new code easier.

picoquic/quicctx.c
```diff
@@ -106,6 +106,18 @@ static int picoquic_compare_cnx_waketime(void * v_cnxleft, void * v_cnxright) {
     return ret;
 }
 
+picoquic_packet_context_enum picoquic_context_from_epoch(int epoch)
```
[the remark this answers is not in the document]
huitema: I will change the epoch to an enum in a separate PR. It has to be consistent with the values used in picotls, where it is defined as size_t indexing an array[5].

picoquic/sender.c
huitema: Yes. This is not new code, but code moved from the previous "single context" implementation. I wanted to keep changes to a minimum -- debug was hard enough...

picoquic/sender.c
huitema: OK, if you like it better. But then you are relying on the parsing precedence in compilers, and I have been bitten by that before.

picoquic/frames.c
```diff
@@ -1035,18 +1035,19 @@ int picoquic_parse_ack_header(uint8_t const* bytes, size_t bytes_max,
 }
 
 void picoquic_check_spurious_retransmission(picoquic_cnx_t* cnx,
-    uint64_t start_of_range, uint64_t end_of_range, uint64_t current_time)
+    uint64_t start_of_range, uint64_t end_of_range, uint64_t current_time,
+    picoquic_packet_context_enum pc)
 {
-    picoquic_packet* p = cnx->retransmitted_newest;
+    picoquic_packet* p = cnx->pkt_ctx[pc].retransmitted_newest;
```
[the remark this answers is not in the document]
huitema: Yes, I was a bit lazy. Changed now.

picoquic/sender.c
huitema: But I changed it anyhow.

picoquic/picoquic_internal.h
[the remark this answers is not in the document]
igorlord: So, is it "latest ACK was received", as you say; or "highest acknowledged was sent", as the comment says?

picoquic/sender.c
igorlord: Yah, then I type "C" into Google, it learned enough to autocomplete for me: "C operator precedence". :)   Nowadays, you'd expect all compilers to get these right.

Once thing I remember is to always quote binary bit operators, because it is just a bug in the spec -- their precedence is with the logical operators (&&, ||) instead of the more natural binary arithmetic operators!

Other than this, I never quote anything inside ?: -- its precedence is super-low.

PR #226, opened 2018-07-09
Implement the retry token, and fix a number of issues 

huitema:
There were a number of issues in the parsing of packet headers and the handling of initial packets, so what started as just a few lines of change became much bigger.


huitema:
Merging now, as the tests pass and we are short on time. There will be a lot of bugs to fix during/after the interop.

PR #241, opened 2018-07-12
Coalesce multiple segments in a packet

huitema:
This PR implements the "packet coalescing" function. It required some changes in the sending logic, such as passing variable length "max segment size" to functions. It also required changing the the signature of the "picoquic_prepare_packet" API, removing the "packet" parameter which is now handled internally. On one hand, this is a long overdue simplification, and the application now longer has to manage the allocation or free of these data structures. On the other hand, it might affect existing implementation.

@bkchr: you should check the effect on the RUST API.


bkchr:
Looks good to me, I did not see anything that will break my bindings in rust.

PR #244, opened 2018-07-14
Root certificate setter

bkchr:


picoquictest/tls_api_test.c
```diff
@@ -2937,6 +2915,27 @@ int set_certificate_and_key_test()
...
+            BIO* bio_key = BIO_new_file(PICOQUIC_TEST_CERT_STORE, "rb");
+            /* Load cert and convert to DER */
+            X509* cert = PEM_read_bio_X509(bio_key, NULL, NULL, NULL);
+            int length = i2d_X509(cert, NULL);
+            unsigned char* cert_der = (unsigned char*)malloc(length);
+            unsigned char* tmp = cert_der;
+            i2d_X509(cert, &tmp);
+            X509_free(cert);
+            BIO_free(bio_key);
+
```
huitema: That's a lot of glue code. Should we have an API to set the cert root by file name?

bkchr: Hmm we could provide one, currently you can set the root cert by giving it as argument to the initial `picoquic_create` call.

PR #248, opened 2018-07-14
Introduce varint frame types. They do little for now.

igorlord:
As a performance optimization, the code treats frame types as a single octet.  If only checks for an extended frame type, if the octet did not match any of the single-byte frame types.

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -2256,11 +2256,15 @@ int picoquic_decode_frames(picoquic_cnx_t* cnx, uint8_t* bytes,
                 bytes = picoquic_decode_new_token_frame(cnx, bytes, bytes_max);
                 ack_needed = 1;
                 break;
-            default:
-                /* Not implemented yet! */
-                picoquic_connection_error(cnx, PICOQUIC_TRANSPORT_FRAME_FORMAT_ERROR);
-                bytes = NULL;
+            default: {
+                uint64_t frame_id64;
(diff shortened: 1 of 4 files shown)


picoquic/frames.c
```diff
@@ -2256,12 +2256,16 @@ int picoquic_decode_frames(picoquic_cnx_t* cnx, uint8_t* bytes,
...
                 ack_needed = 1;
                 break;
-            default:
-                /* Not implemented yet! */
-                picoquic_connection_error(cnx, PICOQUIC_TRANSPORT_FRAME_FORMAT_ERROR);
-                bytes = NULL;
+            default: {
+                uint64_t frame_id64;
+                if ((bytes = picoquic_frames_varint_decode(bytes, bytes_max, &frame_id64)) != NULL) {
+                    /* Not implemented yet! */
```
huitema: Don't we have a coding rule that frame types should be encoded compactly, e.g. encode type 10 as 0x0A, not 0x400A or 0x8000000A ?

igorlord: It is a good question.  Is this in the spec?

igorlord: Yes, it is in Section 5:
_To ensure simple and efficient implementations of frame parsing, a frame type MUST use the shortest possible encoding._
So it is ok to fail to decode.  An explicit error could work, but it is probably ok as is.

huitema: Theoretical for now, but this length verification needs to be fixed as part of the work on extensibilty.

PR #249, opened 2018-07-15
Simplify stream id tests

igorlord:
No more bit arithmetic for higher-level QUIC stream ID concepts.

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -220,16 +220,16 @@ picoquic_stream_head* picoquic_find_or_create_stream(picoquic_cnx_t* cnx, uint64
 
     if (stream == NULL) {
         /* Verify the stream ID control conditions */
-        int      parity     = cnx->client_mode     ? is_remote                      : is_remote ^ 1;
-        uint64_t max_stream = (stream_id & 2) == 0 ? cnx->max_stream_id_bidir_local : cnx->max_stream_id_unidir_local;
+        int      expect_client_stream = cnx->client_mode ^ is_remote;
+        uint64_t max_stream = IS_BIDIR_STREAM_ID(stream_id) ? cnx->max_stream_id_bidir_local : cnx->max_stream_id_unidir_local;
 
-        if ((stream_id & 1) != parity || stream_id > max_stream) {
+        if (IS_CLIENT_STREAM_ID(stream_id) != expect_client_stream || stream_id > max_stream) {
             picoquic_connection_error(cnx, PICOQUIC_TRANSPORT_STREAM_ID_ERROR);
 
         } else if ((stream = picoquic_create_stream(cnx, stream_id)) == NULL) {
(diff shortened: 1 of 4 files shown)


picoquic/frames.c
```diff
@@ -220,16 +220,16 @@ picoquic_stream_head* picoquic_find_or_create_stream(picoquic_cnx_t* cnx, uint64
...
     if (stream == NULL) {
         /* Verify the stream ID control conditions */
-        int      parity     = cnx->client_mode     ? is_remote                      : is_remote ^ 1;
-        uint64_t max_stream = (stream_id & 2) == 0 ? cnx->max_stream_id_bidir_local : cnx->max_stream_id_unidir_local;
+        int      expect_client_stream = cnx->client_mode ^ is_remote;
+        uint64_t max_stream = IS_BIDIR_STREAM_ID(stream_id) ? cnx->max_stream_id_bidir_local : cnx->max_stream_id_unidir_local;
 
-        if ((stream_id & 1) != parity || stream_id > max_stream) {
+        if (IS_CLIENT_STREAM_ID(stream_id) != expect_client_stream || stream_id > max_stream) {
             picoquic_connection_error(cnx, PICOQUIC_TRANSPORT_STREAM_ID_ERROR);
```
huitema: Can we change the name of this macro? You are not testing "IS_CLIENT_STREAM", but rather "IS_REMOTE_STREAM".

igorlord: IS_CLIENT_STREAM_ID tests stream id per stream id definition.

Low Bits	Stream Type
0x0	Client-Initiated, Bidirectional
0x1	Server-Initiated, Bidirectional
0x2	Client-Initiated, Unidirectional
0x3	Server-Initiated, Unidirectional

huitema: OK, but then you don't actually meet your goal of "no more bit twiddling" because of this statement:
```
expect_client_stream = cnx->client_mode ^ is_remote;
```
So maybe you need another macro to compute "expect_client_stream" without embedding the knowledge of how client_mode is encoded.

igorlord: :-) I was using XOR as a logical operation, not a bit operation.

huitema: So you have basically renamed my old "parity" bit. If you want to check it in as is, you can do it. But I wish you would also fix this parity/expect bit...

igorlord: > So you have basically renamed my old "parity" bit.
In this particular case, yes.  Since the purpose of this PR to name things in "quic-domain" language ("client stream", "bidir stream", etc), I renamed "parity", which was "bits-domain" language.

> But I wish you would also fix this parity/expect bit...
I am not in a rush to check in. I would love to fix everything that needs fixing.  I just do not understand, yet, what needs fixing about this.

huitema:
Resolved conflicts and merged.

Issue #206, opened 2018-06-16
Marked CPU increase when stress runs more than 2 minutes


huitema:
Running the stress test for various duration values shows that the execution time increases significantly when the stress runs more than 2 minutes, and grows linearly after that.

| simulated time (sec) | execution time (sec) |
|---------------------:|--------------------:|
| 120 | 9 |
| 180 | 38 |
| 240 | 68 |
| 300 | 100 |
| 360 | 127 |
|  420 | 157 |

Up to 2 minutes, the stress time is a small fraction of wall time. After that, each additional minute of simulation appears to add about 30 seconds of run time on an Intel I7 PC. 

Two minutes correspond to the default inactivity timer on the server. The most likely explanation is that during stress a queue of timed out connections builds up, and reaches its maximum size after 2 minutes. Some algorithm somewhere scales as O(number of active connections), and causes the CPU time to increase as the queue builds up.

huitema:
The splay tree was not a magic bullet. But there is something else happening. When running stress under debugger, I observe that the program memory increases almost linearly with time, from a few MB at the beginning of the run to 131 MB after 5 minutes. Yet the windows debugger does not complain about memory leaks, probably because all that memory is eventually freed when the quic contexts are deleted. This needs to be investigated.

huitema:
The program memory issue appears much improved after the recent changes to the picoquic_prepare_packet API -- it may well be that the stress test was leaking the "packet" structure. But there is still some memory increase, from 35MB to 37MB.

huitema:
The performance is also improved:

simulated time (sec) | execution time (sec) 
-- | --
120 | 6
180 | 23
240 | 41
300 | 57
360 | 76
420 | 93

We should still look at improved performances, but the real issue is the memory leakage.

PR #255, opened 2018-07-16
Fixes memory leaks

bkchr:
Especially the one in sender.c is very important!

### DIFF
--- a/picoquic/sender.c
+++ b/picoquic/sender.c
@@ -2142,7 +2142,7 @@ int picoquic_prepare_packet(picoquic_cnx_t* cnx,
             break;
         }
         else {
-            ret = picoquic_prepare_segment(cnx, path_x, packet, current_time, 
+            ret = picoquic_prepare_segment(cnx, path_x, packet, current_time,
                 send_buffer + *send_length, available, &segment_length);
 
             if (ret == 0) {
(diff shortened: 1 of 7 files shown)


picoquic/sender.c
```diff
@@ -2142,14 +2142,17 @@ int picoquic_prepare_packet(picoquic_cnx_t* cnx,
...
 
             if (ret == 0) {
                 *send_length += segment_length;
                 if (packet->length == 0 ||
                     packet->ptype == picoquic_packet_1rtt_protected_phi0 ||
                     packet->ptype == picoquic_packet_1rtt_protected_phi1) {
+                    if (packet->length == 0) {
+                        free(packet);
+                    }
                     break;
```
huitema: Weird. I am pretty sure I coded exactly that in my recent patch. But you are correct, it did not make it into the PR. Something went wrong. Thanks for fixing it.

picoquictest/tls_api_test.c
```diff
@@ -779,6 +779,8 @@ static int tls_api_one_sim_round(picoquic_test_tls_api_ctx_t* test_ctx,
                     /* useless test, but makes it easier to add a breakpoint under debugger */
                     ret = -1;
                 }
+
+                free(packet);
```
huitema: Isn't the data already freed at line 743?

bkchr: No, the packet is created in line 762

bkchr:
@huitema I added more memory leak fixes. Some of them are also in the stress test, maybe that helps you with https://github.com/private-octopus/picoquic/issues/206, to reduce the execution time.

PR #257, opened 2018-07-22
Rework handling of `RST_STREAM`

bkchr:
When `RST_STREAM` is sent, the stream can still send `STOP_SENDING`.
The relevant check is now in `picoquic_add_to_stream` that ensures when
`RST_STREAM` was sent or `STOP_SENDING` was received, the stream does not send
anymore data.

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -709,13 +709,10 @@ picoquic_stream_head* picoquic_find_ready_stream(picoquic_cnx_t* cnx)
 
     if (cnx->maxdata_remote > cnx->data_sent) {
         while (stream) {
-            if ((stream->stream_flags & picoquic_stream_flag_reset_sent) == 0 && ((stream->send_queue != NULL &&
-                stream->send_queue->length > stream->send_queue->offset &&
-                stream->sent_offset < stream->maxdata_remote) ||
-                ((stream->stream_flags & picoquic_stream_flag_fin_notified) != 0 &&
-                (stream->stream_flags & picoquic_stream_flag_fin_sent) == 0 &&
-                    (stream->sent_offset < stream->maxdata_remote)) ||
-                    (stream->stream_flags & picoquic_stream_flag_reset_requested) != 0 || ((stream->stream_flags & picoquic_stream_flag_stop_sending_requested) != 0 && (stream->stream_flags & picoquic_stream_flag_stop_sending_sent) == 0))) {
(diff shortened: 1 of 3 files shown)


picoquic/frames.c
```diff
@@ -703,19 +703,45 @@ uint8_t* picoquic_decode_stream_frame(picoquic_cnx_t* cnx, uint8_t* bytes, const
...
+}
+
+int stream_fin_notified(picoquic_stream_head* stream) {
+    return (stream->stream_flags & picoquic_stream_flag_fin_notified) != 0;
+}
+
+int stream_fin_sent(picoquic_stream_head* stream) {
+    return (stream->stream_flags & picoquic_stream_flag_fin_sent) != 0;
+}
+
```
huitema: This is very similar to the work that @igorlord did to "Simplify stream id tests" in PR #249. The difference is that Igor used macros instead of C functions. We can argue that compilers can inline the C functions just like they can with macros, and the there are no performance differences. But there is an impact on name spaces. The globally defined functions become visible in the picoquic lib, and their names can conflict with similar names used for example by the application that handles the streams. In contrast, the macros are defined in the "internal" header, are not visible by the application, and thus will not create name space collisions.

I think there are three options: use macros, use a picoquic_ prefix in the name of functions, or define the functions as "static". What do you prefer?

picoquic/picoquic_internal.h
```diff
@@ -775,6 +775,14 @@ void picoquic_queue_stateless_reset(picoquic_cnx_t* cnx,
...
 picoquic_misc_frame_header_t* picoquic_create_misc_frame(const uint8_t* bytes, size_t length);
 
+int stream_reset_sent(picoquic_stream_head* stream);
+int stream_reset_requested(picoquic_stream_head* stream);
+int stream_stop_sending_requested(picoquic_stream_head* stream);
+int stream_stop_sending_sent(picoquic_stream_head* stream);
+int stream_stop_sending_received(picoquic_stream_head* stream);
+int stream_fin_notified(picoquic_stream_head* stream);
+int stream_fin_sent(picoquic_stream_head* stream);
+
```
huitema: See previous remark on macros vs functions. You could easily replace these function declarations by macros.

bkchr:
Yeah, I will convert these functions to macros.

bkchr:
@huitema sorry for the delay, but now the changes are integrated.

bkchr:
@huitema could you restart ci? There was some weird error.

PR #258, opened 2018-07-25
header changes per QUIC PR 1498 + fix warnings

huitema:
PR 1498 simplifies encoding of the Initial and Retry packet headers. The code changes also include fixes for compilation warnings caused by recent PR.


huitema:
Merging, since the changes are really simple.

Issue #243, opened 2018-07-13
Failure when attempting to use chacha20


huitema:
If the client and server negotiate chacha20, the PN encryption breaks.

huitema:
This may be due to issue #245

huitema:
Verified that this does not repro. Reinstated support for ChaCha20 in PR #266.

PR #266, opened 2018-08-08
Reinstate support for chacha20 & poly1305

huitema:
Bug 243 does not repro, was probably a wrong diagnostic.


huitema:
Close issue #243

PR #268, opened 2018-08-08
Set the IPv4 don't fragment bit on packets larger than the minimum MTU

huitema:


bkchr:
I think you need to use `IP_PMTUDISC_DO` on Linux.

huitema:
@bkchr : OK, will set to _DO. I was a bit confused by the Linux documentation.

PR #269, opened 2018-08-12
Draft13 interop

huitema:
The draft-13 code was not tested for interop before. Now it is. Connections against nghttp2.org work, and all tests are passing. The major pain point was the separation of the old stream zero in 4 different streams, one per epoch.


huitema:
Merging now. Need to work next on the certificate validation, as the current procedure appears rather hard to deploy.

PR #284, opened 2018-08-16
Add frame type parameter to connection close frame

huitema:


huitema:
Checking in, before moving on to the other draft-14 changes.

PR #287, opened 2018-08-19
Fuzz test

huitema:
Add an API to the QUIC context for programming a "fuzz" callback, that fill be called just before a packet is encrypted and sent. This allows testing arbitrary fuzzing logics.
Create a "fuzz" variant of the stress test that uses a basic fuzzer. Make it accessible through the normal test program, e.g. "picoquic_ct -f 5" to run the test on 5 simulated minutes.
Fix bugs discovered when running the fuzz test for 5 minutes, including three hot loops conditions and one crash.

### DIFF
--- a/UnitTest1/unittest1.cpp
+++ b/UnitTest1/unittest1.cpp
@@ -604,5 +604,12 @@ namespace UnitTest1
 
             Assert::AreEqual(ret, 0);
         }
+
+        TEST_METHOD(fuzz)
+        {
+            int ret = fuzz_test();
+
+            Assert::AreEqual(ret, 0);
+        }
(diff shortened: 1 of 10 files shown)


huitema:
@bkchr You may want to check whether it makes sense to use the fuzzing callback API in Rust

huitema:
Found a fuzzer issue running the fuzzer for 20 simulated minutes. Crash if the returned length is shorter than the header length. Fixed.

huitema:
Found yet another bug when running fuzz longer -- discrepancy between wake time prediction and code when ack needed at epoch=1, leading to busy loop.

PR #290, opened 2018-08-21
Hunt leaks

huitema:
Debugged the memory leaks experienced in the stress and fuzz tests. There were two big packet leaks, and then a couple of small leaks tied to context management. There is still a residual leak, but we went from leaking 2.8MB in a stress session to leaking 140KB.

There are two causes for the residual leak. One is the initialization of OpenSSL global variables. I tried fixing using procedures like "OpenSSL_cleanup", using ref-counting to do it once for many QUIC contexts. But something is broken, because if one tries to re-open OpenSSL again in the same process, there is a crash.

The other cause is the management of the "verify certificate" call back. The current code is a bit messy, as the knowledge of the Picotls function based on OpenSSL has polluted the common path. It need sto be rewritten, but that is another PR.

### DIFF
--- a/picoquic/tls_api.c
+++ b/picoquic/tls_api.c
@@ -31,6 +31,7 @@
 #include <openssl/pem.h>
 #include <openssl/err.h>
 #include <openssl/engine.h>
+#include <openssl/conf.h>
 #include <stdio.h>
 #include <string.h>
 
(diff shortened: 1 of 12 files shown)


picoquic/tls_api.c
```diff
@@ -89,9 +89,37 @@ static void picoquic_init_openssl()
...
     }
 }
 
+static void picoquic_release_openssl()
+{
+
+#if 0
+    /* Invalidating the openssl release code for now, as trying multiple open/close in
+     * the same process causes a crash */
+    if (openssl_is_init > 0) {
```
bkchr: I would not do this. You are right that these are leaked resources, but they should be leaked only once per process.
If you really want to release this resources, we should use some sort of global destructor.

huitema: OK. There is in fact code in OpenSSL that attempts to release the resource on exit, so this may not be needed at all. In any case, I removed the ref counting and the attempt at automated release, and just added a global destructor to be called before exit. I only use it in the test program. It does free 37K out of the remaining 140K on the heap. After that, it seems that we are mostly left with file buffers for stdio, internal allocations for some system libraries, and possibly traces of openssl still left.

Most of the memory tests were done on Windows, because Visual Studio is a much nicer UI than valgrind. But we still need to see the result of valgrind on Linux.

huitema:
OK, I tried adding a global destructor but it does not compile on Linux. So I just removed it, following the advice of @bkchr

Issue #298, opened 2018-08-24
Fuzz fails on Linux/gccbut not Linux/clang


huitema:
The "gcc on Linux" variant of the fuzz test failed during the validation of a simple fix. There is no indication that the fix caused the failure -- the fix was related to the handling of retry packets, while the core dump indicate a failure on processing a stream input with end mark set.

The failure shows two different problems. Obviously, the failure itself should be examined and fixed. But maybe more importantly, we have to understand why the test failed with Linux/gcc, but did not fail with Linux/clang or on Windows with the VS 2017 compiler. Also, we have to understand why a change in a piece of code that is probably not even executed during the test caused the fuzz test to fail.

In particular, we need to verify that the random number draws and other sequencing operation behave in the same way with Linux/gcc and with other platforms.

huitema:
Fixed by PR #300

Issue #299, opened 2018-08-24
Need option to mute debug message when running stress


huitema:
Running error tests such as fuzz or even stress causes lots of error conditions (doh), each of which generates a debug message. This makes build logs almost unreadable. There should be an option to mute these messages, only keeping the most important.

huitema:
Fixed by PR #300, which created a "-n" option.

PR #300, opened 2018-08-25
Random tester

huitema:
The random generator for test was incorrectly using the "picoquic_public_random" function, which does not guarantee that the sequence of random numbers is repeatable. In addition, the public random generator was incorrectly initialized on 32 bit machines, leading to different values generated with clang and gcc.


huitema:
This fixes issue #298 and #299

PR #302, opened 2018-09-06
WIP: Add scan-build (clang-analyzer) to Travis and fix warning

alagoutte:
Add scan-build (clang-analyzer) to Travis

and try to fix also warning found by scan-build

### DIFF
--- a/picoquic/logger.c
+++ b/picoquic/logger.c
@@ -497,14 +497,12 @@ size_t picoquic_log_ack_frame(FILE* F, uint64_t cnx_id64, uint8_t* bytes, size_t
 
         if (byte_index >= bytes_max) {
             fprintf(F, "    Malformed ACK RANGE, %d blocks remain.\n", (int)num_block);
-            ret = -1;
             break;
         }
 
         size_t l_range = picoquic_varint_decode(bytes + byte_index, bytes_max - byte_index, &range);
         if (l_range == 0) {
             byte_index = bytes_max;
-            ret = -1;
             fprintf(F, "    Malformed ACK RANGE, requires %d bytes out of %d", (int)picoquic_varint_skip(bytes),
                 (int)(bytes_max - byte_index));
             break;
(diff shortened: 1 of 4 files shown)


picoquic/logger.c
```diff
@@ -658,9 +651,6 @@ size_t picoquic_log_generic_close_frame(FILE* F, uint8_t* bytes, size_t bytes_ma
         }
         if (ftype != picoquic_frame_type_connection_close || lf != 0) {
             l1 = picoquic_varint_decode(bytes + byte_index, bytes_max - byte_index, &string_length);
-            if (l1 == 0) {
-                required = byte_index + picoquic_varint_skip(bytes + byte_index);
-            }
         }
```
huitema: My code seems wrong, but I am not sure that this is the right fix. Let me take a further look.

huitema: We should keep this test, remove the unused reference to "required", and add the same kind of code as in the block above:
```
    if (l1 == 0) {
        byte_index = bytes_max;
    } else {
        byte_index += l1;
    }
```
The "required" variable was originally used in an error message, "only x bytes available when y bytes are required". Instead of removing the references, I would consider adding back the error message.

picoquic/quicctx.c
```diff
@@ -1206,7 +1206,7 @@ int picoquic_reset_cnx(picoquic_cnx_t* cnx, uint64_t current_time)
     }
 
     if (ret == 0) {
-        ret = picoquic_setup_initial_traffic_keys(cnx);
+        picoquic_setup_initial_traffic_keys(cnx);
     }
```
huitema: I see why the analyzer protests, but I am uneasy about not testing the return value. If initialization fails, we should somehow fail the call. Instead of removing the reference to the return code here, we should replace line 1215:
```
    ret = picoquic_tlscontext_create(cnx->quic, cnx, current_time);
```
by
```
    if (ret == 0) {
        ret = picoquic_tlscontext_create(cnx->quic, cnx, current_time);
    }
```

picoquic/logger.c
huitema: Sorry, wrong suggestion. Your fixes here are fine.

huitema:
@alagoutte I propose proceeding in three steps. First check-in this PR as is. Then do a second PR that fixes all the issues found in scan. Then a 3rd PR to update the travis script to fail if there is a scan failure. Do you agree?

PR #306, opened 2018-09-14
Remove not working IPv4 DONT FRAG attempt on Linux

huitema:
Trying to use the equivalent of:
```
s = setsockopt(sockfd, IPPROTO_IP, IP_MTU_DISCOVER, &on, sizeof(on));
```
on individual packets through the "sendmsg" API on Linux causes EINVAL errors. It works for IPv6, and it works for IPv4 on WIndows and BSD, but not on Linux.

A possible fix is to set the option for the whole socket on Linux, but that would prevent working on some networks.

### DIFF
--- a/picoquic/picosocks.c
+++ b/picoquic/picosocks.c
@@ -496,6 +496,7 @@ int picoquic_sendmsg(SOCKET_TYPE fd,
         }
 #endif
 
+#if 0
 #if defined(IP_PMTUDISC_DO) || defined(IP_DONTFRAG)
         if (addr_from->sa_family == AF_INET && length > PICOQUIC_INITIAL_MTU_IPV4) {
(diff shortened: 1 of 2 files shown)


igorlord:
Yes, Linux will not allow you to set this per-datagram.  You need to set this per-socket.

igorlord:
Of course, you can always write RAW sockets (IP raw, not Packet raw).  But that requires you to have root (at at least CAP_NET) privileges.

It looks like selecting the value of a DF flag (and TOS/DS field for ECN) from userspace is a useful feature.  If you think it is useful, I can get someone on my team to try put it into the upstream Linux kernel.

huitema:
If you could do that, that would be great. It would also be parity with other systems, and with what is available on Linux on IPv6.

Issue #301, opened 2018-09-06
Log of retry packet


huitema:
Retry packet is not found in the server's logs. This makes debugging harder.

PR #307, opened 2018-09-14
Add logging of retry and negotiation packets in server log.

huitema:


huitema:
This closes issue #301

PR #308, opened 2018-09-15
Implement connection migration

huitema:
This massive PR implements connection migration. It includes the management of connection IDs, the handling of paths, sending probes, and of course lots of tests.


huitema:
Still on my TO-DO list is simulation of connection migration on the client. Would be nice for demonstration during the interop, but no other implementation claims to support migration.

huitema:
Checking this in, in order to facilitate interop tests during the QUIC meeting.

PR #313, opened 2018-09-18
Add an option for key logging (draft -14)

Lekensteyn:
The log_labels logic is based on setup_traffic_protection in picotls.c.
Enables QUIC decryption with Wireshark v2.9.0rc0-1850-g2fd42045f5.

The file handle will be leaked on exit, since this is a debug feature I
am not too worried about that.
___
This was tested on Linux where setting environment variables is rather easy. I can also replace it with a command-line argument if you prefer that. Do you think something like this would be nice to have?

(This feature was requested by @alagoutte)

### DIFF
--- a/picoquicfirst/picoquicdemo.c
+++ b/picoquicfirst/picoquicdemo.c
@@ -144,6 +144,30 @@ static char* strip_endofline(char* buf, size_t bufmax, char const* line)
     return buf;
 }
 
+static void picoquic_set_key_log_file_from_env(picoquic_quic_t* quic)
+{
+    const char* keylog_filename;
+    FILE* F = NULL;
+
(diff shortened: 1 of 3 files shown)


alagoutte:
LGTM

Thanks Peter

picoquicfirst/picoquicdemo.c
```diff
@@ -899,6 +925,8 @@ int quic_client(const char* ip_address_text, int server_port, const char * sni,
 
             PICOQUIC_SET_LOG(qclient, F_log);
 
+            picoquic_set_key_log_file_from_env(qclient);
+
             if (sni == NULL) {
```
huitema: I wonder whether the client should display some text "Keys are logged on file %s", so the operators knows what is happening.

PR #315, opened 2018-09-18
CI for Windows: integration with AppVeyor

Lekensteyn:
Linux builds are already covered by Travis, but there was no CI for Windows. This PR tries to rectify that. See individual commits for details. Example passing build: https://ci.appveyor.com/project/Lekensteyn/picoquic/build/build23

To enable AppVeyor, you need to login at https://ci.appveyor.com/ with your Github account and enable this repo. After that, go to https://ci.appveyor.com/project/private-octopus/picoquic/settings and check *Skip branches without appveyor.yml* (optional, but recommended).

At first I tried to integrate with CMake which has better support for discovering OpenSSL automatically, but since the number of Windows developers using CMake probably approach 0, I decided to use the existing VS project (where the number of users is at least 1).

### DIFF
--- a/UnitTest1/unittest1.cpp
+++ b/UnitTest1/unittest1.cpp
@@ -30,6 +30,11 @@ namespace UnitTest1
 	TEST_CLASS(UnitTest1)
 	{
 	public:
+        TEST_CLASS_INITIALIZE(setup) {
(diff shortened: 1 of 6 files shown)


huitema:
Very cool. I did authorize appveyor. We shall see how it works next..

Issue #340, opened 2018-10-03
Seg fault on close connection in initial


kenmcmil:
commit: cb34de2

Seg fault occurs when server receives an initial packet with client hello handshake message followed immediately by an initial packet with a connection close frame. It appears that the server tries to send a 1rtt propected packet as a response before the packet number encoder for this packet type has been initialized. Stack trace follows:

```
#0  0x000055555557a7bd in picoquic_pn_iv_size (pn_enc=0x0) at /home/mcmillan/projects/picoquic/picoquic/tls_api.c:1345
#1  0x000055555557227f in picoquic_protect_packet (cnx=0x5555557cb8e0, ptype=picoquic_packet_1rtt_protected_phi0, bytes=0x5555557cc67d "", sequence_number=0, length=18, header_length=13, send_buffer=0x7fffffffdbf0 "0", send_buffer_max=1252, aead_context=0x0, pn_enc=0x0) at /home/mcmillan/projects/picoquic/picoquic/sender.c:395
#2  0x0000555555572c4a in picoquic_finalize_and_protect_packet (cnx=0x5555557cb8e0, packet=0x5555557cc640, ret=0, length=18, header_length=13, checksum_overhead=16, send_length=0x7fffffffd2e8, send_buffer=0x7fffffffdbf0 "0", send_buffer_max=1252, path_x=0x5555557c7be0, current_time=1538606342246257) at /home/mcmillan/projects/picoquic/picoquic/sender.c:659
#3  0x000055555557617b in picoquic_prepare_packet_closing (cnx=0x5555557cb8e0, path_x=0x5555557c7be0, packet=0x5555557cc640, current_time=1538606342246257, send_buffer=0x7fffffffdbf0 "0", send_buffer_max=1252, send_length=0x7fffffffd2e8) at /home/mcmillan/projects/picoquic/picoquic/sender.c:2024
#4  0x0000555555576d68 in picoquic_prepare_segment (cnx=0x5555557cb8e0, path_x=0x5555557c7be0, packet=0x5555557cc640, current_time=1538606342246257, send_buffer=0x7fffffffdbf0 "0", send_buffer_max=1536, send_length=0x7fffffffd2e8) at /home/mcmillan/projects/picoquic/picoquic/sender.c:2314
[...]

PR #345, opened 2018-10-05
Fixes for draft-15 interop

huitema:
Support the change in new connection ID frame in draft-15
Support the retire connection ID frames per draft 15
Handle server busy flag, and add test.
Add tests to verify that issue #340 is fixed.


huitema:
Also enforces a couple of small issues, like only one retry per connection, no empty payload, always initial CID for 0-RTT, no VN generated for short packets.

PR #357, opened 2018-10-13
Implement SQR (spin+loss) and make it available as version PCQ1

ferrieux:
This is to demo a shortcoming of the API, which lacks some modularity: all spin variants' state fields still coexist in generic structures (path and packet header). I also had to extend spinbit_outgoing to receive a bit of context. But it does the job. When in Rome...

Note there are two ways forward: modularize fully, or do away with the API. Your call.


picoquic/picoquic_internal.h
```diff
@@ -805,7 +813,8 @@ uint32_t picoquic_create_packet_header(
     picoquic_connection_id_t * srce_cnxid,
     uint8_t* bytes,
     uint32_t * pn_offset,
-    uint32_t * pn_length);
+    uint32_t * pn_length,
+    int contains_retrans);
```
huitema: I don't think that this is right. The fact that a packet "contains a retransmission" is an artifact of the current code organization. Also, in a multi-path environment, a loss on one path may be corrected by a retransmission on another path -- and that can also happen in a migration scenario.

Bottom line, no, we should not add a "contains retransmission" parameter in the "create packet header" call.

picoquic/picoquic_internal.h
```diff
@@ -844,7 +853,8 @@ uint32_t picoquic_protect_packet(picoquic_cnx_t* cnx,
     picoquic_connection_id_t * local_cnxid,
     uint32_t length, uint32_t header_length,
     uint8_t* send_buffer, uint32_t send_buffer_max,
-    void * aead_context, void* pn_enc);
+    void * aead_context, void* pn_enc,
+    int contains_retrans);
```
huitema: Same remark. Please find another way to carry loss indications. For example, it would be less intrusive to have a loss detected flag per path, that would be reset when your loss detected bit is sent, or something like that.

picoquic/sender.c
```diff
@@ -258,7 +259,7 @@ uint32_t picoquic_create_packet_header(
         bytes[length++] = (K | C | spin_bit | spin_vec);
 #else
         length = 0;
-        bytes[length++] = (K | C | picoquic_spin_function_table[picoquic_supported_versions[cnx->version_index].spinbit_version].spinbit_outgoing(cnx));
+        bytes[length++] = (K | C | picoquic_spin_function_table[picoquic_supported_versions[cnx->version_index].spinbit_version].spinbit_outgoing(cnx,(contains_retrans?PICOQUIC_SPINFLAGS_RETRANS:0)));
 #endif
```
huitema: Again, no, please find another way.

picoquic/sender.c
```diff
@@ -402,7 +403,8 @@ uint32_t picoquic_protect_packet(picoquic_cnx_t* cnx,
     picoquic_connection_id_t * local_cnxid,
     uint32_t length, uint32_t header_length,
     uint8_t* send_buffer, uint32_t send_buffer_max,
-    void * aead_context, void* pn_enc)
+    void * aead_context, void* pn_enc,
+    int contains_retrans)
 {
```
huitema: This is wrong. We go to great length to ensure that retransmission is a "frame level" issue, not a "packet level" issue. You may find that the current code does that, but it can change at any time. For example, instead of triggering an actual packet retransmission, the code could simply add the frames from the lost packet to the "misc frames" queue. In fact, we will have to do that if PMTU discover that the packet size is now shorter than expected. There will not be any retransmitted packet, but instead the lost stream frame will be split in two "misc" frames, retransmitted in independent packets.

picoquic/sender.c
```diff
@@ -650,31 +652,31 @@ void picoquic_finalize_and_protect_packet(picoquic_cnx_t *cnx, picoquic_packet_t
...
-                send_buffer, send_buffer_max, cnx->crypto_context[1].aead_encrypt, cnx->crypto_context[1].pn_enc);
+                send_buffer, send_buffer_max, cnx->crypto_context[1].aead_encrypt, cnx->crypto_context[1].pn_enc, packet->contains_retrans);
             break;
         case picoquic_packet_1rtt_protected:
             length = picoquic_protect_packet(cnx, packet->ptype, packet->bytes, packet->sequence_number,
                 remote_cnxid, local_cnxid,
                 length, header_length,
-                send_buffer, send_buffer_max, cnx->crypto_context[3].aead_encrypt, cnx->crypto_context[3].pn_enc);
+                send_buffer, send_buffer_max, cnx->crypto_context[3].aead_encrypt, cnx->crypto_context[3].pn_enc, packet->contains_retrans);
             break;
```
huitema: Same comment at all these places: carrying bits in the calling stack is the wrong way to do this.

picoquic/sender.c
```diff
@@ -2225,6 +2227,7 @@ int picoquic_prepare_packet_ready(picoquic_cnx_t* cnx, picoquic_path_t * path_x,
 
         if (ret == 0 && retransmit_possible &&
             (length = picoquic_retransmit_needed(cnx, pc, path_x, current_time, packet, send_buffer_min_max, &is_cleartext_mode, &header_length)) > 0) {
+            packet->contains_retrans = 1; 
             /* Set the new checksum length */
```
huitema: Please don't do that. Instead, you may want to call something like "picoquic_loss_detected(cnx, pc, path_x)" that will start some accounting in your "spin bit QR" context. (Note that there is a separate context for each path).

huitema: Also, that call should be inside the `picoquic_retransmit_needed()` function, not in the calling function.

picoquic/spinbit.c
```diff
@@ -34,9 +34,10 @@ void picoquic_spinbit_basic_incoming(picoquic_cnx_t * cnx, picoquic_path_t * pat
     path_x->current_spin = ph->spin ^ cnx->client_mode;
 }
 
-uint8_t picoquic_spinbit_basic_outgoing(picoquic_cnx_t * cnx)
+uint8_t picoquic_spinbit_basic_outgoing(picoquic_cnx_t * cnx,uint32_t flags)
 {
```
huitema: I would rather not do that. How about having the logical content of these flags inside the "spin bit" context?

picoquic/spinbit.c
```diff
@@ -77,6 +79,26 @@ uint8_t picoquic_spinbit_vec_outgoing(picoquic_cnx_t * cnx)
...
+uint8_t picoquic_spinbit_sqr_outgoing(picoquic_cnx_t * cnx,uint32_t flags)
+{
+  picoquic_path_t *pa = cnx->path[0];
+  pa->loss_q_index++;
+  if (pa->loss_q_index>=PICOQUIC_LOSS_Q_PERIOD) {
+    pa->loss_q_index=0;
+    pa->loss_q=(1-pa->loss_q);
+  }
+  return (uint8_t)((pa->current_spin << 2)|(pa->loss_q << 1)|((flags&PICOQUIC_SPINFLAGS_RETRANS)!=0));
+}
```
huitema: You can achieve the same result by having an element in the path context:
```
unsigned int loss_detected:1;
```
Then do:
```
uint8_t spinbyte_sqr = (uint8_t)((pa->current_spin << 2)|(pa->loss_q << 1)|(pa->loss_detected);
pa->loss_detected = 0;
return spinbyte_sqr;
```

picoquictest/parseheadertest.c
```diff
@@ -391,7 +391,7 @@ int parseheadertest()
             cnx_10->path[0]->remote_cnxid = test_cnxid_r10;
         }
         header_length = picoquic_create_packet_header(cnx_10, test_entries[i].ph->ptype,
-            test_entries[i].ph->pn, &cnx_10->path[0]->remote_cnxid, &cnx_10->path[0]->local_cnxid, packet, &pn_offset, &pn_length);
+            test_entries[i].ph->pn, &cnx_10->path[0]->remote_cnxid, &cnx_10->path[0]->local_cnxid, packet, &pn_offset, &pn_length,0);
         picoquic_update_payload_length(packet, pn_offset, pn_offset, pn_offset +
```
huitema: This will not be necessary if you do not change the signature of the packet header function.

huitema:
The "transport param" test fails because the transport parameters include the list of versions supported by the server. If you add a supported version, you need to update the test vectors describing the "server parameters". You will also need after that to update the reference text for the transport param logger test.

picoquic/spinbit.c
ferrieux: OK, done as suggested. The per-path  state is now a retrans_count, to help accurate reporting of the number of lost packets regardless of frame re-aggregation. It is also accounted for in the original lossy path, so that a middle observer has a consistent, local view.

ferrieux:
Well, transport_param_test.c is hairy: since the test vector format is undocumented and the semantics of the tests is unclear, I would appreciate a hand in doing that just once ;)

ferrieux:
Please review.

 Note that my changes to the test system are brittle since I did the minimal changes to hammer it into work with the new version PCQ1 (spin+QR). It would probably be better to keep testing PCQ0 (spin+VEC) and probably also the other variants (spin alone ;no spin at all).

 Thanks also for giving your view on whether the current semantics of the R bit properly addresses your remarks about frame retransmission and multipath.

huitema:
OK, let's check this in as is. I think we did learn quite a few things doing this exercise:

1) There really should be an union type so each spin bit variant can define its own set of state data.
2) Need to add a "spinbit_init" function in the API, otherwise developers have to go inside the code.
3) Need do document and simplify the transport parameter test.

Thanks for doing this work! I will try to address the issues in new PR.

ferrieux:
Great, thanks !

PR #361, opened 2018-10-14
Separate structures for each spin variant

huitema:
Also, remove assumptions about "vec" in header parsing, and update the SQR counting of losses so the main code does not have to take a dependency on SQR.


huitema:
The monotonic counter is so that the main code does not have to write into the spin-specific data part. Plus, I also need to expose the losses to the congestion algorithm, so that does it.

ferrieux:
Ah OK, if there's a second need then it's not cheating :)
Forget about the other nits. Please merge.

PR #362, opened 2018-10-14
Use macros to document versions in TP test

huitema:
This should simplify the experience when updating the list of supported versions.

### DIFF
--- a/picoquictest/transport_param_test.c
+++ b/picoquictest/transport_param_test.c
@@ -27,10 +27,38 @@
 #include "../picoquic/tls_api.h"
 #include "picoquictest_internal.h"
 
-/* Start with a series of test vectors to test that 
- * encoding and decoding are OK. 
- * Then, add fuzz testing.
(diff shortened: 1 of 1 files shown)


ferrieux:
At the beginning, transport_param_test failed without explanation (-1), and under gdb I saw it was a length comparison, followed by a memcmp, on  binary vectors. Then later, after fixing some of the vectors (but not all), transport_param_log_test came into view, and it was more readable. So if transport_param_log_test could capture all possible divergences, it would be self-sufficient.

huitema:
But the transport param test is for the code in transport.c, and the log test for the code in logger.c. One deals with binary, the other with text. I suppose that we could use better messages -- I tend to run these tests under debugger, so I just set a break point and look at the diffs. Better error messages might help understanding what is happening. For example, I could compare the content of the macros to the current version of the code, and give a warning if they do not match. I will open an issue for that.

Issue #354, opened 2018-10-12
migration during handshake


kenmcmil:
commit: be5a9b1

If migration of the client occurs during the handshake, strange server behavior follows. That is, it seems that the handshake completes and the server continues to accept stream frames, but it stops sending any ack packets. I'm not sure what the correct server behavior would be in this odd case, but I imagine it should either close the connection or continue to function normally. 

The server log output follows. The two received packets marked with *** were sent from a different ip address from the previous client packets. The client is sending random bytes on stream 4, which is why the funny characters appear in the log.

```
Starting PicoQUIC server on port 4443, server name = ::, just_once = 0, hrr= 0
b: Receiving packet type: 2 (initial), S0, Version ff00000e,
b:     <000000000000000b>, <000000000000000d>, Seq: 0, pl: 1171
b:     Decrypted 1171 bytes
b:     Crypto HS frame, offset 0, length 398: 0100018a0303915a...
b:     padding, 769 bytes

b: 0.003504 : Connection established, state = 11, from length: 16
b: Client address: 127.0.0.1, port 31763
b: Received SNI: localhost
b: ALPN not received.
b: Did not receive transport parameter TLS extension.
b: Sending packet type: 2 (initial), S0, Version ff00000e,
b:     <000000000000000d>, <dd3c698f32501834>, Seq: 0, pl: 132
b:     Prepared 132 bytes
b:     ACK (nb=0), 0
b:     Crypto HS frame, offset 0, length 123: 0200007703030d9e...

b: Sending packet type: 4 (handshake), S0, Version ff00000e,
b:     <000000000000000d>, <dd3c698f32501834>, Seq: 0, pl: 1031
b:     Prepared 1031 bytes
b:     Crypto HS frame, offset 0, length 1027: 0800000600040000...

b: Connection state = 6
b: Sending packet type: 4 (handshake), S0, Version ff00000e,
b:     <000000000000000d>, <dd3c698f32501834>, Seq: 1, pl: 132
b:     Prepared 132 bytes
b:     Crypto HS frame, offset 1027, length 127: b836306dd3150c89...

b: Sending packet type: 6 (1rtt protected phi0), S0,
b:     <000000000000000d>, Seq: 0 (0)
b:     Prepared 357 bytes
b:     Crypto HS frame, offset 0, length 137: 04000085000186a0...
b:     NEW CONNECTION ID[1]: 0xd2613bd458f92c7f, fd0b82a45fc2e51f5d94137f006da14f
[...]

PR #366, opened 2018-10-19
proposed fix for issue #354

kenmcmil:
Here is a proposed fix for issue 354. It prevents the path challenge timer for a path from being reset when a new address is seen, in the case when a path challenge was already scheduled. This prevents the indefinite postponement of the path challenge in case the client address changes frequently. This in turn prevents a possible man-on-the-side attack in which frequent replay packets  could indefinitely postpone the path challenge and thus block the connection.


huitema:
Looks like a simple change. I wish we had a test case to verify it, but I cannot really work on that while traveling.

Issue #367, opened 2018-10-21
Need to add test of VNEGO + 0-RTT


huitema:
when a picoquic client tries 0-rtt and receives a vneg response, it's next Initial isn't padded to 1200 bytes

huitema:
Patched with PR #368 but not closing this bug yet. Need to:

    Add a test of vnego with 0-RTT
    Repeat 0-RTT packet if doing VNEGO

huitema:
Done in Pr #513.

PR #368, opened 2018-10-21
Force padding of CI after negotiation

huitema:


huitema:
This is a fix for issue #367. The code expects to coalesce CI and 0-RTT when 0-RTT is available, and to pad the 0-RTT packet so it fills the required packet length. But the padding only happens for the 1st 0-RTT packet in a session,and in case of version negotiation the 1st packet was sent prior to repeating the CI. The best solution would be to also repeat the 0-RTT packet if doing V-NEGO, but for now we will simply pad the CI sent after 0-RTT.

Issue #372, opened 2018-10-24
Server does not check packet number when detecting migration


kenmcmil:
The server seems to ignore the the packet number of a packet arriving from a new address when detecting migration. 

The spec says the follow in section 6.11.3:

   An endpoint only changes the address that it sends packets to in
   response to the highest-numbered non-probing packet. This ensures
   that an endpoint does not send packets to an old peer address in the
   case that it receives reordered packets.

The following is a log of an execution that violates this condition. The received packets marked *** with sequence number 6,4,1 arrive from a new address. These packet numbers are not the largest seen so far (the highest is 0xa)  which means the server should not start sending to the new address. However, the server sends a path challenge to the new address. 

This violation seems to have fairly low severity. The worst case seems to be that some packets will be lost in a NAT rebinding scenario.
 
```
Starting PicoQUIC server on port 4443, server name = ::, just_once = 0, hrr= 0
1: Receiving packet type: 2 (initial), S0, Version ff00000f,
1:     <0000000000000001>, <000000000000000d>, Seq: 0, pl: 1170
1:     Decrypted 1170 bytes
1:     Crypto HS frame, offset 0, length 414: 0100019a03033ee8...
1:     padding, 752 bytes

1: 0.002029 : Connection established, state = 11, from length: 16
1: Client address: 127.0.0.1, port 31507
1: Received SNI: localhost
1: ALPN not received.
1: Received transport parameter TLS extension (44 bytes):
1: Proposed version: ff00000f
1:     Extension list (38 bytes):
1:         Extension type: 0, length 4 (0x0000 / 0x0004), 00002000
1:         Extension type: 1, length 4 (0x0001 / 0x0004), 00004000
1:         Extension type: 3, length 2 (0x0003 / 0x0002), 003c
1:         Extension type: 10, length 4 (0x000a / 0x0004), 00002000
1:         Extension type: 11, length 4 (0x000b / 0x0004), 00002000
1: Sending packet type: 2 (initial), S0, Version ff00000f,
1:     <000000000000000d>, <074ff267c23d161e>, Seq: 0, pl: 132
1:     Prepared 132 bytes
1:     ACK (nb=0), 0
1:     Crypto HS frame, offset 0, length 123: 02000077030350a8...

1: Sending packet type: 4 (handshake), S0, Version ff00000f,
1:     <000000000000000d>, <074ff267c23d161e>, Seq: 0, pl: 1031
1:     Prepared 1031 bytes
[...]

PR #373, opened 2018-10-27
Check PN before updating addresses

huitema:
This should address issue #372


kenmcmil:
Yes, I have checked this and it does now pass my test.

PR #395, opened 2018-11-12
Fix prepare stream frame

bkchr:
Hey,
I encountered some bugs, while porting your latest changes to rust.

The first one was that a packet was padded and the stream was sending data with a length of `0`. It happened that the padding data was interpreted as stream data, because no length was set!

Another bug I discovered while doing that was the `consumed` argument for `picoquic_prepare_stream_frame`. The argument was `4`, because some function before set it and so, the data was still send. However, `consumed` should be set by each call of `prepare_stream_frame`.

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -1074,8 +1074,8 @@ int picoquic_prepare_stream_frame(picoquic_cnx_t* cnx, picoquic_stream_head* str
                 if (stream->stream_id != 0) {
                     cnx->data_sent += length;
                 }
-                *consumed = byte_index;
             }
+            *consumed = byte_index;
 
             if (ret == 0 && STREAM_FIN_NOTIFIED(stream) && stream->send_queue == 0) {
                 /* Set the fin bit */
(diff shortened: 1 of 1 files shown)


picoquic/frames.c
```diff
@@ -1017,46 +1017,44 @@ int picoquic_prepare_stream_frame(picoquic_cnx_t* cnx, picoquic_stream_head* str
...
-                if (stream->stream_id != 0) {
-                    if (length > (cnx->maxdata_remote - cnx->data_sent)) {
-                        length = (size_t)(cnx->maxdata_remote - cnx->data_sent);
-                    }
+            /* Abide by flow control restrictions, stream 0 is exempt */
+            if (stream->stream_id != 0) {
+                if (length > (cnx->maxdata_remote - cnx->data_sent)) {
+                    length = (size_t)(cnx->maxdata_remote - cnx->data_sent);
                 }
+            }
```
huitema: Should get rid of the stream 0 test, this is a leftover of the pre-draft-13 definition of stream 0

bkchr: Done

bkchr:
@huitema did you see that I pushed another commit?

PR #398, opened 2018-11-21
Remove support for draft-13 and draft-14

huitema:


huitema:
Added the h3zero.[ch] files because I somehow got a dependency of the windows build on that work in progress. They are not actually required yet, will be fully developed in the h3zero PR which is work in progress. They are also not yet needed in the Unix builds.

huitema:
Restarted the appveyor build after updating Picotls -- latest Picotls Pr fixes Windows build issues.

PR #401, opened 2018-11-22
Fix the windows build after picotls changes

huitema:


huitema:
Need much more cleaning than than, due to my mistakes.

PR #402, opened 2018-11-22
Build fixes and the draft-14 removal

huitema:
Fixes the github fumbling that caused some fixes to disappear.
The fix to client hello callback is restored.
Get the latest version of picotls, after updating of windows builds.
Get the removal of the old draft-14 compatibility.


huitema:
Finally got some sanity. This fixes issue #398, and also makes sure that the Windows test builds do work.

Issue #387, opened 2018-11-04
Add server false start state


huitema:
The current state machine sees the server moving to a "server ready" state just after receiving the 1-RTT write keys. But the 1-RTT receive key is not available yet, and will only be when the client "finished" message is received in TLS. Before that message is received, TLS does not know whether there is a man-in-the-middle on the link, so it really is in a "false start" state. Applications should only send data that they are comfortable to disclose to a man in the middle.

There is actually another transition, happening when the first handshake message from the client is received. This message proves that all initial packets have been received, and do not need to be retransmitted any more. In most cases, the first handshake message from the client will be the TLS finished message, but not always.

huitema:
Done in PR #404

Issue #384, opened 2018-11-03
lnitial packets should not be processed after handshake


huitema:
Once the handshake is complete, there is no point in accepting data from Initial packets, except maybe to process acknowledgements. There is in fact a potential attack in which an attacker injects a spoofed initial packet with a bogus frame, causing the logic to fail the connection. This breaks the expectation that connections are only vulnerable during the initial handshake.

huitema:
Waiting until after the next QUIC meeting in Tokyo to implement this code change. There are two sides of the arguments: 

1) Implicit ACKs prevent attacks after a full handshake, and that's easy to implement.

2) But we discussed implicit ACKs before, are worried about side effects, and it is only a partial fix in any case, since the connection is still exposed to hacks during the handshake itself.

Depending on the discussion, the outcome maybe either to do implicit ACKs as planned, or to merely arm a timer and stop accepting initial or handshake packets 3*RTO after the handshake complete.

huitema:
Solved in issue #500

PR #404, opened 2018-11-23
Server false start and implicit ack

huitema:
Implicit acknowledgement of handshake packets when the handshake is verified complete. On server, this means obtaining the decryption key for the client data, which only happens when the client "finished" message is received and decoded. On client, this means receiving an acknowledgement for 1RTT data sent by the client, which is proof that the server could receive packets.

This is preliminary work for draft-17, which will mandate implicit ack. This is not yet a complete fix for issue #384, which will only be solved with draft 17. It is however a fix for issue #387.

The PR introduces two new connection states:

1) server_false_start: when the server can send 1RTT data, but cannot yet receive it.
2) client_ready_start: when the client can send and receive 1RTT data, but has not yet received an ACK of its own 1RTT data.

The PR also merges the "client_ready" and "server_ready" states in a single "ready" state.


picoquic/picoquic_internal.h
```diff
@@ -614,6 +614,7 @@ typedef struct st_picoquic_cnx_t {
     unsigned int client_mode : 1; /* Is this connection the client side? */
     unsigned int key_phase_enc : 1; /* Key phase used in outgoing packets */
     unsigned int key_phase_dec : 1; /* Key phase expected in incoming packets */
+    unsigned int data_acknowledged : 1; /* 1RTT data acknowledged by peer */
```
bkchr: Maybe should be renamed to `data_1rtt_acknowledged`.

picoquictest/tls_api_test.c
```diff
@@ -1338,17 +1341,17 @@ int tls_api_one_scenario_test(test_api_stream_desc_t* scenario,
 
 int tls_api_oneway_stream_test()
 {
-    return tls_api_one_scenario_test(test_scenario_oneway, sizeof(test_scenario_oneway), 0, 0, 0, 0, 70000, NULL);
+    return tls_api_one_scenario_test(test_scenario_oneway, sizeof(test_scenario_oneway), 0, 0, 0, 0, 100000, NULL);
```
bkchr: Why did you increase the max timeout?

huitema: There is a regression in the duration of tests. I believe it is because PMTU discovery starts later in the new state machine. I am going to try fix that.

huitema: Yes, that was the reason. Fixed now. Thanks for prodding me!

huitema: Fixed.

picoquic/picoquic_internal.h
huitema: OK. Renamed to "one_rtt_data_acknowledged".

Issue #214, opened 2018-07-05
Cleaning up stream flags -- state reduction


igorlord:
I am confused about the amount of transport state flags for streams.

It seems that *_signalled flags can all be removed, since there is already a "received" flags.  The code logic should always be:
```
if ((flags & foo_received) == 0) {
    flags |= foo_received;
    ....
    call callback();
}
```
So if the received flag is set, the callback has been called.  No need to ponder the meaning of cases where one flag is set but the other is not.

I am also a bit confused about the meaning of "fin_noftified".  Is it "application closed the sending part of stream"?  If so, should it be "fin_requested" for consistency with other flags?

huitema:
Fixed with PR #410

PR #410, opened 2018-11-27
Streamline stream state flags

huitema:
Addresses issue #214, clean up state flag.

Replaced the use of a bit mask encoding state flags by a set of boolean variables, defined as length 1 integers.

@igorlord I kept your macros in place, but they are probably less useful now.

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -261,7 +261,8 @@ picoquic_stream_head* picoquic_find_or_create_stream(picoquic_cnx_t* cnx, uint64
 
         } else if (!IS_BIDIR_STREAM_ID(stream_id)) {
             /* Mark the stream as already finished in our direction */
-            stream->stream_flags |= picoquic_stream_flag_fin_notified | picoquic_stream_flag_fin_sent;
+            stream->fin_requested = 1;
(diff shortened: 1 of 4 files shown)


picoquic/frames.c
```diff
@@ -1113,7 +1116,7 @@ int picoquic_prepare_crypto_hs_frame(picoquic_cnx_t* cnx, int epoch,
     int ret = 0;
     picoquic_stream_head* stream = &cnx->tls_stream[epoch];
 
-    if ((stream->send_queue == NULL || stream->send_queue->length <= stream->send_queue->offset) && ((stream->stream_flags & picoquic_stream_flag_fin_notified) == 0 || (stream->stream_flags & picoquic_stream_flag_fin_sent) != 0)) {
+    if (stream->send_queue == NULL || stream->send_queue->length <= stream->send_queue->offset) {
```
igorlord: Just want to double-check that the omission of `&& (!stream->fin_requested || stream->fin_sent)` was deliberate.

huitema: Yes it is. You cannot set the FIN bit on a crypto stream.

Issue #285, opened 2018-08-17
Padding should be driven by policy


huitema:
As part of the draft-14 support work, we added systematic padding of short messages to 38 bytes, so they can all elicit a stateless reset if the peer is gone. This is however partly a policy decision, and such decisions should not be hard-coded. We need some policy control there, and probably several options:

* No padding, for minimal overhead;
* Pad to 38, to elicit stateless reset; 
* Pad for privacy, probably using some logarithmic scale to trade off privacy and overhead.

huitema:
Done in PR #414 .

PR #414, opened 2018-12-02
Padding policy

huitema:
Implement support for basic padding policies, and corresponding test.


huitema:
This resolves issue #285. @bkchr , you may want to take a look at the "padding policy" API, they could be useful in Rust too.

bkchr:
Okay, I will take a look.

Issue #415, opened 2018-12-02
Implement pacing, again


huitema:
The support for pacing seems to have disappeared with the rework of the waiting loop. Bring it back!

huitema:
The spec says:

_This document does not specify a pacer, but it is RECOMMENDED that a sender pace sending of all in-flight packets based on input from the congestion controller. For example, a pacer might distribute the congestion window over the SRTT when used with a window-based controller, and a pacer might use the rate estimate of a rate-based controller.

An implementation should take care to architect its congestion controller to work well with a pacer. For instance, a pacer might wrap the congestion controller and control the availability of the congestion window, or a pacer might pace out packets handed to it by the congestion controller. Timely delivery of ACK frames is important for efficient loss recovery. Packets containing only ACK frames should therefore not be paced, to avoid delaying their delivery to the peer.

As an example of a well-known and publicly available implementation of a flow pacer, implementers are referred to the Fair Queue packet scheduler (fq qdisc) in Linux (3.11 onwards)._

huitema:
Basic pacing spec: just implement a leaky bucket.

huitema:
Done in PR #417

Issue #391, opened 2018-11-06
RTO notification for congestion control


huitema:
Only do that when the loss is verified, i.e. when the packet is confirmend lost, to give a chance for spurious RTO detection.

huitema:
The proposed solution does not work very well. It is simpler to just roll back the spurious congestion notification, at least for now.

huitema:
Mitigated in PR #417

PR #417, opened 2018-12-03
Pacing tlp congestion

huitema:
Implement pacing, using a simple leaky bucket algorithm. Pacing does improve performance a lot. @bkchr: you should look at the reduction in execution time of several long tests.

Tried to implement the idea of only notifying the congestion algorithm when a loss is confirmed, but this caused a bunch of regressions. Instead, implemented the simpler solution of rolling back the congestion window decrease if a repetition is marked as spurious.

Tried to implement Tail Loss Probe, but this requires more work. The naive spec just causes spurious retransmission and performance regression.


huitema:
This addresses issues #391 and #415.

Issue #310, opened 2018-09-17
CMake enhancement


promgamer:
We are trying to use PICOQUIC as a 3rdparty dependency on a CMake project. We found some problems that we solve by applying patches after checkout.

- Relative paths (with "../../../").
- Include picotls easily
- Duplicated declarations on CMakeLists.txt

Will you accept a CMAKE enhancement that solves some of those problems?

bkchr:
Just make the pull request and than let's see :)

huitema:
@promgamer as Bastian said, yes, we are absolutely accepting patches.

huitema:
PR #420 removes the usage of relative paths (../.., etc.) in header declarations and also in test file configuration. Tests and demo programs can be run from any location, provided that the user specifies the path to the Picoquic solution in a command line argument, e.g. "picoquic_ct -S ~/picoquic". @promgamer, does that resolve your "../../.." comment?

huitema:
There are some remaining direct references in `CMakeList.txt`, such as:
```
FIND_LIBRARY(PTLS_CORE picotls-core PATH ../picotls)
MESSAGE(STATUS "Found picotls-core at : ${PTLS_CORE} " )

FIND_LIBRARY(PTLS_MINICRYPTO picotls-minicrypto PATH ../picotls)
MESSAGE(STATUS "Found picotls-crypto at : ${PTLS_MINICRYPTO} " )

FIND_LIBRARY(PTLS_OPENSSL picotls-openssl PATH ../picotls)
MESSAGE(STATUS "Found picotls-openssl at : ${PTLS_OPENSSL} " )
```
Any suggestion on how to replace them?

promgamer:
Well, I see at least 3 options here:

1. Use `find_library` on picoquic and provide a `FindPicoTLS.cmake` with picoTLS.
2. Include picoTLS as a git submodule of picoquic
3. Use Cmake variables like `-DPICOTLS_DIR` that we can define to point it to picoTLS

1 and 3 "decouple" more picoquic and picotls by being more flexible with what you can reference (either via CMake's find package or via CMake variables. But it may be harder to execute and maintain.
2 seems easier execute and maintain, but it couples picoquic's repository to picotls repository (which doesn't seem a problem to me, to be honest :smile: )

huitema:
[...]

PR #420, opened 2018-12-08
Remove usage of relative paths (../xxx.h) in headers and test configuration

huitema:
Addressing some of the comments in issue #310, use of relative paths in include files.


huitema:
Also added a "-S solution_dir" parameter to the demo program and the test program, so they can be run from other locations than directly under the solution.

Issue #421, opened 2018-12-09
Crash on migration attempt


huitema:
On 12/9/2018 8:50 AM, איתמר אורדני (Itamar Ordani) wrote:

_We have been experimenting with the migration exercise in your QUIC implementation, specifically the migration to a new IP exercise, and we have been getting a segmentation fault once the migration starts.
We are working on Ubuntu and have followed the setup guide in your github page.
The server command we used:
picoquicdemo -c certs/cert1.pem -k certs/key1.pem
(we used our own certificate in case it matters)
The command we used to run the client is:
picoquicdemo -f 3 127.0.0.1_

Attached logs show that the client crashes after starting a migration:
```
Connection established. Version = 50435131, I-CID: 216df16b8f10f6b2
Opening stream 0 to GET /index.html
Switch to new port, sending probe.
216df16b8f10f6b2: Sending packet type: 6 (1rtt protected), S0,
216df16b8f10f6b2:     <f27b15dd8dedd683>, Seq: 1 (1), Phi: 0,
216df16b8f10f6b2:     Prepared 1203 bytes
216df16b8f10f6b2:     NEW CONNECTION ID[8]: 0xd8db4acb6d88a088, 0dd3fa3e7f512912b125eeb1ece82af1
216df16b8f10f6b2:     path_challenge: fcb0f680b8872bce
216df16b8f10f6b2:     padding, 1167 bytes


Program received signal SIGSEGV, Segmentation fault.
picoquic_log_packet_address (F=0x7ffff79f7400 <_IO_2_1_stdout_>, 
    log_cnxid64=<optimized out>, cnx=cnx@entry=0x66a450, 
    addr_peer=addr_peer@entry=0x0, receiving=receiving@entry=0, 
    length=length@entry=1232, current_time=current_time@entry=1544371070220630)
    at /home/itamar/picoquicclean/picoquic/picoquic/logger.c:117
117	    if (addr_peer->sa_family == AF_INET) {
```

PR #422, opened 2018-12-09
crashes and hot loops on migration

huitema:
This PR addresses the issue #421, as well as a couple of other related bugs found when debugging the issue.


huitema:
As a matter of expediency, one of the fixes was to disable congestion control feedback on non-default paths. This is tracked in issue #421.

PR #439, opened 2018-12-17
Fix New Reno bug

huitema:
The congestion avoidance in the new reno implementation was way too conservative, one byte increase per RTT instead of 1 packet!


huitema:
@bkchr this fixes a long standing issue, and explains why the performance were so sensitive to errors on the link. The "test with error" now executes in 2.1 sec instead of more than 4 sec, and the "congestion test" executes in less than 1 sec instead of 1.5 sec.

bkchr:
Nice, maybe this resolves some long standing connections problems I have in my application :) I will try your fix. Ty.

Issue #446, opened 2018-12-21
Path challenge sent to wrong path after migration


kenmcmil:
The following log shows a case where a migration occurs from port 5047 to port 5048. However, instead of sending path challenges to both ports, the server sends two path challenges to port 5047 and continues to try to validate this path without ever validating the path to the new port, eventually timing out and closing the connection.

Also, notice that the server does not seem to recognize the PATH_RESPONSE to the first of the two challenges. It seems to indicate that the second PATH_CHALLENGE overwrote the first, which was possibly intended for the new path.  

```
Starting PicoQUIC server on port 4443, server name = ::, just_once = 0, hrr= 0
000000000000003d: Receiving 1232 bytes from 127.0.0.1:5047 at T=0.000000 (57d7e36c87452)
000000000000003d: Receiving packet type: 2 (initial), S0, Version ff000010,
000000000000003d:     <000000000000003d>, <000000000000003c>, Seq: 0, pl: 1207
000000000000003d:     Token length: 0
000000000000003d:     Decrypted 1187 bytes
000000000000003d:     Crypto HS frame, offset 0, length 255: 010000fb0303284c...
000000000000003d:     padding, 928 bytes

3d: 0.002320 : Connection established, state = 11, from length: 16
3d: Client address: 127.0.0.1, port 46867
000000000000003d: SNI not received.
000000000000003d: Received ALPN: hq-16
000000000000003d: Received transport parameter TLS extension (44 bytes):
3d: Proposed version: ff000010
3d:     Extension list (38 bytes):
3d:         Extension type: 0, length 4 (0x0000 / 0x0004), 00002000
3d:         Extension type: 1, length 4 (0x0001 / 0x0004), 00004000
3d:         Extension type: 3, length 2 (0x0003 / 0x0002), 003c
3d:         Extension type: 10, length 4 (0x000a / 0x0004), 00002000
3d:         Extension type: 11, length 4 (0x000b / 0x0004), 00002000
000000000000003d: Sending packet type: 2 (initial), S0, Version ff000010,
000000000000003d:     <000000000000003c>, <e3e4b9d1983d775e>, Seq: 0, pl: 184
000000000000003d:     Token length: 0
000000000000003d:     Prepared 164 bytes
000000000000003d:     ACK (nb=0), 0
000000000000003d:     Crypto HS frame, offset 0, length 155: 0200009703039c0b...

000000000000003d: Sending packet type: 4 (handshake), S0, Version ff000010,
[...]

PR #449, opened 2018-12-23
Check addresses of probes and responses

huitema:
This PR addresses issue #446 and implements the address checks required in draft-17. A challenge response is only accepted if it comes from exactly the same addresses to which the challenge was sent. Responses to NAT rebinding are similarly processed. The NAT rebinding stress test succeeds even in presence of a concerted attack rewriting the addresses of 1/2 the packets.


huitema:
@kenmcmil you may want to take a look at the rebinding stress test. I believe that it simulates the address rewriting test that you were doing.

PR #452, opened 2018-12-27
.travis(.yml): Update (matrix) travis build

alagoutte:
split clang and clang analyzer (scan-build) build
use matrix (with name !) to different build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,5 @@
 dist: xenial
 language: cpp
-os:
-  - linux
-compiler:
-  - clang
-  - gcc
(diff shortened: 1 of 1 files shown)


huitema:
The pull request itself is fine, but I have a hard time understanding the root cause of the clang analyzer warnings. Is there a way to ask the analyzer to provide more detailed feedback, such as maybe the code path that may code the leak? (The Windows analyzer does that.)

alagoutte:
Hi @huitema 

Yes it is possible, scan-build generate html with code path
i can send to you if you want !

PR #458, opened 2018-12-28
Travis: add cppcheck

alagoutte:
add new job cppcheck (1.84)

found actually 2 "issue"
[picoquic/picosocks.c:541]: (error) failed to expand 'DBG_PRINTF', it is invalid to use a preprocessor directive as macro parameter
Look a issue with copy/paste (from vim ?) on commit 34243d67
[picoquic/picoquic/sender.c:252*]: (error) Address of local auto-variable assigned to a function parameter.
get 5 time this warning


huitema:
Thanks. The check is really useful.

PR #461, opened 2018-12-28
Cpp check fix

huitema:


huitema:
This needs review. The purpose of the PR is to fix the warnings from `cppcheck`. One of these warnings was about "Address of local auto-variable assigned to a function parameter", in code lines like:
```
*addr_to = (struct sockaddr *)&probe->peer_addr;
```
GCC Checks complain because "probe" is a stack variable. It is actually set to the address of an actual probe message queued for the connection, but somehow GCC does not believe that.

Thinking about it, the passing of IP address references in the current version is actually somewhat dangerous. It does save a copy, but it also passes pointers around. The addresses behind these pointers may change over time with migration, which can complicate behavior. It could also lead to problems in multi-threaded environments.

The proposed PR replaces the passing of IP addresses pointers by the copying of IP addresses values. I don't think that it degrades performance, because the existing code would often just copy again tehse addresses to their intended location. But it certainly changes the API, so we need a review.

Issue #463, opened 2018-12-28
Call `picoquic_prepare_packet` in a loop


bkchr:
I think we need to update the demo [here](https://github.com/private-octopus/picoquic/blob/f9798134f3e98ee4ad39d120e9f451b7edd9856b/picoquicfirst/picoquicdemo.c#L532).
We should call `picoquic_prepare_packet` in a loop until all packets for a connection are send. I think that this is one of the main issues, I have seen for connections with multiple streams. The main stream was generating a lot of data and the other streams did not get any real change to send something.
I built the rust bindings, based on your demo. So, I imported this problem upstream^^

bkchr:
Hmm that still did not really solved my problem. I need to take another look at it.
If I have multiple streams and do not add any data between the `prepare_packet` calls, all the data of all streams should be send?

bkchr:
Is the stress test using multiple streams?
I added another stream to my application today and have the feeling that it made everything more fragile.

Multiple concurrent steams that are sending data, should be doable with quic?

huitema:
I have been using multiple streams in the demo server for some time, and there was no obvious problem.  The stress test defaults to 4 streams max per connection; that number is really easy to change. However, the streams are drained in order, and the test servers just post a variable size message on each stream. I don't think we have a test in which the streams are just continuing, and that would certainly be a good addition to the test cases -- or to the stress behavior.

huitema:
The loop is written with a "receive first" logic: process all the packets that are already received through the sockets, or none if there is nothing ready; then prepare packets for the next connection in order. There is a next wake time for each connection, based on data queued, flow control, etc. The connections are then polled in order of the wake time. There could obviously be a bug, so that a given connection would always ask to be awaken "right now".

bkchr:
If I poll the connections based on their wake time, what happens if a connection has more data than what fits into one package?
As I understand it, the connection is scheduled to be woken up at "current_time", so that theoretically I will be requested to call picoquic immediately.
Now, when I restart the loop, I will begin with polling the streams again to see if they want to send something. The streams can send data from multiple threads. So, if stream 1 wants to send data  again, the other steams will not get any chance to send their data(if stream 1 is filling up one package again). Because of that, I thought that calling `prepare_packet` again, until no package is returned could enable the other streams to send their data.

huitema:
[...]

Issue #447, opened 2018-12-21
Data corruption, possible vulnerability


kenmcmil:
Possible vulnerability due to corruption of data in a retransmission
by server. The log is an excerpt of a longer log with many
connections. The relevant events are:

1) T=1.137155 Client sends HTTP request for /index.html on stream 24, seq=1

2) T=2.456074 Server sends 503 bytes on stream 24, offset 0, seq=13

3) T=3.758303 Server sends 130 bytes on stream 24, offset 503, seq=15.
The first byte of data at offets 503 is 0x4e.

4) T=4.252034 Server resends 513 bytes on stream 24, offset 0, seq=16.
Byte 503 is 0x1a which conflicts with the value sent in step 2.

The original text sent in steps 1 and 2 is:
```
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<HTML>
<HEAD>
<TITLE>PicoQuic HTTP 0.9 service</TITLE>
</HEAD><BODY>
<h1>Simple HTTP 0.9 Responder</h1>
<p>GET /, and GET index.html returns this text</p>
<p>Get /doc-NNNNN.html returns html document of length NNNNN bytes(decimal)</p>
<p>Get /doc-NNNNN also returns html document of length NNNNN bytes(decimal)</p>
<p>Get /doc-NNNNN.txt returns txt document of length NNNNN bytes(decimal)</p>
<p>Get /NNNNN returns html document of length NNNNN bytes(decimal)</p>
<p>Any other command will result in an error, and an empty response.</p>
<h1>Enjoy!</h1>
</BODY></HTML>
```

The corrupted text sent in step 3 is:

```
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<HTML>
<HEAD>
<TITLE>PicoQuic HTTP 0.9 service</TITLE>
</HEAD><BODY>
<h1>Simple HTTP 0.9 Responder</h1>
<p>GET /, and GET index.html returns this text</p>
<p>Get /doc-NNNNN.html returns html document of length NNNNN bytes(decimal)</p>
<p>Get /doc-NNNNN also returns html document of length NNNNN bytes(decimal)</p>
<p>Get /doc-NNNNN.txt returns txt document of length NNNNN bytes(decimal)</p>
<p>Get /NNNNN returns html document of length NNN^Z^X\200^Cl\357^@^D^@
```

It seems clear that a buffer has been overwritten. This could be a
[...]

PR #465, opened 2018-12-29
Ensure every sending stream is visited in turn

huitema:
This is trying to address issue #463. The first step is to make sure that each stream is visited in turn. But the solution guarantees that the time to find available data scales with the number of streams opened since the beginning of the connection, which will be a problem for many applications.


huitema:
Also adding a "ready to send" callback, and a test to verify that this works. The combination of rotation and callback should enable applications to work around flow control issues.

bkchr:
Hey, I look over your changes and would still prefer a second list for "active" streams. If you want, I could prepare a pull request on what I think.

And for the send call back, I would do this different as well. I would expose a function for the streams to inform that they want to send data. This just sets a variable to `true`. Then, if we prepare a packet, we would ask all the streams which want to send data to give us data. We would also give them already the maximum number of bytes we can accept in this turn. The callback could look like:
`unsigned int get_stream_data(max_bytes: unsigned int, buffer: *char)`
The callback gets the maximum number of bytes to write into `buffer` and `buffer` to write the result. The return value is the number of bytes written.

huitema:
Your idea of an active stream is a nice optimization. I like it. Instead of managing buffers of queued data, the transport would just ask the application to provide "up to N bytes of data". This could be done at the very last moment, inside `picoquic_prepare_stream_frame`. If the stream is marked as "active", the function would just activate the call back and get just enough bytes to fill the packet. That would save a lot of copying, buffering and memory allocation.

My concern is that many applications do not naturally follow a poll model. There is an important class of applications that use the query/response paradigm: client opens a stream and sends a query; server does some processing and some time after sends a response; then the stream is closed. If we only managed data using the callback model, these applications would have to keep the query or response in memory, mark the stream as active, and wait for a series of callbacks until the data is sent. My first idea is to accommodate that by making the "active" mark optional; if a stream is marked active, use the call back, if not, use the current data queue management.

I am less convinced by the idea of a separate list of connections that are actively sending data. We have to manage two events: data provided by the application, and "max stream data" authorized by the peer for that stream. We have also to be concerned with "stop sending" and "reset" events that can change the state of the stream, and take it out of the active list. Then, we also need to visit "receiving" streams, to check whether we need to update their "max stream data" limit. We do all that today by polling every stream in turn. It is simple and robust, but yes it may later become a bit slow if there are many streams.

If I did an optimization, my first priority would be to reduce the size of the current list so it only contains the streams that are "not finished yet". This would have benefits across the board, as the list of streams is used in many places. I opened a separate issue for that (#467).

Can we start simply with the better callback implementation for active streams?

bkchr:
Yeah, then let's start with the callback implementation :) 
We could use the `add_data_to_stream` function for applications that want to use a push model, these applications would not register a `poll_stream_data_callback`. We would provide this callbakc and the user could just call `add_data_to_stream` and we take care of rest. Aka, our `poll_stream_data_callback` is called :)

huitema:
Do you want to register one callback per stream, or just one callback per connection? With or without some kind of context argument?

I would think something like: int (callback_fn)(void * callback_ctx, uint64_t stream_id, uint32_t max_length, (uint8_t **) data_ptr, (uint32_t *) data_length)

huitema:
Sorry, did not fully read the message. You mention calling `add_data_to_stream` from within the call back, but I think it is even better to provide the bytes as function arguments. That way, the `prepare_stream_frame` function can just pull the bytes from the application context and put them in the frame.

There is a weird off-by-one issue in that design. The stream frames have two potential formats:

1) no length field, just data bytes to the end of the packet. N bytes available mean N bytes of data
2) variable length length field of LL bytes, followed by L bytes of data.

The length encoding is:

- 0 to 63 bytes: 1 byte
- 64 to 16K: 2 bytes

So assume a callback says "you can write 1000 bytes, how many can you give me?". If the answer is 1000 bytes, things are good. But if the answer is 999 bytes, the encoding will require two bytes of length plus 999 bytes, and that does not fit.

Or maybe it does. I just need to insert one padding byte in front of the stream frame, and then there are only 999 bytes available. Yes, that will work, with moderate ugliness. So let's go for that.

Still one question: register the function once per stream, or globally? I think once per stream is the most robust. Will cost two pointers for each stream, but we are not really optimizing memory usage that much.

huitema:
In fact, I may be overthinking this. It may be better to just reserve 2 bytes for encoding the length of the data, and then let the application write directly in the frame buffer. Here is the dilemma:

1) spend two bytes, but then the application can write the data directly in the frame buffer
2) save two bytes, but then the transport has to copy the bytes from application data to frame buffer

Any preference?

bkchr:
I think doing it as @igorlord has proposed it, would be the best way. We would not require to use 2 bytes every time and the user can safely write its data. Regarding the number of callbacks, I would go with the current style, one callback per connection.

huitema:
I like @igorlord idea. Basically, two call backs, one after the other.

1) Hello app, how many bytes (up to N) do you have ready for stream X? Return: L bytes, L <= N.
2) App, please write L bytes at this memory address. Return: OK or catastrophic fail.

I can implement that with the existing call back API and just two changes: providing an integer return code, which seems like a good idea anyhow, and adding 2 values to the callback type enum, which does not seem too onerous.

This ensures a single copy of data to the outgoing frame, without sacrificing 2 bytes in the process. As for retransmission, the code already "holds on to the cleartext packets until they are ACKed", and copies the frames from old to new packet. I have opened issue #473 to track the need for splitting retransmitted data frames when they don't fit in the new packet length.

igorlord:
@huitema, so you prefer two callbacks:

(library) packet composing thread:
-> (app) number_of_bytes_for_next_write_for_stream(stream, max_n)
-> (app) write_bytes_for_stream(stream, n, buffer)

I do not love this, because the time between the two callbacks is undefined.  The app may need to prepare some state to assess the number of bytes, and the state would be lost between the callbacks and would need to be re-established in write_bytes_for_stream().  It is also not clear whether one callback will follow after the other, or maybe the library will call number_of_bytes_for_next_write_for_stream() many times before it starts to call write_bytes_for_stream().  Can the app rely on write_bytes_for_stream() even being called?

I may prefer a simpler version (I tried to outline it before, but I was writing from a phone...)

(library) packet composing thread:
-> (app) write_bytes_for_stream(callback_ctx, stream_id, max_length, data_ctx)
      -> get_data_ptr(data_ctx, data_length): returns uint8_t *
      -> memcpy

So there is only one callback.  But that callback does not get a data ptr.  Instead, it gets an opaque data_ctx, which can be used inside write_bytes_for_stream callback to obtain the pointer to the buffer to write the data.  A "side effect" of obtaining the pointer to the buffer is the library writing out the appropriately-sized data length.

huitema:
@bkchr I see what you mean. Yes, from the application point of view that's better, but we have to design a way to do that pointer management without getting something hyper-complicated. 

It would also be very nice if the application could indicate that it has reached the end of the stream. I think that can be done in the same call. So that would be:

-> (app) write_bytes_for_stream(callback_ctx, stream_id, max_length, data_ctx)
-> get_data_ptr(data_ctx, data_length, fin_bit): returns uint8_t *
-> memcpy

The stack will have some housekeeping to do after that, but I think it can manage. I will update the PR with that design.

huitema:
Oops, answered @bkchr when the suggestion was from @igorlord. Sorry Igor. Tells me that I should not try working too much on New Years Eve!

Happy New Year!

huitema:
Could not quite keep my hands off the keyboard, but then the party only starts in two hours. The latest check-in implements the complete idea. It looks like a reasonably clean programming model.

picoquic/frames.c
```diff
@@ -904,32 +914,102 @@ uint8_t* picoquic_decode_stream_frame(picoquic_cnx_t* cnx, uint8_t* bytes, const
...
+    if (length <= data_ctx->space) {
+        data_ctx->length = length;
+
+        if (is_fin) {
+            data_ctx->is_fin = 1;
+            data_ctx->bytes[0] |= 1;
+        }
+
+        if (length < data_ctx->space) {
+            if (length == data_ctx->space - 1) {
```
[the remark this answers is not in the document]
huitema: There are N bytes available, the application wants to write N-1 bytes. We can encode N bytes because then we don't need a length field, just a flag in the first byte. But if we had to encode "length=N-1", that would typically require 2 bytes, for a total of (2 + N-1)=N+1 bytes, larger than the packet size. We also don't want to avoid the length field, because the encoding would be shorter than the packet size, and other parts of the code might add a byte after that, e.g. padding,  which the receiver would mistake as data because of the "implicit length" encoding. So we work against that issue by inserting a single padding byte in front of the text header.

huitema:
Thanks for the review.

I am not merging this immediately. I did more testing, and found that we need an additional argument to
"picoquic_provide_stream_data_buffer", to state whether the stream "is_still_active". Without that, the code can be driven in a loop of active polling.

When doing the additional tests, I also reproduced the data corruption issue noted by @kenmcmil in bug #447. It is caused by using implicit length encoding when the data does not in fact extend to the end of the packet. Working to fix that before checking in the PR.

picoquic/picoquic.h
```diff
@@ -449,13 +450,47 @@ int picoquic_prepare_packet(picoquic_cnx_t* cnx,
...
+ * pointer, with the number of bytes that it wants to write, with an indication
+ * of whether or not the fin of the stream was reached, and also an indication
+ * of whether or not the stream is still active. The function
+ * returns the pointer to a memory address where to write the byte -- or
+ * a NULL pointer in case of error. The application then copies the specified 
+ * number of bytes at the provided address, and provide a return code 0 from
+ * the callback in case of success, or non zero in case of error.
+ */
+
+uint8_t* picoquic_provide_stream_data_buffer(void* context, size_t nb_bytes, int is_fin, int is_still_active);
```
igorlord: Since there are only three possible values for (is_fin, is_still_active) tuple, I can see a case for an enum instead of two booleans.  But this is only a minor nit.

picoquic/picoquic.h
```diff
@@ -449,13 +450,47 @@ int picoquic_prepare_packet(picoquic_cnx_t* cnx,
     uint64_t current_time, uint8_t* send_buffer, size_t send_buffer_max, size_t* send_length,
     struct sockaddr_storage * p_addr_to, int * to_len, struct sockaddr_storage * p_addr_from, int * from_len);
 
-/* send and receive data on streams */
+/* Mark stream as active, or not.
+ * If a stream is active, it will be polled for data when the transport
+ * is ready to send.
+ * Returns an error if data was previously queued on the stream using
+ * "picoquic_add_to_stream" and the data has not been sent yet.
```
igorlord: I do not love "Returns an error if data was previously queued on the stream using 'picoquic_add_to_stream' and the data has not been sent yet".

The interface seems to allow the application to sometimes use picoquic_add_to_stream and sometimes use the callback functionality.  However, the application cannot reliably use callbacks after using picoquic_add_to_stream, because it does not known when all data queued via picoquic_add_to_stream has been sent.

The cleaner semantics would allow picoquic_mark_active_stream any time.  When the transport sends all previously-buffered data, if the steam is marked active, the application callback will be called to get more data.  It is important that picoquic_add_to_stream cancels the "active" status (as it does now).

I'd propose this:

picoquic_mark_active_stream IS allowed while there is buffered data.  The library implementation is something like:

```
if stream is active:
    callback_max_bytes = max_bytes_for_stream - num_buffered_bytes
    if callback_max_bytes > 0:
          call_callback(..., callback_max_bytes, .... )
     else
           write_buffered_bytes_to_stream_frame(min(max_bytes_for_stream, num_buffered_bytes))
```
picoquic_provide_stream_data_buffer implementation:
```
    data_buf_ptr = write_stream_frame_header(num_buffered_bytes + nb_bytes)
    if buffered_bytes > 0
         copy_buffered_bytes(data_buf_ptr, num_buffered_bytes)
    return data_buf_ptr + num_buffered_bytes
```

A solid implementation will also make sure to disallow picoquic_add_to_stream while in the callback (via a mutex, for example -- deadlock if called from the callback thread itself; wait if multiple threads).

picoquicfirst/picoquicdemo.c
```diff
@@ -248,6 +248,11 @@ static void first_server_callback(picoquic_cnx_t* cnx,
     printf("Server CB, Stream: %" PRIu64 ", %" PRIst " bytes, fin=%d (%s)\n",
         stream_id, length, fin_or_event, picoquic_log_fin_or_event_name(fin_or_event));
 
+    if (fin_or_event == picoquic_callback_prepare_to_send) {
```
igorlord: Seems like a good candidate for a switch statement now. :)

picoquic/frames.c
```diff
@@ -904,32 +914,115 @@ uint8_t* picoquic_decode_stream_frame(picoquic_cnx_t* cnx, uint8_t* bytes, const
...
+                 * the max stream id limit */
+                 /* Check parity */
+                if (IS_CLIENT_STREAM_ID(stream->stream_id) == cnx->client_mode) {
+                    if (stream->stream_id <= cnx->max_stream_id_bidir_remote) {
+                        break;
+                    }
+                }
+                else {
+                    break;
+                }
```
igorlord: Are the above 8 lines equivalent to:
```
if (IS_CLIENT_STREAM_ID(stream->stream_id) != cnx->client_mode || stream->stream_id <= cnx->max_stream_id_bidir_remote) {
    break;
}
```

picoquic/picoquic.h
huitema: You are right, it is not hard to allow active marking at any time. And in fact it makes sense. For example, a web server might queue a header frame on the stream, and then rely on the callback to feed the frame's data.

The relation between application and transport is more "intimate" with an in-process transport like QUIC than with a kernel transport like TCP. If the application process crashes, the transport also crashes. The priority cannot be to try for full isolation -- we cannot. Instead, I would rather try for easy debugging: it should be easy for the app developer to understand why things are breaking, and to fix the application. The priority should thus be, return good error messages, plus have a reasonable documentation. But I guess that creating a deadlock would be a good way to provide feedback to the developer...

picoquic/frames.c
```diff
@@ -904,32 +914,115 @@ uint8_t* picoquic_decode_stream_frame(picoquic_cnx_t* cnx, uint8_t* bytes, const
...
+                 * We can encode N bytes because then we don't need a length field, just a flag in the 
+                 * first byte. But if we had to encode "length=N-1", that would typically require 2 
+                 * bytes, for a total of (2 + N-1)=N+1 bytes, larger than the packet size. We also
+                 * don't want to avoid the length field, because the encoding would be shorter than
+                 * the packet size, and other parts of the code might add a byte after that, e.g. padding,
+                 * which the receiver would mistake as data because of the "implicit length" encoding.
+                 * So we work against that issue by inserting a single padding byte in front of the 
+                 * text header.*/
+
+                for (int i = (int)data_ctx->byte_index - 1; i >= 0; i--) {
```
igorlord: memmove?

picoquic/frames.c
```diff
@@ -904,32 +914,115 @@ uint8_t* picoquic_decode_stream_frame(picoquic_cnx_t* cnx, uint8_t* bytes, const
...
+
+        if (is_fin) {
+            data_ctx->is_fin = 1;
+            data_ctx->bytes[0] |= 1;
+        }
+
+        data_ctx->is_still_active = is_still_active;
+
+        if (length < data_ctx->byte_space) {
+            if (length == data_ctx->byte_space - 1) {
```
igorlord: This special condition only makes sense when length cannot be encoded in a single byte.

picoquicfirst/picoquicdemo.c
huitema: Maybe, but I tried ans it is not obvious. There are several steps in the process, e.g. responses that do need to identify a stream, then responses that work without bothering about stream state, then responses that depend on the stream state. I would rather leave that alone until we actually rewrite the server process, maybe to use the callbacks. Also, we will need to revisit that if we start supporting multiple application protocols, e.g. Http 0.9, Http 1.0, and H3.

picoquic/frames.c
```diff
@@ -904,32 +914,115 @@ uint8_t* picoquic_decode_stream_frame(picoquic_cnx_t* cnx, uint8_t* bytes, const
...
+        /* Do only one pass if we found a stream */
+        if (stream != NULL) {
+            break;
+        }
+    }
+
+    return stream;
+}
+
+typedef struct st_picoquic_stream_data_buffer_argument_t {
```
igorlord: It would be a _really_ good idea to document the fields below.  Some of them have non-trivial semantics (I am looking at `space` vs `byte_space`).

picoquic/frames.c
huitema: Yes, but then we are optimizing an exception case. The single byte is for data shorter than 64 bytes, the average packet length is over 1KB, so we are really looking at 1/16th of the cases. I would rather keep that code path simple, and exercised by the tests.

picoquic/frames.c
```diff
@@ -975,84 +1069,139 @@ int picoquic_prepare_stream_frame(picoquic_cnx_t* cnx, picoquic_stream_head* str
...
-                    stream->send_queue = next;
+                if (length >= allowed_space) {
+                    length = allowed_space;
                 }
 
-                stream->sent_offset += length;
-                cnx->data_sent += length;
-            }
-            *consumed = byte_index;
+                if (length < byte_space) {
```
igorlord: This looks very much like the code above.  And it is a lot of complex logic here.  Refactor into a single function?

igorlord:
Added many code review comments.

huitema:
@igorlord thanks for the comment. I had jumped the gun and merged the PR already, will enter your suggested fixes in another PR.

picoquic/picoquic.h
huitema: I would rather not specify an enum.

In practice, it will most of the type be OK to just set "is_still_active=1". I am mostly concerned with thinks like video streaming, in which the application sends one frame, waits a bit, sends another frame, etc.

picoquic/frames.c
```diff
@@ -904,32 +914,115 @@ uint8_t* picoquic_decode_stream_frame(picoquic_cnx_t* cnx, uint8_t* bytes, const
...
                     break;
                 }
             }
+        };
+
+        /* Do only one pass if we found a stream */
+        if (stream != NULL) {
+            break;
+        }
+    }
```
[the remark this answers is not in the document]
huitema: OK.

picoquic/frames.c
huitema: Yes. There is one slight difference, but it can be accommodated.

picoquic/frames.c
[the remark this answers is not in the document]
huitema: Only problem is that your rewrite returns "start_stream" if no stream is ready. But I fixed that.

Issue #455, opened 2018-12-27
Disable spin bit on 1/8th of connections


huitema:
Even when spin bit is actually enabled, which should be controllable through the API.

huitema:
Done in PR #471

PR #471, opened 2018-12-31
Ensure spin on one eight of connections

huitema:
Also, allow setting of preferred spin bit policy per quic context, and per connection.


huitema:
Resolves issue #455.

Issue #473, opened 2018-12-31
Split data frames before retransmit on lower MTU


huitema:
The current code performs retransmission by copying frames from an old packet to a new one. This only works if the new packet is at least as large as the new one. There is no guarantee of that. For example, after a migration, the new path could have a lower MTU than the previous path.

If frames cannot be copied to the new packet, they should be split into fragments, to be retransmitted later. For example, the fragments could be placed on the "miscellaneous frames" queue. To ensure that there is no further problem, the fragments' sizes should be compatible with minimal MTU packets (1200 bytes).

huitema:
Fixed in Pr #475

PR #475, opened 2019-01-02
better code for frame polling callback

huitema:
Implements a set of code changes suggested by @igorlord when reviewing PR #465.

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -938,11 +938,9 @@ picoquic_stream_head* picoquic_find_ready_stream(picoquic_cnx_t* cnx)
                 (stream->stop_sending_requested && !stream->stop_sending_sent)) {
                 /* Something can be sent */
                 /* if the stream is not active yet, verify that it fits under
-                 * the max stream id limit */
-                 /* Check parity */
+                 * the max stream id limit, which depends of the type of stream */
                 if (IS_CLIENT_STREAM_ID(stream->stream_id) != cnx->client_mode ||
-                    (IS_BIDIR_STREAM_ID(stream->stream_id) && stream->stream_id <= cnx->max_stream_id_bidir_remote) ||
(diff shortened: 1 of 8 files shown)


picoquic/frames.c
```diff
@@ -914,75 +914,93 @@ uint8_t* picoquic_decode_stream_frame(picoquic_cnx_t* cnx, uint8_t* bytes, const
...
                  /* Check parity */
-                if (IS_CLIENT_STREAM_ID(stream->stream_id) == cnx->client_mode) {
-                    if (stream->stream_id <= cnx->max_stream_id_bidir_remote) {
-                        break;
-                    }
-                }
-                else {
+                if (IS_CLIENT_STREAM_ID(stream->stream_id) != cnx->client_mode ||
+                    (IS_BIDIR_STREAM_ID(stream->stream_id) && stream->stream_id <= cnx->max_stream_id_bidir_remote) ||
+                    (!IS_BIDIR_STREAM_ID(stream->stream_id) && stream->stream_id <= cnx->max_stream_id_unidir_remote)) {
```
igorlord: You will probably hate my style, but just for the record, I would had written this as:
```
stream->stream_id <= (IS_BIDIR_STREAM_ID(stream->stream_id) ? cnx->max_stream_id_bidir_remote : cnx->max_stream_id_unidir_remote)
```

picoquic/sender.c
```diff
@@ -84,7 +84,7 @@ int picoquic_mark_active_stream(picoquic_cnx_t* cnx,
     if (ret == 0) {
         if (is_active) {
             /* Verify that no data is queued on the stream */
```
igorlord: Fix the comment.

huitema:
Also added the code to split retransmitted frames if the new send mtu is smaller than the previous one. This addresses issue #473.

picoquic/frames.c
huitema: Yes, I could save a couple lines. I assume that the compiler will optimize all that, so i went for the easy-to-read form, but having a single line is not that bad.

PR #477, opened 2019-01-04
fix bunch of warnings from PVS-studio

huitema:
Thanks @alagoutte for running an analysis of the Picoquic code using VBS. The analysis returned a bunch of warnings, which fall into the following broad categories:

1) Actual bugs. There are a few hard to find issues about errors happening in tests in case of malloc failures, and also a couple of real issues in the production code, including a memory leak.

2) Redundant tests. VBS systematically flags all tests that are "known true" or "know false". I think most compilers do the same tests and optimize out the redundant code, but I fixed these issues in order to diminish the volume of warnings.

3) Guesswork. For example, VBS flags as "underflow" copy operations that only fill part of the target, because that's a frequent error pattern. But it also a very useful pattern in our code, e.g. copying or comparing just some bytes in a message.

4) VBS not getting it. There are a bunch of functions in which the code first tests that everything is OK, then runs code if OK, then tests individual elements to call destructors. That ceates a lot of false warnings in VBS.

In any case, here are a bunch of fixes.


huitema:
That @alagoutte for the second batch of PVS Studio testing. I fixed a couple of residual issues like redundant tests, but nothing indicating actual failures. There are some warnings left. They are due to PVS Studio not actually understanding the code paths, but there does not seem to be any actual error. I could rewrite the code and work around the limitations of the analyzer, but that would create lots of churn.

PR #490, opened 2019-01-19
H3zero dev

huitema:
Client side implementation of HTTP 3. This is a fairly minimal implementation, but it is enough to do a transport test with an HTTP 3 server. Demonstrated interop with lite-speed QUIC (ls-quic).

Server side implementation as well. With these changes, the server behavior depends on the ALPN selected by the client. If the client uses "h3" or "h3-17", the server will respond according to HTTP 3. If the client sets the ALPN to "hq-17", the server will respond per HTTP 0.9. If the server does not recognize the client's ALPN, the choice is to run HTTP 0.9 by default.


huitema:
Checking this in, since we want to start enbaling HTTP 3 interop testing.

PR #496, opened 2019-01-26
Use `last_incoming_packet_time` as next wake for sending data

bkchr:
Given the following structure of calling picoquic:

1. Get current time()
2. Check socket and forward data to picoquic using current time from 1.
3. Check stream structures for data to send and forward to picoquic
4. Get list of connections based on next wake time with max_time set to time from 1.

The current implementation would require a new round, for the connection to send data, as `add_to_stream`/`mark_as_active` were using a time that was higher than the time from 1.

My solution now is to save the time of the last incoming packet, to use it as next wake up time. So, after step 3, the connection already can send data in step 4 and we don't need another round.

### DIFF
--- a/picoquic/packet.c
+++ b/picoquic/packet.c
@@ -63,7 +63,7 @@ int picoquic_parse_packet_header(
             ph->vn = PICOPARSE_32(bytes + 1);
             /* Obtain the connection ID lengths from the byte following the version */
             picoquic_parse_packet_header_cnxid_lengths(bytes[5], &l_dest_id, &l_srce_id);
-            /* Required length: 
+            /* Required length:
              * (packet type(1) + version number(4) + cid_lengths(1) = 6,
              * cid lengths,
              * sequence number (4) */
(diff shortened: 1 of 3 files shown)


bkchr:
The context already has a field for a logic time, that is only used by tests. We could extend this to be used outside of tests as well. We would remove `current_time` from all functions that require it currently and they switch to the logical time of the context. The user needs to ensure to update the logical time of the context, by calling `picoquic_update_time` or something like that. Sounds better than freezing/unfreezing the time.

huitema:
Is this superceded by Pr #499 ?

bkchr:
#499 addressed the case when we already marked a stream as active, the callback requested us to send data and while getting the pointer we set that the stream is still active and wants to send more than what fits into the current buffer.

I'm still not sure about this pr. I could also change my logic and fetch my "streams" for new data, before I get the current time, look into the socket etc. By moving the fetching of the streams, the current implementation would work and I would get the desired behavior. 
I just have the quic stuff in one loop and I just would like to send data in the same iteration, as my streams reported to quic that there is data to send.

Issue #489, opened 2019-01-15
Callback when connection is ready (was: Performance regression?)


bkchr:
I wanted to update my bindings from https://github.com/bkchr/picoquic/commit/afe4397663e6c865718638a0fe139c71e2296729
to latest master. But with latest master my tests are running really slow. I did not had any time yet to look deeper into it, do you maybe implicitly changed something that I miss?

huitema:
@bkchr I don't see performance regression in the current set of tests -- the execution times remained pretty much the same. There was a fix to the newreno implementation that made those tests faster, and after that perf remained pretty much constant.

bkchr:
Okay, sorry for the noise. This was a problem with my rust implementation. I wait until a connection is in ready state, before giving it to the user. In older picoquic versions that does indeed work, because I assume that wake up times was calculate different. Nevertheless, I now switched to `client_ready_start` as the state to wait for :)

huitema:
@bkchr that was the reason why I asked you to review the PR that introduced these two new states, on server and client state. The new state more or less state that this end is ready, but there is no certainty that the other end also is. Specifically:

1) picoquic_state_server_false_start: the server has completed the handshake and computed the session keys, but has not received confirmation that the client is ready. There might be a man-in-the-middle. The server will only move to "picoquic_state_ready" when it receives the OK from the TLS stack, i.e. after the TLS stack has successfully processed the "Client Finished" message.

2) picoquic_state_client_ready_start: the client has completed the handshake and computed the session keys, but does not know yet whether the server is ready. The connection is secure, because the client's TLS stack has processed the "server finished" message, but the client may still need to repeat the handshake messages until the server receives the "Client Finished" message. The client moves to "picoquic_state_ready" when it receives the first acknowledgement on a 1RTT protected packet, which proves that the server can properly receive and process encrypted message.

Once in the "picoquic_state_ready" stage, the peers know that they don't need to process any further Handshake message, and can use a more streamlined processing. The stack waits to reach that stage to enable some functions like key rotation.

bkchr:
I also looked into the code and the `picoquic_ready_start` is only set in the sender, after checking that the corresponding data was received. I wondered why that is not done in the receive frame function. I tested it and put the corresponding code that switch into the ready state into the receive frame function. This worked as well with my implementation of checking for the ready state.
If you also think that the state switch could be put into the receive frame function, I could open a pr.
And sorry for not mentioning it in the corresponding pull request :(

[...]

PR #498, opened 2019-01-27
Transition ready callback

huitema:
This implements the callback discussed in issue #489 -- in fact two callbacks for "almost ready" when the client or the server can start to send data, and "ready" when the connection is verified and advance features like migration can be used.


huitema:
I am working on a complement PR to manage resilience against unexpected Initial or Handshake packets. I will check this one in, then look at moving the transition.

Issue #509, opened 2019-01-31
Local address sometimes not set for incoming connection


bkchr:
Hey, in some cases the local address is not set, when the server is calling with `callback_state_ready`. I can not say which constellation is required to reproduce this. It does not happen that often, I have to let my test running in a loop to get this to happen.

Can we not set the local address as we create the connection context?

huitema:
Bizarre. I would like to understand the circumstances.

Hosts have multiple IP addresses, at least one per network interface: loopback, Ethernet, WiFi, cellular, etc. One set for IPv4, another for IPv6. In fact, for IPv6 there are several addresses per interface: local scope, stable, and possibly several temporary addresses.

The OS chooses which local address to use for a packet based on the combination of destination address, routing tables, and policies such as preferring WiFi over cellular. This may change over time.

The code starts a connection with an unspecified local address, then learns which address was chosen by looking at the first packet that it receives on a path. There may be several paths and addresses used over the duration of a connection.

Normally, the local address should be documented for each path once at least one packet is received on that path.

bkchr:
I already switches to the on_ready callback. If I use the, will be ready callback, my code always crashed at reading the local address. Can we not just take the local address of the initial packet? Looking into the code, just the address the packet is coming from is stored directly. Why can we not also store the local address at this point?

huitema:
Yes, that seems like a bug. We need a fix, and also add a test that the addresses are documented at the time of the callback.

huitema:
@bkchr please check PR #511. I tried to reproduce the bug, but in the normal scenario the addresses appear to be just fine. Can you be more specific about the particular scenario in which the issue happens?

huitema:
Fixed in PR #511.

PR #511, opened 2019-01-31
Add test that addresses are documented during call setup

huitema:
This test tries to reproduce the issue #509 documented by @bkchr. 

The test is valuable in itself as it checks that specific code path, but the issue is not reproduced.

### DIFF
--- a/UnitTest1/unittest1.cpp
+++ b/UnitTest1/unittest1.cpp
@@ -819,6 +819,13 @@ namespace UnitTest1
             Assert::AreEqual(ret, 0);
         }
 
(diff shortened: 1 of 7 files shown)


bkchr:
Could you add this test for the server? For me it was the server failing.

bkchr:
And maybe use the public api to get the address? You are doing the same, but we should test the public api, to not break anything in the future.

bkchr:
And maybe use the public api to get the address? You are doing the same, but we should test the public api, to not break anything in the future.

PR #517, opened 2019-02-03
CID encryption Example

huitema:
The FFX construct allows for encryption of a variable length input, using a modern cipher. This particular implementation is limit to input length <= 31 bytes, which I believe is sufficient for encrypting CID of length 4 to 18.

There are a couple of things that I could consider changing. The construct can use any 128 bit modern cipher, but I only designed it for AES128. We could consider offering ChaCha20 as an option.

The construct take a number of rounds as a parameter. There are theoretical proof that 10 is safe for all applications, 8 for most, but 4 rounds is probably adequate in our case. Each round involves an iteration of the symmetric cipher, so this is clearly a trade-off between safety and performance.

The construct would also allow for "masked bits" that are not encrypted. In the CNXID application, it might be interesting to use the first bit in the CID as a "key phase" bit. This is an easy modification, but not done yet.


huitema:
Added support for selection of CHACHA20 or AES128, specifying the number of rounds, and applying encryption under mask.

huitema:
OK, I realize that this is only useful in some cases, where we want "whole ticket encryption". In other cases, we want to simply encrypt the routing data, assuming that the reminder of the ticket is a set of random bits. For that, we can use something very similar to PN encoding, using AES_CTR or CHACHA20, seeded with the random bit, and only masking the routing bits.

huitema:
@dbavatar, @igorlord this code adds an example (or default value) of a connection ID connection callback that supports encryption. It implements 2 variants:

1) Encrypt of the "unmasked" data only, using the masked data as a seed. This is very similar to PN encoding. It would be an inexpensive way to encrypt the routing data, assuming that the other bits in the CID are random.

2) Encrypt the entire CID. We use the FFX/AES construct to build ciphers of variable size, 4 to 18 bytes. The FFX construct is implemented in picotls, but the code could easily be added. 

In both cases, the encryption key is derived from the "reset seed". It makes sense, since CID and Reset Secret are generated at the same time. 

I did not implement any management of a key identifier. I could do that later. But first, I want to get some feedback and see if this is on the right track.

PR #536, opened 2019-03-17
Win app async

huitema:
Develop a windows application. The quic protocol machinery is running in a background thread, which enables testing the "asynchronous" nature of the API. The current implementation does not include any locks, this should be tested later.


picoquic/quicctx.c
```diff
@@ -1542,6 +1546,8 @@ int picoquic_start_client_cnx(picoquic_cnx_t * cnx)
 {
     int ret = picoquic_initialize_tls_stream(cnx);
 
+    cnx->context_complete = 1;
```
bkchr: Maybe do this via a callback?

huitema: Maybe. I added this "context_complete" test after noticing failures. The UI was starting to create a connection, and the first step in the creation is to add the new connection to the connection list and to the wake up list. Then, the background thread will see that there was new connection and try to send the "initial" message; but the keys were not yet ready, and that means a null pointer de-reference.

I tried deferring the call to "insert in wake list", but it creates another problem: there are a few places that call "reinsert by wake list" during the initialization process. Not only will it place the connections in the list too soon, but calling reinsert on a connection that is not yet inserted messes up the chains of pointers, leading eventually to crashes.

How would you manage that in a callback?

huitema: OK, I am thinking of a potential design. There are a few commands that may synchronization issues. Instead of running them in the UI thread, the UI could post a message to the background thread, such as "create a connection", or "load a document", or "close a connection". The commands often involve some synchronous application action, such as creating the application specific callback context. Those application specific actions could be executed in a callback.

That will work, because all these calls would be serialized in the background thread. But it is still a bit of a cope-out, because I would be swapping on synchronization problem with another. The specific actions might be changing the application state, and running that in the background thread could lead to races, deadlocks, etc.

Still thinking.

bkchr: The rust implementation is also async. What I'm doing is that `picoquic` is running on its own thread and all the data is given/taken from picoquic in this thread. All other threads use channels to send/receive data from this thread.

huitema: @bkchr I think I understand the way to do this for connection creation. It should be something like:
```
User thread                    Background thread

  Call Create_CNX_Async()
  message (includes Callback spec) ---->
                        Creates connection context
                        with call to picoquic_create_cnx()
                        Executes callback
                                  Callback configures
                                  required parameters
                        Activates connection context
  <---- creation complete message
                        Resume thread.
```
I have not found other calls that require this callback structure so far, although the calls to
create streams, add data to stream or close connection should probably be triggered by messages and executed in the context of the background thread.

This does require a messaging system between the user and background thread. The Rust channels provide that in your case. The specific call will probably be dependent on the implementation of the application -- I could see using libevent for that, or some form of Windows messaging for Windows applications.

huitema: @bkchr I have removed the "complete" flag. Instead, the quicwind code is now just a demo on how to use message passing between a windows app and a background thread. It works. You may want to take a look.

PR #554, opened 2019-04-09
Draft 19

### LINKED ISSUE #553: Draft 19 updates
Transport:
1) No automatic version negotiation, except between draft versions. Version negotiation triggers a completely new connection, new initial DCID, etc.
2) Removal of VN blocks from transport parameters:
       TransportParameter TransportParameters<0..2^16-1>;
3) IPv6 flow-ID, but that's a no-op.
4) Wait greater of (3*PTO, idle delay); idle timeout encoded in milliseconds.
5) Detect protocol violation if packet passes decryption and extra bits in header.
6) Check that RETIRE CID does not retire the current CID (protocol violation)
7) Most frames allowed in 0-RTT (not CS, ACK)

HTTP:
1) Uni stream type is varint; C -> 0, P -> 1, reserved: "0x1f * N + 0x21"
2) Frame type is varint, comes before frame length.
3) HTTP_MALFORMED_FRAME only includes frame type if lower than "0xff", uses value "0xff" for all larger types.

huitema:
Close #553 

Implements the changes from draft 18 to draft 19.


huitema:
Verified interop between picoquic client and nghttp2.org using draft-19 and h3-19

PR #561, opened 2019-04-16
Manage a pool of packets per quic context

### LINKED ISSUE #469: Manage pool of packets
The transport code allocates packets in `picoquic_prepare_packet` to hold the clear-text data contained in a packet, and keeps a copy of these packets until they are acknowledged. This requires one malloc/free cycle per packet sent. It would probably be more efficient to manage a pool of packets, and avoid the malloc operations.

huitema:
Close #469

### DIFF
--- a/picoquic/sender.c
+++ b/picoquic/sender.c
@@ -273,6 +273,7 @@ picoquic_packet_t* picoquic_create_packet(picoquic_quic_t * quic)
     }
     else {
         quic->p_first_packet = packet->next_packet;
+        quic->nb_packets_in_pool--;
     }
 
     if (packet != NULL) {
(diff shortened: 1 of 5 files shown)


huitema:
@bkchr Managing a pool of buffers instead of doing malloc/free per packet shows 4 to 5% performance gain on Windows in release mode. I don't think it has an impact on Rust mappings, but just checking...

picoquic/sender.c
```diff
@@ -264,17 +264,32 @@ uint32_t picoquic_pad_to_policy(picoquic_cnx_t * cnx, uint8_t * bytes, uint32_t
...
-        memset(packet, 0, sizeof(picoquic_packet_t));
+        memset(packet, 0, offsetof(struct st_picoquic_packet_t, bytes));
     }
 
     return packet;
 }
 
+void picoquic_recycle_packet(picoquic_quic_t * quic, picoquic_packet_t* packet)
+{
+    if (packet != NULL) {
```
bkchr: Should we maybe introduce an upper bound for the number of packets? Otherwise we maybe create some sort of memory leak?

huitema: Yes, but the question then is where to place the upper bound. Assume 1Gbps connections, 100ms RTT, you need close to 10000 packets. When the connection is up, these packets won't be in the pool, they will be in the retransmit queues. When a connection closes they will move to the pool, ready for the next one. But then how many connections are there?

I suppose we could have an arbitrary limit, like 16K packets or maybe 32K.

huitema: OK, done.

PR #568, opened 2019-04-18
Many stream scenario

huitema:
Add an option to specify "many downloads" in the demo scenario. For example, "*1000:/12345;" will request 1000 downloads of 12345 bytes.
Add logging of TP names, because debugging the "many downloads" trial runs requires looking at parameter values.
Add option to log all packets, instead of just the first 100 packets.
This is part of performance work, to check what happens when a larger number of streams are open.


huitema:
@igorlord you may be interested in the new options in this PR, e.g. the possibility to request a 1000 streams in a single shot. It can make for interesting tests...

PR #582, opened 2019-05-05
Faster ack process and server performance

huitema:
This PR performs most of the obvious performance optimizations on the server side, significantly reducing the cost of packet processing. In Windows release mode, the loopback test runs now at 630 Mbps, 480 Mbps in debug mode. Most of the execution time is spent in the socket code -- 67.76% in SendMsg, 5.55% in RecvMsg.


huitema:
At that point the client is not CPU bound, and the server is only CPU bound for the latter half of the connection. In addition to socket optimization, we probably also need some work on congestion control. Plus, of course, support for large packets.

PR #586, opened 2019-05-12
Stream splay

huitema:
Organized the list of links splay instead of a linked list. This does improve the performance on the "10000 transactions" test, which now executes in just under 1 sec -- i.e. achieving 10,000 transactions per second, which is a nice landmark.

The limiting factor is the client, and especially the time spent in the function `picoquic_find_ready_stream` -- it hits 70% of the CPU, for two reasons:

1) The round robin strategy often requires visiting all 10000 queued streams, even if many streams cannot possibly send anything now
2)  Round robin does not play well with the "splay" logic, which assumes repetitive calls to the same stream. Also, round robin requires looking for "next stream". In a splay that implies walking the tree, and that's not very efficient.

The improvement in TPS comes at a cost for pure speed, because the server spends about 10% of CPU time in `picoquic_find_ready_stream`, again because the splay is less good than a linked list for walking the list. OTOH, the splay works great for finding the stream associated with the received packets.

Still some work to do before checking in. One plausible optimization is to only visit the streams that are authorized to send, i.e. if the stream_id is lower than the max allowed by the peer.


huitema:
Adding a list of "output" connections, defined as the list of connections that have an open stream. This mitigates the client side effect of having to search through a large list of streams. The 100M download tests gets 590 Mbps, i.e.stable perf. The10K transaction test scores 30000 transactions per second.

There are still two possible improvements -- do not vistit the whole splay list when receiving a "MAX STREAMS" frame, and possibly prune the "output" list to only include streams that can actually write data.

huitema:
Fixed the MAX Streams processing to only scale with the number of new streams,not the whole set of streams.

PR #591, opened 2019-05-20
Datagram

huitema:
Implement support for the datagram frame as defined in https://datatracker.ietf.org/doc/draft-pauly-quic-datagram/?include_text=1


bkchr:
Will you add an example as well?

PR #600, opened 2019-05-23
Fix version negotiation log and notify

### LINKED ISSUE #594: VNEGO not logged properly
Also, there probably should be a callback to the application to provide the result.

huitema:
Close #594


huitema:
@bkchr This PR adds a callback item signaling a version negotiation. You may want to expose it in the Rust API.

PR #604, opened 2019-06-21
Basic support for ESNI

huitema:


huitema:
This implements support for Encrypted SNI in Picoquic. Some of the tests depend on the ["ESNI for Windows" PR in Picotls](https://github.com/h2o/picotls/pull/244). The additional unit test verifies that the connection succeeds.

The current implementation is somewhat basic. It supports only on ESNI private key and a single ESNI record, both passed by file names. We may want an option to handle multiple keys and multiple records, in order to support key rotation. We may also want to pass the parameters by value, instead of using a file name. But this will be for a follow-on PR...

PR #605, opened 2019-06-29
Draft 21 fixes

huitema:
Update to support the upcoming draft-21 version.


huitema:
This is almost ready, but we need to test the "retire before" logic that comes with the new definition of the New Connection ID frame. Plus understand why the fuzz test fails in the Windows 32 bits configurations.

PR #617, opened 2019-08-02
Fix size_t vs. uint32_t conversions

steschu77:
This change uses size_t for most of the length variables in picoquic.

It fixes some implicit conversions in arithmetic expressions (calculated in uint32_t and then converted to size_t). It also avoids some explicit size_t -> uint32_t conversions.

### DIFF
--- a/picoquic/logger.c
+++ b/picoquic/logger.c
@@ -461,14 +461,14 @@ void picoquic_log_packet_header(FILE* F, uint64_t log_cnxid64, picoquic_packet_h
         picoquic_log_connection_id(F, &ph->dest_cnx_id);
         fprintf(F, ", ");
         picoquic_log_connection_id(F, &ph->srce_cnx_id);
-        fprintf(F, ", Seq: %d, pl: %d\n", ph->pn, ph->pl_val);
+        fprintf(F, ", Seq: %d, pl: %zd\n", ph->pn, ph->pl_val);
         if (ph->ptype == picoquic_packet_initial) {
             picoquic_log_prefix_initial_cid64(F, log_cnxid64);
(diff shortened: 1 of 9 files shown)


huitema:
I understand the desire for uniformity, but I am not sure that we need to convert every "length" field to `size_t`. The `size_t` type has a number of particularities that make it sometimes undesirable. In particular, `sizeof(size_t)` returns different values depending on compiler, options and platforms. The definition of `size_t` is "enough bits to encode any memory object that can be defined on the system". Sometimes we need that, and sometimes we don't.

My philosophy was to use more constrained types when possible. In particular, packet lengths are typically much more constrained than "size_t", which is why I moved to use "uint32_t" in many places.

huitema:
@steschu77 In any case, please look at the outputs of the integration tests. Changing the type of length fields has cascading effects. The current break from uniformly using size_t was in part an effort to contain these effects. We need to make sure that all the tests run before changing the definitions. Maybe you want to apply some priorities, e.g. apply uniformity to the published APIs in `picotls.h` but be more opportunistic for the private APIs.

steschu77:
Thanks @huitema for the fast feedback and the philosophy behind picoquic's usage of size_t. I actually agree with all you are saying.

I tried to apply `size_t` only when it was used to specify memory limits. Packet lengths, for example, are one of those.

I started this change after seeing Visual Studio complaining about some calculations where the calculation was done in the smaller type and then converted to the bigger type.

I'll fix the tests a.s.a.p. Question: I've 22 failing tests even without my changes. Is that a known situation?

huitema:
The tests are normally all passing. I check them before merging any PR. Please check whether the move to size_t had some side effect. On what platform are they failing?

PR #618, opened 2019-08-03
Adding more debug of a flaky test

huitema:


steschu77:
Thanks for looking into it. Looks good to me.

huitema:
The reason why this test sometimes breaks is the randomized behavior introduced in "sender.c" in `picoquic_insert_hole_in_send_sequence_if_needed()`:
~~~
        /* Predict the next hole*/
        cnx->pkt_ctx[0].next_sequence_hole = cnx->pkt_ctx[picoquic_packet_context_application].send_sequence + 1 + picoquic_public_uniform_random(cnx->quic->sequence_hole_pseudo_period);
~~~
The "uniform random" call returns unpredictable values. This means a different code path for each test run. There is at least one case in which this leads to an error, but this is hard to reproduce.

huitema:
After making sure that the random number is seeded the same way for the test on each platform, all the tests are passing. We can verify that exactly the same numbers are skipped on each platform. However, the execution differs:

OS and compiler | Result
------- | ------------
Linux gcc8 | Complete after 776 packets sent, 769 r. by client, 5 retransmits.
Linux clang | Complete after 1008 packets sent, 1008 r. by client, 221 retransmits.
Linux scan-build | Complete after 823 packets sent, 822 r. by client, 47 retransmits.
Mac OS xcode | Complete after 770 packets sent, 769 r. by client, 0 retransmits.
Mac OS xcode/clang | Complete after 770 packets sent, 769 r. by client, 0 retransmits.
Windows VS17 x64 debug | Complete after 770 packets sent, 769 r. by client, 0 retransmits.
Windows VS17 x64 release | Complete after 770 packets sent, 769 r. by client, 0 retransmits.
Windows VS17 x86 debug | Complete after 770 packets sent, 769 r. by client, 0 retransmits.
Windows VS17 x86 release | Complete after 801 packets sent, 801 r. by client, 25 retransmits.

On some platforms, there are 0 retransmissions and the code works exactly as expected. But on other platforms we see retransmissions. There is something happening to cause a difference, and that needs to be fixed.

PR #619, opened 2019-08-05
Separate http module

steschu77:
Move HTTP related code into separate modules. The picoquic module should only be about the QUIC protocol and should not contain applications of it.

This PR is still WIP. Please do not merge (yet).


picohttp_t/picohttp_t.c
```diff
@@ -0,0 +1,330 @@
+/*
```
steschu77: I'd do a cleanup of test applications in a later change. For now I think it's okay to go ahead with duplicated code. @huitema would that be acceptable?

picohttptest/picohttptest.h
```diff
@@ -0,0 +1,186 @@
...
+* DISCLAIMED. IN NO EVENT SHALL Private Octopus, Inc. BE LIABLE FOR ANY
+* DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+* (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+* LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+* ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+* (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+* SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+#ifndef PICOQUICTEST_H
```
steschu77: `PICOHTTPTEST_H`

picohttptest/picohttptest_internal.h
```diff
@@ -0,0 +1,220 @@
...
+* DISCLAIMED. IN NO EVENT SHALL Private Octopus, Inc. BE LIABLE FOR ANY
+* DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+* (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+* LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+* ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+* (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+* SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+#ifndef PICOQUICTEST_INTERNAL_H
```
steschu77: `PICOHTTPTEST_INTERNAL`

huitema:
This is nice. Separate libraries will avoid name collisions for developers using different http stacks.

Just two requests. I am fine with moving the execution of the http tests from "picoquic_t" to "picohttp_t", but I would rather leave them in the visual studio test project. Otherwise, the test discovery gets very messy. I would also make sure that the "travis.yml" script calls the new "picohttp_t" test and verifies that it works.

The http code is used for demos, but also for tests of the underlying stack. We need to verify that it keeps working with every update, hence the need to integrate with Visual Studio so we can test everything after any change, and also the need to integrate with Travis so we very that it does not get broken by a check-in.

steschu77:
Thanks for the feedback, Christian. I'll undo the picoquictest separation
and update the PR asap.

PR #624, opened 2019-08-07
port to latest windows sdk

huitema:


steschu77:
That is awesome! Thanks! No need to exclude local SDK version changes from vcxproj anymore ;)

PR #626, opened 2019-08-08
Use picoquic_file_open instead of #ifdef

steschu77:


huitema:
Yes of course. I should do have done a global replacement of these ifdef's a long time ago...

PR #627, opened 2019-08-08
Try fixes on initial fuzzer.

huitema:
Fuzz_initial fails randomly at times on Windows "release" builds, but not on Windows "debug" builds. This points to a possible memory initialization issue, which we need to find.

### DIFF
--- a/picoquic/logger.c
+++ b/picoquic/logger.c
@@ -1821,7 +1821,7 @@ void picoquic_open_cc_dump(picoquic_cnx_t * cnx)
         const size_t suffix_len = strlen(suffix);
         size_t folder_length = strlen(cnx->quic->cc_log_dir);
 
-        if (folder_length + 1 + 2*cnx->initial_cnxid.id_len + suffix_len + 1 > sizeof(cc_log_file_name)) {
+        if (folder_length + (size_t)1 + ((size_t)2)* (size_t)cnx->initial_cnxid.id_len + suffix_len + (size_t)1 > sizeof(cc_log_file_name)) {
             ret = -1;
         }
         else {
--- a/picoquic/packet.c
+++ b/picoquic/packet.c
@@ -67,13 +67,13 @@ int picoquic_is_old_header_invariant(
         l1 = 5 + 1 + l_dest_id1 + l_srce_id1;
         if (l1 > length) {
             ret = 0;
-        } else if (l_dest_id2 > 20u || 5u + l_dest_id2 + 1u > length) {
+        } else if (l_dest_id2 > (size_t)20 || (size_t)5 + (size_t)l_dest_id2 + (size_t)1 > length) {
             /* if packet is old invariant, this test will succeed whenever the
              * destination id is longer than 4, which covers a lot of cases */
             ret = 1;
         }
         else {
-            l_srce_id2 = bytes[5 + l_dest_id2 + 1];
+            l_srce_id2 = bytes[(size_t)5 + (size_t)l_dest_id2 + (size_t)1];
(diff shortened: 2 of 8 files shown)


picoquic/logger.c
```diff
@@ -1821,7 +1821,7 @@ void picoquic_open_cc_dump(picoquic_cnx_t * cnx)
         const size_t suffix_len = strlen(suffix);
         size_t folder_length = strlen(cnx->quic->cc_log_dir);
 
-        if (folder_length + 1 + 2*cnx->initial_cnxid.id_len + suffix_len + 1 > sizeof(cc_log_file_name)) {
+        if (folder_length + (size_t)1 + ((size_t)2)* (size_t)cnx->initial_cnxid.id_len + suffix_len + (size_t)1 > sizeof(cc_log_file_name)) {
```
steschu77: Arithmetic operators in the C language are left-to-right associative, the first argument determines the type the operation is performed on. That being said, (size_t)1 shouldn't be necessary and it should be sufficient to have:
```
if (folder_length + 1 + (size_t)cnx->initial_cnxid.id_len * 2 + suffix_len + 1 > sizeof(cc_log_file_name)) {
```

picoquic/packet.c
```diff
@@ -359,7 +359,7 @@ int picoquic_parse_packet_header(
 
          if ((int)length >= 1 + cnxid_length) {
              /* We can identify the connection by its ID */
-             ph->offset = (uint32_t)( 1 + picoquic_parse_connection_id(bytes + 1, cnxid_length, &ph->dest_cnx_id));
+             ph->offset = (uint32_t)((size_t)1 + picoquic_parse_connection_id(bytes + 1, cnxid_length, &ph->dest_cnx_id));
```
steschu77: The final cast to (uint32_t) looks dangerous.

picoquic/picoquic_internal.h
```diff
@@ -439,7 +439,7 @@ typedef struct st_picoquic_stream_head_t {
 #define IS_CLIENT_STREAM_ID(id) (unsigned int)(((id) & 1) == 0)
 #define IS_BIDIR_STREAM_ID(id)  (unsigned int)(((id) & 2) == 0)
 #define IS_LOCAL_STREAM_ID(id, client_mode)  (unsigned int)(((id)^(client_mode)) & 1)
-#define STREAM_ID_FROM_RANK(rank, is_server_stream, is_unidir) (((rank-1ull)<<2)|(uint64_t)((is_unidir)<<1)|(is_server_stream))
+#define STREAM_ID_FROM_RANK(rank, is_server_stream, is_unidir) ((((uint64_t)(rank)-(uint64_t)1)<<2)|(((uint64_t)is_unidir)<<1)|((uint64_t)(is_server_stream)))
```
steschu77: What is wrong with 1ull?

picoquic/logger.c
huitema: Yes, I probably went too far there.

picoquic/picoquic_internal.h
huitema: I have seen systems for which "unsigned long long" is not equivalent to "uint64_t". So I would rather use the keyboard a bit more than take risks.

picoquic/packet.c
huitema: Yes, that code dates from before your PR redefining offset as size_t. Will fix.

picoquic/picoquic_internal.h
steschu77: You're right. I have mistaken "unsigned long long" to be equivalent to "uint64_t".

PR #629, opened 2019-08-10
Move and use universal code to/from util.c for cc file open

steschu77:
- Move sprintf to util.c
- Move connection id to string to util.c
- Add unit tests for util.c

### DIFF
--- a/picoquic/logger.c
+++ b/picoquic/logger.c
@@ -1826,7 +1826,7 @@ void picoquic_open_cc_dump(picoquic_cnx_t * cnx)
 
     char cc_log_file_name[512];
     if (ret == 0) {
-        ret = picoquic_sprintf(cc_log_file_name, sizeof(cc_log_file_name), "%s%c%s-log.bin", cnx->quic->cc_log_dir, PICOQUIC_FILE_SEPARATOR, cnxid_str) <= 0;
+        ret = picoquic_sprintf(cc_log_file_name, sizeof(cc_log_file_name), "%s%c%s-log.bin", cnx->quic->cc_log_dir, PICOQUIC_FILE_SEPARATOR, cnxid_str);
     }
 
     if (ret != 0) {
--- a/picoquic/logger.c
+++ b/picoquic/logger.c
@@ -1821,35 +1821,34 @@ void picoquic_open_cc_dump(picoquic_cnx_t * cnx)
         return;
     }
 
-    char cnxid_str[2 * PICOQUIC_CONNECTION_ID_MAX_SIZE + 1];
-    int ret = picoquic_print_connection_id_hexa(cnxid_str, sizeof(cnxid_str), &cnx->initial_cnxid);
-
     char cc_log_file_name[512];
-    if (ret == 0) {
-        ret = picoquic_sprintf(cc_log_file_name, sizeof(cc_log_file_name), "%s%c%s-log.bin", cnx->quic->cc_log_dir, PICOQUIC_FILE_SEPARATOR, cnxid_str);
-    }
+    char cnxid_str[2 * PICOQUIC_CONNECTION_ID_MAX_SIZE + 1];
+    int ret = 0;
 
-    if (ret != 0) {
+    if (picoquic_print_connection_id_hexa(cnxid_str, sizeof(cnxid_str), &cnx->initial_cnxid) != 0
+        || picoquic_sprintf(cc_log_file_name, sizeof(cc_log_file_name), "%s%c%s-log.bin", cnx->quic->cc_log_dir, PICOQUIC_FILE_SEPARATOR, cnxid_str) != 0)
+    {
         DBG_PRINTF("Cannot format file name into folder %s, id_len = %d\n", cnx->quic->cc_log_dir, cnx->initial_cnxid.id_len);
-        return;
-    }
-
-    cnx->cc_log = picoquic_file_open(cc_log_file_name, "wb");
-    if (cnx->cc_log == NULL) {
-        DBG_PRINTF("Cannot open file %s for write.\n", cc_log_file_name);
         ret = -1;
-    } else {
-        /* Write a header text with three words: magic number, number of integers, and current date  */
-        uint32_t header[3];
+    }
+    else {
+        cnx->cc_log = picoquic_file_open(cc_log_file_name, "wb");
+        if (cnx->cc_log == NULL) {
+            DBG_PRINTF("Cannot open file %s for write.\n", cc_log_file_name);
+            ret = -1;
+        } else {
+            /* Write a header text with three words: magic number, number of integers, and current date  */
+            uint32_t header[3];
                 
-        header[0] = PICOQUIC_LOG_CC_MAGIC;
-        header[1] = PICOQUIC_LOG_CC_NB;
-        header[2] = (uint32_t)(picoquic_current_time() / 1000000ll);
+            header[0] = PICOQUIC_LOG_CC_MAGIC;
+            header[1] = PICOQUIC_LOG_CC_NB;
+            header[2] = (uint32_t)(picoquic_current_time() / 1000000ll);
 
-        if (fwrite(header, 3 * sizeof(uint32_t), 1, cnx->cc_log) <= 0) {
-            DBG_PRINTF("Cannot write header for file %s.\n", cc_log_file_name);
-            fclose(cnx->cc_log);
-            cnx->cc_log = NULL;
+            if (fwrite(header, 3 * sizeof(uint32_t), 1, cnx->cc_log) <= 0) {
+                DBG_PRINTF("Cannot write header for file %s.\n", cc_log_file_name);
+                fclose(cnx->cc_log);
+                cnx->cc_log = NULL;
+            }
         }
     }
 }
--- a/picoquic/logger.c
+++ b/picoquic/logger.c
@@ -1811,73 +1811,45 @@ void picoquic_log_picotls_ticket(FILE* F, picoquic_connection_id_t cnx_id,
 
 void picoquic_open_cc_dump(picoquic_cnx_t * cnx)
 {
-    int ret = 0;
-
     if (cnx->cc_log != NULL) {
         DBG_PRINTF("%s", "CC LOG File is already open!\n");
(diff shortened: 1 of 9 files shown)


picoquic/logger.c
```diff
@@ -1811,73 +1811,45 @@ void picoquic_log_picotls_ticket(FILE* F, picoquic_connection_id_t cnx_id,
...
+        return;
+    }
 
-            if (ret == 0) {
-                memcpy(&cc_log_file_name[byte_index], suffix, suffix_len);
-                byte_index += suffix_len;
-                cc_log_file_name[byte_index] = 0;
-            }
+    if (cnx->quic->cc_log_dir == NULL) {
+        DBG_PRINTF("%s", "CC LOG directory not set!\n");
```
steschu77: This trace comes now pretty often during unit tests. I'll fix that in a later PR.

picoquic/util.c
```diff
@@ -164,6 +164,35 @@ int debug_printf_reset(int suspended)
...
+    va_list args;
+    va_start(args, fmt);
+    int res = vsnprintf(buf, buf_len, fmt, args);
+    va_end(args);
+
+    // vsnprintf returns <0 for errors and >=0 for nb of characters required.
+    // We return 0 when printing was successful.
+    return res >= 0 ? (res >= buf_len) : res;
+}
+
```
huitema: With this, you can remove a bunch of remove a bunch of `#ifdef` from the code, and that is very good. But the problem with `vsnprintf` is the same as with `sprintf`. According to https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/vsnprintf-vsnprintf-vsnprintf-l-vsnwprintf-vsnwprintf-l?view=vs-2019, _More secure versions of these functions are available; see vsnprintf_s, _vsnprintf_s, _vsnprintf_s_l, _vsnwprintf_s, _vsnwprintf_s_l._

That means you still need an `#ifdef _WINDOWS` here, to use the secure version on Windows.

picoquic/util.h
```diff
@@ -69,6 +72,19 @@ int picoquic_get_input_path(char * target_file_path, size_t file_path_max, const
 #define MAX(a, b) ((a) > (b) ? (a) : (b))
 #endif
 
+/* Return a directory path based on solution dir and file name */
+#ifdef _WINDOWS
```
huitema: Wrong comment. That should be a comment for the function, not for the separator definitions.

picoquic/logger.c
```diff
@@ -1811,73 +1811,45 @@ void picoquic_log_picotls_ticket(FILE* F, picoquic_connection_id_t cnx_id,
...
-            for (size_t i = 0; ret == 0 && i < cnx->initial_cnxid.id_len; i++) {
-#ifdef _WINDOWS
-                ret = sprintf_s(&cc_log_file_name[byte_index], sizeof(cc_log_file_name) - byte_index, "%02x", cnx->initial_cnxid.id[i]) <= 0;
-#else
-                ret = sprintf(&cc_log_file_name[byte_index], "%02x", cnx->initial_cnxid.id[i]) <= 0;
-#endif
-
-                byte_index += 2;
-            }
+        return;
```
huitema: I could not find where the deleted code went. It is actually used on the server side, where we want one cc log file per connection. The convention is to name that file based on the initial CID of the connection. This will then enable retrieval of the trace by the tester, using something similar to the support for connection logs at https://test.privateoctopus.com/.

steschu77:
> I like the general idea of reducing the number of `#ifdef` in the code, and it is also a good idea to write tests for the utilities. But I have a couple issues with your proposal:
> 
> 1. Minor: you do need an `#ifdef _WINDOWS` to call the safe function in the proposed replacement for `sprintf`
> 2. Discuss: the design was to have one CC file per connection, named by default by the connection ID. Are you proposing to replace that?

1. I'll change it to using vsnprintf_s.
2. The proposal should not change any behavior of the code. The connection ID is now written within the sprintf call.

picoquic/logger.c
steschu77: The deleted code mostly went into the single sprintf call now.

picoquic/logger.c
```diff
@@ -1811,73 +1811,45 @@ void picoquic_log_picotls_ticket(FILE* F, picoquic_connection_id_t cnx_id,
...
+    }
```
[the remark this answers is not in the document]
steschu77: I imagined that is the reason for the sparely used returns. While I totally agree to that I'd prefer returns for parameter checks, i.e. no resources are allocated yet that need to be removed. The reason for that is readability of the code, especially to avoid moving the code too far into the "right" direction.

picoquic/logger.c
huitema: I don't understand. There should be a branch here thatgoes something like:
~~~
if file is already open
    do nothing
else
    format file name in selected directory using sprintf
    open file
~~~
I don't see that code anymore.

picoquic/logger.c
```diff
@@ -1811,73 +1811,45 @@ void picoquic_log_picotls_ticket(FILE* F, picoquic_connection_id_t cnx_id,
...
-                DBG_PRINTF("Cannot format file name into folder %s, id_len = %d\n", cnx->quic->cc_log_dir, cnx->initial_cnxid.id_len);
-            }
-        }
+    char cnxid_str[2 * PICOQUIC_CONNECTION_ID_MAX_SIZE + 1];
+    int ret = picoquic_print_connection_id_hexa(cnxid_str, sizeof(cnxid_str), &cnx->initial_cnxid);
 
-        if (ret == 0) {
+    char cc_log_file_name[512];
+    if (ret == 0) {
+        ret = picoquic_sprintf(cc_log_file_name, sizeof(cc_log_file_name), "%s%c%s-log.bin", cnx->quic->cc_log_dir, PICOQUIC_FILE_SEPARATOR, cnxid_str);
```
steschu77: This is the place that builds the while filename at once including directory, connection id, and suffix.

picoquic/logger.c
huitema: The API is designed  for requesting logs on the server side. That cannot be done on a per connection basis. Setting the cc_log directory signals that the server should create one file per connection in that directory.

picoquic/logger.c
```diff
@@ -1811,73 +1811,45 @@ void picoquic_log_picotls_ticket(FILE* F, picoquic_connection_id_t cnx_id,
...
 
-            if (ret == 0) {
-                memcpy(&cc_log_file_name[byte_index], suffix, suffix_len);
-                byte_index += suffix_len;
-                cc_log_file_name[byte_index] = 0;
-            }
+    if (cnx->quic->cc_log_dir == NULL) {
+        DBG_PRINTF("%s", "CC LOG directory not set!\n");
+        return;
+    }
```
huitema: As I mentioned above, I try to avoid return statements mid function. I would prefer having an "else" block. Also, it results in declaring variables at the beginning of the block, which some C compiler used to insist on.

picoquic/logger.c
steschu77: I didn't change that. That's why it is not in the diff, but it is still in the code.

picoquictest/util_test.c
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    // Test invalid call
+    if (picoquic_print_connection_id_hexa("", 0, &expected_cnxid[0]) == 0) {
+        return -1;
+    }
+
+    return 0;
+}
+
```
huitema: This is really a matter of style, but I would rather use a return variable than mid block returns. Also, when writing tests, it is nice to have DBG_PRINTF explaining which subtest failed.

picoquic/logger.c
steschu77: See the full file here:
https://github.com/steschu77/picoquic/blob/refactor_open_cc_log/picoquic/logger.c#L1812

picoquictest/util_test.c
```diff
@@ -0,0 +1,93 @@
...
+        return -1;
+    }
+    if (picoquic_sprintf(str, sizeof(str), "%s%c%s", "fooo", PICOQUIC_FILE_SEPARATOR, "bar") == 0) {
+        return -1;
+    }
+    if (picoquic_sprintf(str, sizeof(str), "%s%c%s", "fooo", PICOQUIC_FILE_SEPARATOR, "barr") == 0) {
+        return -1;
+    }
+    return 0;
+}
```
huitema: Have you considered adding these tests to the unit tests programs, so they be verified on different ports?

picoquic/logger.c
steschu77: The code still does the same thing it did before. This is meant to be a refactoring only without changing any behavior.

The library still creates a log file per connection when the cc_log_dir is set to != NULL. My comment was just a reminder that the unittests, that don't set a directory, now complain a lot. I fixed that on another branch already.

picoquic/logger.c
huitema: OK. I need to look at the whole code after you are done. If the "trace" test works, then this definitely work.

picoquic/logger.c
huitema: OK. I was not seeing the complete picture.

picoquic/util.c
huitema: OK.

picoquictest/util_test.c
steschu77: I added these tests to picoquic_t / picoquic_ct. Are there other places I should consider? (I haven't added anything to UnitTest1. I thought it catches additions automatically).

huitema:
I like this work. We do have a difference of opinion about coding style. This is a very old debate. We were having it often at Microsoft, with proponents of indent, goto fin, carrying a status, and early returns. The general consensus was to avoid the "goto fin" construct, except maybe in test code, and to also avoid early return, except maybe in an initial parameter validation block, before doing any actual code. There may not be problems with early returns when you write them, but you have to consider code maintenance by multiple developers over the years, and strict rules like that end up helping. But yes, you can end up with too much right shift -- more than 3 or 4 is probably not the best idea. In that case, either break in multiple functions or carry a status.

picoquictest/util_test.c
steschu77: Done.

PR #630, opened 2019-08-10
Fewer ifdef windows

huitema:
Use the newly defined picoquic_file_open and picoquic_sprint and reduce the number of "#ifdef _WINDOWS" branches.

### DIFF
--- a/UnitTest1/unittest1.cpp
+++ b/UnitTest1/unittest1.cpp
@@ -42,6 +42,26 @@ namespace UnitTest1
             Assert::AreEqual(ret, 0);
 	    }
 
+        TEST_METHOD(connection_id_print)
+        {
+            int ret = util_connection_id_print_test();
+
+            Assert::AreEqual(ret, 0);
+        }
+
+        TEST_METHOD(connection_id_parse)
+        {
+            int ret = util_connection_id_parse_test();
+
+            Assert::AreEqual(ret, 0);
+        }
+
+        TEST_METHOD(sprintf)
+        {
+            int ret = util_sprintf_test();
+
+            Assert::AreEqual(ret, 0);
+        }
         TEST_METHOD(random_tester)
         {
             int ret = random_tester_test();
--- a/picoquic/util.c
+++ b/picoquic/util.c
@@ -51,7 +51,7 @@ char* picoquic_string_create(const char* original, size_t len)
             }
             else if (allocated > len) {
                 memcpy(str, original, len);
-                str[len] = 0;
+                str[allocated - 1] = 0;
             }
             else {
                 /* This could happen only in case of integer overflow */
@@ -183,7 +183,7 @@ int picoquic_sprintf(char* buf, size_t buf_len, const char* fmt, ...)
 int picoquic_print_connection_id_hexa(char* buf, size_t buf_len, const picoquic_connection_id_t * cnxid)
 {
     static const char hex_to_char[16] = { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f' };
-    if (buf_len < cnxid->id_len * 2u + 1u) {
+    if (buf_len < ((size_t)cnxid->id_len) * 2u + 1u) {
         return -1;  
     }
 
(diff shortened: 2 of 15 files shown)


picoquic/util.c
```diff
@@ -51,7 +52,7 @@ char* picoquic_string_create(const char* original, size_t len)
             }
             else if (allocated > len) {
                 memcpy(str, original, len);
-                str[len] = 0;
+                str[allocated - 1] = 0;
```
steschu77: If the intend is to set the character right after 'len' to '\0', I'd prefer the original version. The new version reads more like setting the last allocated character to '\0', which doesn't necessarily mean it's the character right after the copied string if you don't look at the whole function.

picoquic/util.c
```diff
@@ -458,19 +463,32 @@ int picoquic_get_input_path(char * target_file_path, size_t file_path_max, const
...
     F = fopen(file_name, flags);
+    if (F == NULL && last_err != NULL) {
+        *last_err = errno;
+    }
 #endif
 
     return F;
+}
+
+FILE* picoquic_file_open(char const* file_name, char const* flags)
```
steschu77: Does C support default initialized parameters? If so, we could do
```
FILE * picoquic_file_open(char const * file_name, char const * flags, int * last_err = NULL)
```

UnitTest1/unittest1.cpp
```diff
@@ -42,6 +42,26 @@ namespace UnitTest1
             Assert::AreEqual(ret, 0);
 	    }
 
+        TEST_METHOD(connection_id_print)
```
steschu77: Ah, good to know these need to be added here as well :)

picoquictest/skip_frame_test.c
```diff
@@ -543,35 +543,20 @@ static int compare_lines(char const* b1, char const* b2)
 
 int picoquic_test_compare_files(char const* fname1, char const* fname2)
 {
-    FILE* F1 = NULL;
+    FILE* F1;
```
steschu77: I'd leave the default initialization to NULL. You never know. It's better to have the code to fail immediately than some random undefined behavior. The compiler will figure out that the default initialization is not necessary anyhow. With GCC that is the case at -O1 aleady, see here: https://godbolt.org/z/091yDs

picoquictest/util_test.c
```diff
@@ -90,21 +90,23 @@ int util_connection_id_parse_test()
...
-        DBG_PRINTF("%s", "'foobar' test failed.\n");
+    if (picoquic_sprintf(str, sizeof(str), NULL, "%s%s", "foo", "bar") != 0) {
+        DBG_PRINTF("%s", "'foobar' test failed.");
         ret = -1;
     }
-    if (picoquic_sprintf(str, sizeof(str), "%s%c%s", "foo", PICOQUIC_FILE_SEPARATOR, "bar") != 0) {
-        DBG_PRINTF("%s", "'foo/bar' test failed.\n");
+    if (picoquic_sprintf(str, sizeof(str), &nb_chars, "%s%c%s", "foo", PICOQUIC_FILE_SEPARATOR, "bar") != 0 ||
+        nb_chars != 7) {
+        DBG_PRINTF("'foo/bar' test failed. Ret = %d, Nb_chars = %d", ret, (int)nb_chars);
```
steschu77: `ret` is not the result of the `picoquic_sprintf` call. It's either 0 or -1 depending on the previous test.

picoquic/util.c
huitema: That was the only way to silence the static analyzer of Visual Studio. I think there is something subtle going on with possible integer overflow.

picoquic/util.c
huitema: I don't think that would fly through all C compilers...

picoquictest/skip_frame_test.c
huitema: Not sure. That will also get you a warning, "value is set but never used".

picoquictest/util_test.c
huitema: Thanks. Fixing that.

picoquic/util.c
steschu77: Yep. Tried it, Visual Studio already complains :)

picoquic/util.c
steschu77: I think the Analyzer is just broken. If you try this obviously wrong code, it doesn't complain anymore:
```
        str = (char*)malloc(len);
        if (str != NULL) {
            memcpy(str, original, len);
            str[len] = 0;
        }
```

steschu77:
I think we can merge this.

picoquic/util.c
huitema: I am a bit gun shy there. In Windows OS development, we used an internal version of the analyzer called Prefix -- similar to what ships in VS, but more capable. There were a few incidents were a bug was found "in the field", and postmortem inquiry showed that it was flagged by the static analyzer, but the developer discarded the warning as noise from a broken analyzer. Are you sure that the analyzer was actually running in your test? You need to have the "Control Flow Guard" option set to "Yes" in C/C++ "All options".

steschu77: Yes, "Control Flow Guard" option set to "Yes" in C/C++ "All options". I worked in util.c all the time and changed the code of `picoquic_string_create()`. You should be able to reproduce this by changing the code to the above version. Btw. I am running VS 2019 but with VS 2017 tool chain.

PR #632, opened 2019-08-12
More utils

steschu77:
Only minor changes

### DIFF
--- a/picoquic/util.c
+++ b/picoquic/util.c
@@ -487,8 +487,15 @@ FILE * picoquic_file_open_ex(char const * file_name, char const * flags, int * l
 
     return F;
 }
-
 FILE* picoquic_file_open(char const* file_name, char const* flags)
 {
     return picoquic_file_open_ex(file_name, flags, NULL);
+}
+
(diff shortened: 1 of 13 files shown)


picoquic/util.c
```diff
@@ -487,8 +490,15 @@ FILE * picoquic_file_open_ex(char const * file_name, char const * flags, int * l
...
\ No newline at end of file
+}
+
+/* Safely close files in a portable way */
+void picoquic_file_close(FILE * F)
+{
+    if (F != NULL) {
+        (void)fclose(F);
+    }
+}
```
huitema: I wonder whether this should return `FILE * NULL`, so we can then write`F = picoquic_file_close(F)`, instead of
~~~
picoquic_file_close(F)
F = NULL
~~~
Or maybe use a macro:
~~~
#define PICOQUIC_FILE_CLOSE(F) {if(F != NULL){ fclose(F); F=NULL;}}
~~~

steschu77:
That is a great idea! I'll change that tomorrow.

PR #634, opened 2019-08-18
Avoid uint32 casts

steschu77:
Some more cleanup after the change to size_t for memory buffer sizes/lengths.


picoquic/sender.c
```diff
@@ -1385,7 +1385,6 @@ size_t picoquic_prepare_mtu_probe(picoquic_cnx_t* cnx,
     size_t length = header_length;
 
     bytes[length++] = picoquic_frame_type_ping;
-    bytes[length++] = 0;
```
steschu77: Please verify this as it seemed superfluous to do the extra assignment.

The semantic change without the extra assignment should just be the case `probe_length - checksum_length - length = 0`. That should have crashed before, now it doesn't assign any 0. I assume that case never happens as it has been checked by the calling function.

huitema: Yes. The probe is never smaller than the 1200 bytes min MTU size, and thus the issue never arises.

Issue #564, opened 2019-04-17
Select and RecvMsg are costly on Windows


huitema:
Client loopback test shows that 32% of CPU time is spent in the "select" (22.5%) and "recvmsg" (9.5%) calls for the Windows client. Not clear whether the same issues also happen in Unix builds.

huitema:
On the server side, "send through server sockets" accounts for 50% of CPU, and "select + receive" for another 10%

huitema:
On linux, there is now an option to do "GSO offload" for UDP sockets. See Fastly's blog: https://www.fastly.com/blog/measuring-quic-vs-tcp-computational-efficiency

huitema:
Fastly's paper cites this paper: https://lwn.net/Articles/188489/

huitema:
The equivalent to the UDP GSO option in windows is UDP_SEND_SEGMENT_SIZE or UDP_RECV_MAX_COALESCED_SIZE. Same behavior as GSO offload. Example of code to test availability from `datapath_winuser.c` in  msquic 

```
#ifdef UDP_SEND_MSG_SIZE
{
    DWORD SegmentSize;
    OptionLength = sizeof(SegmentSize);
    Result =
        getsockopt(
            UdpSocket,
            IPPROTO_UDP,
            UDP_SEND_MSG_SIZE,
            (char*)&SegmentSize,
            &OptionLength);
    if (Result != NO_ERROR) {
        int WsaError = WSAGetLastError();
        QuicTraceLogWarning(
            DatapathQueryUdpSendMsgFailed,
            "[ udp] Query for UDP_SEND_MSG_SIZE failed, 0x%x",
            WsaError);
    } else {
        Datapath->Features |= QUIC_DATAPATH_FEATURE_SEND_SEGMENTATION;
    }
}
#endif

#ifdef UDP_RECV_MAX_COALESCED_SIZE
{
    DWORD UroMaxCoalescedMsgSize = TRUE;
    OptionLength = sizeof(UroMaxCoalescedMsgSize);
    Result =
        getsockopt(
            UdpSocket,
            IPPROTO_UDP,
            UDP_RECV_MAX_COALESCED_SIZE,
            (char*)&UroMaxCoalescedMsgSize,
            &OptionLength);
    if (Result != NO_ERROR) {
        int WsaError = WSAGetLastError();
        QuicTraceLogWarning(
            DatapathQueryRecvMaxCoalescedSizeFailed,
[...]

PR #637, opened 2019-08-24
Add bytestream

steschu77:
This is a set of helper functions that deal with reading and writing QUIC formatted values to a memory buffer. It currently supports reading, writing, and skipping of:

* uint8, uint16, uint32, uint64 in network byte order
* variable length integer defined by QUIC
* buffers

I have pending branches that use these helpers for reading and writing the binary log file and for refactorings of reading/writing frames.


steschu77:
I'll create follow up PRs that clean up similar functionality like in frames.c for example.

picoquic/bytestream.c
```diff
@@ -0,0 +1,306 @@
+/*
+* Author: Christian Huitema
+* Copyright (c) 2018, Private Octopus, Inc.
+* All rights reserved.
```
huitema: 2019...

picoquic/bytestream.h
```diff
@@ -0,0 +1,77 @@
+/*
+* Author: Christian Huitema
+* Copyright (c) 2018, Private Octopus, Inc.
+* All rights reserved.
```
huitema: 2019

picoquic/bytestream.c
```diff
@@ -0,0 +1,306 @@
...
+
+bytestream * bytewriter_init(bytestream_buf * s)
+{
+    s->s.data = s->buf;
+    s->s.size = BYTESTREAM_MAX_BUFFER_SIZE;
+    s->s.ptr = 0;
+
+    return &s->s;
+}
+
```
huitema: Why can we specify a size in the "reader_init" command but not in the "writer_init" equivalent?

picoquic/bytestream.h
```diff
@@ -0,0 +1,77 @@
...
+* (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+* SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+#ifndef PICOQUIC_BYTESTREAM_H
+#define PICOQUIC_BYTESTREAM_H
+
+#include "picoquic_internal.h"
+
+#define BYTESTREAM_MAX_BUFFER_SIZE 2048
```
huitema: It would be nice to add comments about the principles of operation. I understand that you want to do buffered IO without using FILE from stdio, but when I read the code that's not clear. Do you intend to have one buffer per frame, or per file? In the read case, how do you manage overlapping buffers?

picoquic/bytestream.c
```diff
@@ -0,0 +1,306 @@
...
+void bytestream_delete(bytestream * s)
+{
+    if (s->data != NULL) {
+        free(s->data);
+        s->data = NULL;
+    }
+}
+
```
huitema: If we are going to have separate byte streams for each packet in the file, you may consider some kind of "recycle" structure, similar to what we do for packet buffers. That will cut down on the number of "malloc".

picoquictest/bytestream_test.c
```diff
@@ -0,0 +1,607 @@
+/*
+* Author: Christian Huitema
+* Copyright (c) 2017, Private Octopus, Inc.
+* All rights reserved.
```
huitema: 2019

steschu77:
> That's a bit hard to review without understanding how this will be used. I suspect that each byte stream will be written or read as a record in the binary log file. Is that the intent?

Here is a bit more context. The intend of the bytestream reading and writing was to have some tools for serializing and deserializing values to and from memory buffers, which was required for the binary log file.

In the meantime I found similar functionality throughout picoquic and replacing that with the bytestream as well. It makes the code much more readable. Here is an example for `picoquic_parse_ack_header()`:

Instead of:

```
    size_t byte_index = 1;
    size_t l_largest = 0;
    size_t l_delay = 0;

    if (bytes_max > byte_index) {
        l_largest = picoquic_varint_decode(bytes + byte_index, bytes_max - byte_index, largest);
        byte_index += l_largest;
    }
    if (bytes_max > byte_index) {
        l_delay = picoquic_varint_decode(bytes + byte_index, bytes_max - byte_index, ack_delay);
        byte_index += l_delay;
    }
```
You can now write:
```
    bytestream stream;
    bytestream * s = bytereader_init(&stream, bytes, bytes_max);
 
    int ret = 0;
    ret |= bytestream_skip(s, 1);
    ret |= byteread_vint(s, largest);
    ret |= byteread_vint(s, ack_delay);
```
If you'd also change the function signature and pass in the bytestream directly instead of `bytes` and `bytes_max` it gets even simpler.

picoquic/bytestream.c
steschu77: Good point.

The naming is not the best here. It depends more on how you want to use the bytestream. If you already have memory allocated, use the reader init. If not, like in a usual write use case, use the writer init, which provides temporary memory of 2048 bytes, which should be enough to write whole packets. If you need bigger buffers, use the bytestream_alloc initialization, which allocates the memory on the heap.

I'll rework the naming.

picoquic/bytestream.c
steschu77: I introduced the version with the 2048 byte buffer for that, so that no malloc is necessary at all. The malloc initialized bytestreams are currently only used in the conversion tool for reading/decoding the byte streams again.

picoquic/bytestream.h
steschu77: Have a look in picoquictest/bytestream_test.c on how to use the bytestream.

The connection to the file IO is really only that the resulting memory buffers can be read or written to an from files.

I'll open another PR on how to use the bytestream for binary file logging so you get an idea how that is intended to work.

picoquic/bytestream.c
steschu77: I've also added a size parameter to the "writer_init". That totally makes sense! Thanks for pointing that out 👍

huitema:
I am looking at your comment about rewriting the parsing of packets. I like the idea of using the bytestream construct for the log files, but there is a lot of history behind the current encoding/decoding of packets. @igorlord did a fair bit of work there, using a construct:
```
    next_pointer = decode_something(current_pointer, max_pointer, &variable)
```
That replaced a previous structure:
```
   byte_index += decode_something(bytes + byte_index, byte_max - byte_index, &variable)
```
We did not do it all the way, but I could see applying the construct proposed by @igorlord for encoding functions as well as decoding functions. I would much rather do that than introduce an intermediate "bytestream" structure.

Of course, we could do a test and look at performance differences. But I have no indication from previous performance tests that packet encoding is a performance bottleneck. The main culprit right now is the socket code (issue #564), but we only get to that bottleneck if we suppress logging. The next issue is congestion control, and specifically the "slow start" behavior which takes a long time reaching nominal throughput on high bandwidth*delay links. To investigate that, we need to turn on logs, but that will slow the code -- unless we can get high performance logs.

So I think high performance log writing should be the focus for now.

steschu77:
I wouldn't propose using the bytestream for performance reasons. I think the current solution is already optimal in that regard and the bytestream would probably be close or the same. I am proposing using the bytestream for code readability. It hides the little details of checking and maintaining the pointers from the actual parsing and encoding code. I finished replacing the decoding functions in frames.c today and could do that in 130 lines of code less.

Being new to the code it's sometimes hard to understand what the code does or to find the places where to hook in new code.

steschu77:
Besides the decode_frame refactoring discussion, is this change (add_bytestream) ok to merge?

PR #639, opened 2019-08-25
Refactor decode frames

steschu77:
This PR is FYI only. Please don't merge.

This PR shows how frames.c could look like when using bytestream.c functionality for parsing/decoding frames.


huitema:
I understand the code readability issues, but I would refrain from making these many changes to the code in "frames.c" right now. I think that we could achieve the same readability without the bytestream concept, on a piecemeal basis.

PR #642, opened 2019-09-03
Post call back

huitema:
Add feature to picohttp server, so a URL can be associated with a callback. Post to the URL are passed as data to the callback. Output produced by the callback is sent as response to POST.

This PR does not address client side processing.

The work is not fully complete. Another PR should add processing of GET arguments (as in Get /some-url?specific&arguments"), and also add a way to specify the content type returned by the POST command.


steschu77:
Looks good to me.

PR #645, opened 2019-09-09
Fix "unused variable" warnings after file close.

huitema:


steschu77:
I was afraid a compiler or sanitizer will complain about this sooner or later :)

Issue #653, opened 2019-09-13
Draft23: suppress SETTINGS_NUM_PLACEHOLDERS


huitema:
The definition of SETTINGS_NUM_PLACEHOLDERS (0x9) is removed in draft-23.
The PRIORITY frame (code 2) is also removed.

Issue #652, opened 2019-09-13
Draft 23: HTTP error code renumbered


huitema:
For example, HTTP_NO_ERROR (0x00) becomes HTTP_NO_ERROR (0x100)

Issue #651, opened 2019-09-13
Satellite: encode latency and bandwidth in ticket or token, remember at client side


huitema:
The HTTP spec says:

_When a 0-RTT QUIC connection is being used, the initial value of each server setting is the value used in the previous session. Clients SHOULD store the settings the server provided in the connection where resumption information was provided, but MAY opt not to store settings in certain cases (e.g., if the session ticket is received before the SETTINGS frame). A client MUST comply with stored settings - or default values, if no values are stored - when attempting 0-RTT. Once a server has provided new settings, clients MUST comply with those values._

On the client side, this implies remembering an application setting frame and storing it together with the session token. On the server side, this implies either always reusing the same parameters, or alternately encoding the value of the parameters in the session ticket. Or, possibly, using some other alternative, such as setting identifiers.

In any case, this requires providing an API for optionally setting initial application parameters in tickets and retrieving them on the server side, and possibly a similar API on the client side for storing information with tickets and retrieving it.

huitema:
Confirmed that resumption tickets are for SNI+ALPN.

jlaine:
What kind of API are you imagining on the client side? Being able to pass in a single arbitrary binary blob of settings, and getting it handed back on session resumption?

huitema:
@jlaine revisiting this issue. The initial motivation was the satellite work, in which there is a lot to be gained if the resumed connection can start with the right congestion control parameters. This would require knowledge of:

* observed bandwidth in previous connection
* observed RTT in previous connection
* date of last connection, because the older the measurement the less trust in them
* IP address of last connection, because measurements are only valid for some given network path

The client could place these data in a transport parameter, but that opens a can of worms. What if the client is hostile? What if it is trying to DOS a remote network? It seems safer to have the server place that data in an encrypted blob, which could be either the "session resume ticket" or the "retry token". Of these two, the retry token seems like a better fit, because its content is entirely managed by the Quic stack, rather than exported through TLS. Also, the format of the token is entirely defined by the implementation, and the implementation can send it whenever it wants.

So the design would be:

1) Add some fields in the encrypted token
2) Delay sending the NEW TOKEN frame until RTT and bandwidth have been measured
3) Send the NEW TOKEN then, even if the server is not using Retry for protection
4) On incoming connection, pass the estimated RTT and Bandwidth to the CC algorithms

huitema:
Fixed.

huitema:
[...]

Issue #650, opened 2019-09-13
Draft 23: ChaCha20-Based Header Protection


huitema:
Check text in new TLS draft.

PR #654, opened 2019-09-13
Support for Draft23 and remove of support for Draft 20

huitema:
Draft23 are just published. The changes are not big, as described in issues #647, #648, #649, #650, #652 and #653. All these issues are addressed in the PR. Issue #651 is not addressed, because it requires some clarification from the WG.

The PR also removes the hacks introduced to support draft-20 simultaneously with draft-22, and thus the support for draft 30.


picoquic/picoquic.vcxproj
```diff
@@ -166,8 +166,6 @@
     <ClInclude Include="bytestream.h" />
     <ClInclude Include="democlient.h" />
     <ClInclude Include="demoserver.h" />
-    <ClInclude Include="fnv1a.h" />
```
steschu77: Was this intentional?

I like having the headers inside the solution as well to support "search in files" for these headers.

picoquic/logger.c
```diff
@@ -784,13 +772,7 @@ size_t picoquic_log_generic_close_frame(picoquic_cnx_t * cnx, FILE* F, uint8_t*
     size_t l0 = 0;
 
     if (bytes_max >= 3) {
```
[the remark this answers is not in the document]
huitema: Should be 2 in fact...

picoquic/logger.c
```diff
@@ -725,13 +719,7 @@ size_t picoquic_log_reset_stream_frame(picoquic_cnx_t * cnx, FILE* F, uint8_t* b
...
         byte_index += l1;
         if (l1 > 0 && bytes_max >= byte_index + 2) {
-            if (cnx != NULL && picoquic_supported_versions[cnx->version_index].version == PICOQUIC_TWELFTH_INTEROP_VERSION) {
-                error_code = PICOPARSE_16(bytes + byte_index);
-                l2 = 2;
-            }
-            else {
-                l2 = picoquic_varint_decode(bytes + byte_index, bytes_max - byte_index, &error_code);
-            }
+            l2 = picoquic_varint_decode(bytes + byte_index, bytes_max - byte_index, &error_code);
```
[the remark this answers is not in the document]
huitema: OK. But I actually added a better check on l1 and l2.

picoquic/picoquic.vcxproj
huitema: fnv1a was removed from the solution. Democlient and Demoserver were moved to a different project. But removing bytestream is a mistake, thanks for noticing.

huitema: Actually, I should also remove democlient and demoserver, since they are not part of the main picoquic project.

huitema: Will leave fnv1a here for now, deal with removal later.

PR #656, opened 2019-09-13
Conid is zero

steschu77:
Changed `picoquic_is_connection_id_null` to take a const pointer instead of the whole structure.

### DIFF
--- a/picoquic/util.h
+++ b/picoquic/util.h
@@ -57,7 +57,7 @@ extern const picoquic_connection_id_t picoquic_null_connection_id;
 uint8_t picoquic_format_connection_id(uint8_t* bytes, size_t bytes_max, picoquic_connection_id_t cnx_id);
 int picoquic_is_connection_id_length_valid(uint8_t len);
 uint8_t picoquic_parse_connection_id(const uint8_t* bytes, uint8_t len, picoquic_connection_id_t *cnx_id);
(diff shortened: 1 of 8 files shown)


picoquic/util.h
```diff
@@ -57,7 +57,7 @@ extern const picoquic_connection_id_t picoquic_null_connection_id;
 uint8_t picoquic_format_connection_id(uint8_t* bytes, size_t bytes_max, picoquic_connection_id_t cnx_id);
 int picoquic_is_connection_id_length_valid(uint8_t len);
 uint8_t picoquic_parse_connection_id(const uint8_t* bytes, uint8_t len, picoquic_connection_id_t *cnx_id);
-int picoquic_is_connection_id_null(picoquic_connection_id_t cnx_id);
+int picoquic_is_connection_id_null(const picoquic_connection_id_t * cnx_id);
 int picoquic_compare_connection_id(const picoquic_connection_id_t * cnx_id1, const picoquic_connection_id_t * cnx_id2);
 uint64_t picoquic_val64_connection_id(picoquic_connection_id_t cnx_id);
 void picoquic_set64_connection_id(picoquic_connection_id_t * cnx_id, uint64_t val64);
```
huitema: Maybe we should apply the pointer treatment to all the connection ID references, including val64. Could do that later of course.

steschu77: Good idea. I am a big fan of constant refactoring ;)

PR #657, opened 2019-09-13
Add logging library

steschu77:
This is the first step of introducing binary packet log conversion in a separate logging library.

This PR extracts the CSV conversion from the picoquic library into the separate logging library.


steschu77:
> Can you check that?

Fixed.

I was treating NULL pointer strings as empty strings, so the bytewrite_cstr method checked the string for NULL pointer before retrieving the string length. The analyzer was complaining then as the NULL pointer string was handed in to memcpy() with a zero length to copy.

I removed the support for NULL pointer strings for now.

PR #658, opened 2019-09-13
Add picolog binary log conversion tool

steschu77:
picolog is considered to replace picoquic_c2csv in the future but provides higher flexibility in terms of output formats and conversion options.

This first version supports conversion into csv.

### DIFF
--- a/picolog/picolog.c
+++ b/picolog/picolog.c
@@ -1,0 +1,113 @@
+/*
+* Author: Christian Huitema
+* Copyright (c) 2019, Private Octopus, Inc.
+* All rights reserved.
+*
+* Permission to use, copy, modify, and distribute this software for any
+* purpose with or without fee is hereby granted, provided that the above
+* copyright notice and this permission notice appear in all copies.
+*
+* THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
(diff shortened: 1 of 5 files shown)


picolog/picolog.c
```diff
@@ -0,0 +1,113 @@
...
+
+    int opt;
+    while ((opt = getopt(argc, argv, "o:f")) != -1) {
+        switch (opt) {
+        case 'o':
+            out_file = optarg;
+            break;
+        case 'f':
+            out_format = optarg;
+            break;
```
huitema: You may consider parsing the output format here, so you can reject incorrect output format early, and also provide "usage" feedback with the list of supported formats.

picolog/picolog.c
```diff
@@ -0,0 +1,113 @@
...
+    if (log_name != NULL && log == NULL) {
+        fprintf(stderr, "Could not open file %s\n", log_name);
+        exit(1);
+    }
+
+    if (strcmp(out_format, "csv") == 0) {
+        ret = picoquic_cc_bin_to_csv(log, open_outfile(log_name, out_file, "csv"));
+    } else {
+        fprintf(stderr, "Invalid output format %s\n", out_format);
+        ret = 1;
```
huitema: Or alternately, provide additional info on supported formats at this stage.

PR #661, opened 2019-09-14
Fix hash table count on item delete

steschu77:
* Add tests for hash table count
* Add messages in hash table tests
* Decrement count on item delete

### DIFF
--- a/picoquictest/hashtest.c
+++ b/picoquictest/hashtest.c
@@ -23,6 +23,8 @@
 #ifdef _WINDOWS
 #include <malloc.h>
 #endif
+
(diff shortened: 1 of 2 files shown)


picoquictest/hashtest.c
```diff
@@ -120,11 +152,17 @@ int picohash_test()
 
         /* Check that the deleted are gone */
 
+        if (t->count != 8) {
+            DBG_PRINTF("picohash table count != 8 (count=%"PRIst")\n", t->count);
+            ret = -1;
+        }
+
```
huitema: Yes!

PR #662, opened 2019-09-15
Add binary packet logging and conversion to SVG flow graph

steschu77:
This PR adds binary packet logging into a new binary log file. The binary log file is currently written in parallel to the existing text based log file. That can be cleaned up later once binary log file conversion to a text based format is available.

The picolog tool is able to extract all connection ids from a given binary log file and displays them. A new command line parameter has been added to select the desired connection. If no connection is selected, picolog converts log data of all connections in a binary logfile to the given output format.

There are currently no tests yet. Instead, a conversion to an SVG flow graph has been added for basic verification. Tests for binary logging and conversion will follow shortly.


steschu77:
> I understand that you are building this program one part at a time. Can you add a few comments before each function to explain what it does? Pretty hard to review otherwise...

I've added a few comments to header files. Let me know if that helps or if you need more documentation.

I also tried to make the code more readable where possible.

loglib/logreader.c
```diff
@@ -0,0 +1,226 @@
...
+    convert_log_file_event_t* ctx = (convert_log_file_event_t*)ptr;
+    void * cbptr = ctx->callbacks->ptr;
+
+    picoquic_connection_id_t cid;
+    ret |= byteread_cid(s, &cid);
+
+    /* filter for connection id */
+    if (ret != 0 || picoquic_compare_connection_id(&cid, ctx->cid) != 0) {
+        return ret;
+    }
```
huitema: Is the CID test specific to this particular callback?

steschu77: Parsing the log file for one specific connection is the main use case for the log file event enumeration. Therefore it's done early so you can skip all the unnecessary parsing in case the log event doesn't belong to the connection you are looking for.

PR #663, opened 2019-09-17
picohttp_demo_post_sha256_callback is never used

steschu77:
`picohttp_demo_post_sha256_callback()` currently issues a "parameter not used" warning in VS. Looking into it, it turns out the whole function is never used except by itself. Therefore I removed it.

Feel free to reject this PR is you want to keep `picohttp_demo_post_sha256_callback()` for other reasons.


huitema:
I agree that this code does not belong. It was thought of as an example of callback function for the POST API, but a simpler example is used for tests.

PR #669, opened 2019-10-01
Qlog script readiness

steschu77:
Final tweaks for QLOG conversion

* Flush binary log when connections close
* Enable output on `stdout` if no output directory is specified on picolog
* Fix executable name for picolog in Visual Studio
* Remove template parameter requirement for QLOG conversion


huitema:
I am merging this because we need it for the interop, but there are two issues: some compile warnings due to variable set but not used (scan-build) and I don't understand how we close the qlog file. Is that autoclose at the end of the process?

steschu77:
It's not autoclose yet. But I can add that tomorrow.

PR #672, opened 2019-10-02
Move reference time to connection start time

steschu77:


picolog/picolog.c
```diff
@@ -635,7 +636,7 @@ int convert_qlog(const picoquic_connection_id_t* cid, void* ptr)
             qlog.f_txtlog = f_txtlog;
             qlog.f_template = appctx->f_template;
             qlog.cid_name = cid_name;
-            qlog.start_time = appctx->log_time;
+            qlog.start_time = 0;
```
huitema: I am not sure about that. I think we should distinguish two times: the start of log time (qlog start time) which should hold the UTC time at the beginning of the log; and, the event time, which should be relative to the start of the log. We should be able to merge qlogs from different sources, e.g. server, client and network, and be able to align the time of the events. 

Thinking "end to end", the start of log should be set to `cnx->start_time`, i.e. the time at which the logged connection was initiated. The event time should be set to `current_time - cnx->start_time`. I am not sure how to do that when writing the bin log.

huitema: OK, I understand that the change does in fact do what I expect.

PR #673, opened 2019-10-02
Fix static code analysis

steschu77:
Fix false positive static code analysis warnings

### DIFF
--- a/picolog/picolog.c
+++ b/picolog/picolog.c
@@ -588,6 +588,7 @@ int qlog_connection_start(uint64_t time, const picoquic_connection_id_t * cid, i
     svg_context_t * ctx = (svg_context_t*)ptr;
     FILE * f = ctx->f_txtlog;
 
+    ctx->start_time = time;
     ctx->packet_count = 0;
 
     fprintf(f, "{ \"qlog_version\": \"draft-00\", \"title\": \"picoquic\", \"traces\": [\n");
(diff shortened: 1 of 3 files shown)


picoquictest/stresstest.c
```diff
@@ -966,7 +966,7 @@ static int stress_or_fuzz_test(picoquic_fuzz_fn fuzz_fn, void * fuzz_ctx, uint64
     /* Initialization */
     memset(&stress_ctx, 0, sizeof(picoquic_stress_ctx_t));
     stress_set_ip_address_from_index(&stress_ctx.server_addr, -1);
-    stress_ctx.nb_clients = (int)picoquic_stress_nb_clients;
+    stress_ctx.nb_clients = nb_clients;
     if (stress_ctx.nb_clients > PICOQUIC_MAX_STRESS_CLIENTS) {
         DBG_PRINTF("Number of stress clients too high (%d). Should be lower than %d\n",
```
huitema: why not use nbclients here too, just like in the other lines of code further down?

picolog/picolog.c
```diff
@@ -635,7 +636,7 @@ int convert_qlog(const picoquic_connection_id_t* cid, void* ptr)
             qlog.f_txtlog = f_txtlog;
             qlog.f_template = appctx->f_template;
             qlog.cid_name = cid_name;
-            qlog.start_time = appctx->log_time;
+            qlog.start_time = 0;
```
huitema: Please see comments on PR #672

picoquictest/stresstest.c
huitema: But then, it does not change the actual results...

PR #682, opened 2019-10-16
Fix size_t vs uint64_t formating

steschu77:


picolog/picolog.c
```diff
@@ -393,13 +393,13 @@ int svg_packet_start(uint64_t time, uint64_t size, const picoquic_packet_header
 
     if (rxtx == 0) {
         fprintf(f, "  <text x=\"%d\" y=\"%d\" text-anchor=\"end\" class=\"seq_%s\">%"PRIu64"</text>\n", x_pos - 4, y_pos - 4, dir, ph->pn64);
-        fprintf(f, "  <text x=\"%d\" y=\"%d\" text-anchor=\"end\" class=\"arw\">%zu b</text>\n", 80, y_pos - 2, size);
+        fprintf(f, "  <text x=\"%d\" y=\"%d\" text-anchor=\"end\" class=\"arw\">%"PRIu64" b</text>\n", 80, y_pos - 2, size);
```
huitema: That won't work on 32 bit systems in which size = uint32_t. You need:

```
fprintf(f, "  <text x=\"%d\" y=\"%d\" text-anchor=\"end\" class=\"arw\">%"PRIu64" b</text>\n", 80, y_pos - 2, (uint64_t)size);
```

steschu77:
The type of size is uint64_t. That is because it is read from the log file
as variable integer and a conversion to size_t would be a downcast on 32
bit systems.

Christian Huitema <notifications@github.com> schrieb am Mi., 16. Okt. 2019
16:47:

picolog/picolog.c
huitema: OK, forget that. I understand that it is not an actual problem.

PR #692, opened 2019-10-21
Fix refactoring warnings

steschu77:
Enable Warnings as Errors and lift the warning level of some Visual Studio projects to level 4.

Plus the required fixes for a green build :)


loglib/logreader.c
```diff
@@ -129,8 +129,15 @@ static int binlog_convert_event(bytestream * s, void * ptr)
...
-            ret |= byteread_vint(s, &len);
+
+            uint64_t len_read = 0;
+            ret |= byteread_vint(s, &len_read);
+
+            size_t len = (size_t)len_read;
+            if (len != len_read) {
+                ret = -1;
+                break;
+            }
```
huitema: That works. "byteread_vsize(s, &len)" would also have worked.

huitema: Actually, why not use the function `byteread_vlen(s, &len)` since you created it?

steschu77: Good catch :) I've changed it.

PR #693, opened 2019-10-22
Remove pre draft-21 connection id parsing in packet header

steschu77:
I've deleted the old invariant connection id parsing in the packet header.

That required a functional change in the way the version for packet header parsing is evaluated. I tried to stay close to the current QUIC specification.

I had to change test_tls_api_version_negotiation, which was using the old invariant. Now it also uses the new invariant. A version of ?A?A?A?A was causing picoquic to send packets with the old invariant. I deleted that code as well.

### DIFF
--- a/picoquic/packet.c
+++ b/picoquic/packet.c
@@ -34,110 +34,30 @@
 #include <stdlib.h>
 #include <string.h>
 
-/**
- * If the version is not know, need special code as long as we support
- * draft-20 and the corresponding format */
-
-int picoquic_is_old_header_invariant(
-    picoquic_quic_t* quic,
-    uint8_t* bytes,
-    size_t length,
-    picoquic_packet_header* ph)
+int picoquic_is_header_version_supported(uint32_t version)
 {
-    int ret = 0; /* return 1 if old invariant */
-    if ((ph->vn & 0xFFFFFF00) == 0xFF000000) {
-        /* Draft versions before #20 use the old invariants */
-        int draft_nb = ph->vn & 0xFF;
-        ret = (draft_nb <= 20) ? 1 : 0;
-    }
-    else if (ph->vn != 0 && (
-        (ph->vn & 0xFFFFFFF0) == 0 ||
-        ph->vn == PICOQUIC_INTERNAL_TEST_VERSION_1 ||
-        ph->vn == PICOQUIC_INTERNAL_TEST_VERSION_2)) {
-        /* Final versions and internal versions use the new invariants */
-        ret = 0;
-    } else {
-        /* Cases where the version is either NULL, or one of the greasing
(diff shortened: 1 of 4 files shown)


picoquic/packet.c
```diff
@@ -338,77 +251,122 @@ int picoquic_parse_packet_header(
...
-             } else {
-                 ph->payload_length = (uint16_t)(length - ph->offset);
-             }
-         } else {
-             /* This may be a packet to a forgotten connection */
-             ph->ptype = picoquic_packet_1rtt_protected;
-             ph->payload_length = (uint16_t)((length > ph->offset)?length - ph->offset:0);
-         }
+        ret = picoquic_parse_short_packet_header(quic, bytes, length, addr_from, ph, pcnx, receiving);
     }
```
huitema: This split makes a lot of sense for readability, but it complicates reviewing the changes...

picoquic/packet.c
```diff
@@ -34,110 +34,30 @@
...
+    } else if ((version & 0x0F0F0F0F) == 0x0A0A0A0A) {
+        /* version greasing, i.e. force version negotiation */
+        ret = 1;
+    } else if (
+        version == PICOQUIC_INTERNAL_TEST_VERSION_1 ||
+        version == PICOQUIC_INTERNAL_TEST_VERSION_2) {
+        /* Internal versions use the new invariants */
+        ret = 1;
     }
-    else if (ph->vn != 0 && (
```
huitema: I would personally just remove this entirely. The protocol defines the version independent invariants that constrain all future versions. The change in May 2019 was pretty much the last chance of fixing something. After that, once some deployment has started, that becomes impossible. Facebook, Cloudflare and Google are already shipping products with the new format, so change is pretty much impossible.

picoquic/packet.c
```diff
@@ -148,53 +68,45 @@ int picoquic_parse_packet_header(
...
-    if ((bytes[0] & 0x80) == 0x80) {
-        if (length < 6) {
+    if (length < 6) {
+        ret = -1;
+    }
+    else {
+        /* The bytes at position 1..4 describe the version */
+        ph->vn = PICOPARSE_32(bytes + 1);
+        if (!picoquic_is_header_version_supported(ph->vn)) {
             ret = -1;
```
huitema: As I say above, we can just skip this and assume that we only have supported version. Everything else would be a parsing error.

picoquic/packet.c
steschu77: 👍

picoquic/packet.c
steschu77: Done.

picoquic/packet.c
steschu77: I saw that once I opened the PR. I'll try to separate refactorings from functional changes better in the future.

PR #695, opened 2019-10-23
More header parsing refactoring

steschu77:
- Second try of using bytestream for parsing :)
- Remove picoquic_version_header_17 evaluation as it is always picoquic_version_header_17

### DIFF
--- a/picoquic/packet.c
+++ b/picoquic/packet.c
@@ -87,61 +87,50 @@ int picoquic_parse_long_packet_header(
             if (ph->version_index >= 0) {
                 /* If the version is supported now, the format field in the version table
                 * describes the encoding. */
-                switch (picoquic_supported_versions[ph->version_index].version_header_encoding) {
-                case picoquic_version_header_17:
-                    ph->spin = 0;
-                    ph->has_spin_bit = 0;
-                    switch ((flags >> 4) & 7) {
-                    case 4: /* Initial */
-                    {
-                        /* special case of the initial packets. They contain a retry token between the header
-                        * and the encrypted payload */
-                        uint64_t tok_len = 0;
-                        size_t l_tok_len = picoquic_varint_decode(bytes + ph->offset, length - ph->offset, &tok_len);
+                ph->spin = 0;
+                ph->has_spin_bit = 0;
+                switch ((flags >> 4) & 7) {
+                case 4: /* Initial */
+                {
+                    /* special case of the initial packets. They contain a retry token between the header
+                    * and the encrypted payload */
+                    uint64_t tok_len = 0;
+                    size_t l_tok_len = picoquic_varint_decode(bytes + ph->offset, length - ph->offset, &tok_len);
 
-                        ph->ptype = picoquic_packet_initial;
-                        ph->pc = picoquic_packet_context_initial;
-                        ph->epoch = 0;
-                        if (l_tok_len == 0) {
-                            /* packet is malformed */
-                            ph->offset = length;
-                            ph->ptype = picoquic_packet_error;
-                            ph->pc = 0;
-                        }
-                        else {
-                            ph->token_length = (size_t)tok_len;
-                            ph->token_bytes = bytes + ph->offset + l_tok_len;
-                            ph->offset += l_tok_len + (size_t)tok_len;
-                        }
-
-                        break;
-                    }
-                    case 5: /* 0-RTT Protected */
-                        ph->ptype = picoquic_packet_0rtt_protected;
-                        ph->pc = picoquic_packet_context_application;
-                        ph->epoch = 1;
-                        break;
-                    case 6: /* Handshake */
-                        ph->ptype = picoquic_packet_handshake;
-                        ph->pc = picoquic_packet_context_handshake;
-                        ph->epoch = 2;
-                        break;
-                    case 7: /* Retry */
-                        ph->ptype = picoquic_packet_retry;
-                        ph->pc = picoquic_packet_context_initial;
-                        ph->epoch = 0;
-                        break;
-                    default: /* Not a valid packet type */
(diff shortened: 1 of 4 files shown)


picoquic/packet.c
```diff
@@ -40,199 +41,167 @@ int picoquic_parse_long_packet_header(
...
-        /* The bytes at position 1..4 describe the version */
-        ph->vn = PICOPARSE_32(bytes + 1);
-        uint8_t l_dest_id = 0;
-        uint8_t l_srce_id = 0;
-        uint32_t i_srce_id = 0;
+    uint8_t flags = 0;
+    ret |= byteread_int8(s, &flags);
+    ret |= byteread_int32(s, &ph->vn);
+    ret |= byteread_cid(s, &ph->dest_cnx_id);
+    ret |= byteread_cid(s, &ph->srce_cnx_id);
```
huitema: Compare that to using the constructs used in frame.c, such as picoquic_frames_uint8_decode, etc. You would write:

```
uint8_t * bytes_max = bytes + length;

if ((bytes = picoquic_frames_uint8_decode(bytes, bytes_max, &flags)) == NULL ||
    (bytes = picoquic_frames_uint32_decode(bytes, bytes_max, &ph->vn)) == NULL  ||
    (bytes = picoquic_frames_cid_decode(bytes, bytes_max, &ph->dest_cnx_id)) == NULL ||
    (bytes = picoquic_frames_cid_decode(bytes, bytes_max, &ph->srce_cnx_id)) == NULL){
          ret = some_error_code;
}
```

Avoiding the use of a specific stream construct, and also stopping parsing as soon as an error is detected.

On the other hand, I don't want to discourage progress, I know that the differences are minimal, and I also know that we are not anywhere near the critical path.

I guess my real beef is with the `ret |= ` construct, which is not right because ret is not a boolean, and also does not stop parsing as soon as an error is detected.

```
if ((ret = byteread_int8(s, &flags)) == 0 &&
    (ret |= byteread_int32(s, &ph->vn)) == 0 &&
    (ret |= byteread_cid(s, &ph->dest_cnx_id)) == 0 &&
    (ret |= byteread_cid(s, &ph->srce_cnx_id)) {
          /* continue with processing */
```
would work.

picoquic/quicctx.c
```diff
@@ -144,19 +144,15 @@ static uint8_t picoquic_cleartext_draft_23_salt[] = {
...
         picoquic_cleartext_draft_22_salt },
     { PICOQUIC_INTERNAL_TEST_VERSION_2,
-        picoquic_version_header_17,
         sizeof(picoquic_cleartext_internal_test_1_salt),
         picoquic_cleartext_internal_test_1_salt },
     { PICOQUIC_INTERNAL_TEST_VERSION_1,
-        picoquic_version_header_17,
         sizeof(picoquic_cleartext_internal_test_1_salt),
         picoquic_cleartext_internal_test_1_salt }
 };
```
huitema: Yes, "version 17" is the last time the packet code changed, we use only one code now, and we can remove that as you suggest. On the other hand, the packet codes are specified to be version dependent. There are talks of exercising that, as part of protection against "version number ossification". But I guess that we can change the code again when we do have a spec that exercises the variations.

picoquic/packet.c
steschu77: Thanks, Christian, for your feedback. I totally agree with your comment about the `ret |= ...` construct and I'll change that.

The other thing is just a matter of taste and I prefer to implement your taste as you're the maintainer of picoquic :) So, I'll change that too.

picoquic/packet.c
```diff
@@ -34,205 +34,183 @@
...
 #include <string.h>
 
+uint8_t* picoquic_frames_varint_decode(uint8_t* bytes, const uint8_t* bytes_max, uint64_t* n64);
+uint8_t* picoquic_frames_varlen_decode(uint8_t* bytes, const uint8_t* bytes_max, size_t* n);
+uint8_t* picoquic_frames_uint8_decode(uint8_t* bytes, const uint8_t* bytes_max, uint8_t* n);
+uint8_t* picoquic_frames_uint16_decode(uint8_t* bytes, const uint8_t* bytes_max, uint16_t* n);
+uint8_t* picoquic_frames_uint32_decode(uint8_t* bytes, const uint8_t* bytes_max, uint32_t* n);
+uint8_t* picoquic_frames_uint64_decode(uint8_t* bytes, const uint8_t* bytes_max, uint64_t* n);
+uint8_t* picoquic_frames_cid_decode(uint8_t* bytes, const uint8_t* bytes_max, picoquic_connection_id_t* n);
+
```
huitema: Yes, that works. What also work would be to put these definitions in "util.h"...

steschu77: Let me do that in a follow up PR because I would then also move the functions to utils.c and rename them (without "frames"). Thats a little more work because there are already some functions with the same name I would need to deal with.

steschu77:
@huitema feel free to merge if you are ok with the changes.

picoquic/packet.c
huitema: Thanks for being flexible. Apart from taste, there is also the issue of rewriting frame.c or not. If we pick a standard way, we will have to apply it everywhere.

The _frames_ functions were suggested by @igorlord, as a way to shorten the size of the frame decoding functions. However, we did it halfway, and only considered the decoding functions. I suppose I have now to come up with the "picoquic_frames_*_encode()" functions.

picoquic/packet.c
huitema: OK.

picoquic/packet.c
steschu77: Fixing frames.c decoding is on my list already. I also will see if I can change the bytestream stuff to the picoquic_*_decode functions.

PR #711, opened 2019-11-16
Quic in quic

huitema:
This is the first check-in of the "quic in quic" proxy library. No demo client or server yet, just a set of unit tests that manage to validate the very first design of the protocol. The unit test checks that the client can do a connection to the remote server through the proxy.

There is plenty of work remaining to do in that project, but we are checking that in so as to avoid having parallel branches for a long time.


picoqinq/qinqclient.c
```diff
@@ -0,0 +1,451 @@
...
+            stream_ctx->hc = NULL;
+        }
+        free(stream_ctx);
+    }
+}
+
+/* Management of header compression
+ */
+
+int picoqinq_client_is_register_address_pending(picoqinq_client_callback_ctx_t* ctx, struct sockaddr* addr, picoquic_connection_id_t* cid, uint64_t direction)
```
steschu77: All passed pointers should be const here.

picoqinq/qinqclient.c
```diff
@@ -0,0 +1,451 @@
...
+    while (stream_ctx != NULL && stream_ctx->stream_id != stream_id) {
+        stream_ctx = stream_ctx->next_stream;
+    }
+
+    if (stream_ctx == NULL && should_create) {
+        stream_ctx = (picoqinq_client_stream_ctx_t*)
+            malloc(sizeof(picoqinq_client_stream_ctx_t));
+        if (stream_ctx == NULL) {
+            /* Could not handle this stream */
+            picoquic_reset_stream(cnx, stream_id, PICOQINQ_ERROR_INTERNAL);
```
steschu77: That should be worth a trace :)

steschu77: picoqinq doesn't seem to trace at all. There are more trace-worthy events in this file, but I skip over that for now.

picoquic/util.c
```diff
@@ -366,6 +366,26 @@ void picoquic_set64_connection_id(picoquic_connection_id_t * cnx_id, uint64_t va
...
+
+    if (addr->sa_family == AF_INET) {
+        struct sockaddr_in* a4 = (struct sockaddr_in*)addr;
+        h = picohash_bytes((uint8_t*)&a4->sin_addr , 4);
+        h += 128ull * a4->sin_port;
+        h ^= 0x04040404;
+    }
+    else {
+        struct sockaddr_in6* a6 = (struct sockaddr_in6*)addr;
+        h = picohash_bytes((uint8_t*)& a6->sin6_addr, 4);
```
steschu77: Copy & paste? 4 should be 16 or some "sizeof"

picoquic/util.c
```diff
@@ -366,6 +366,26 @@ void picoquic_set64_connection_id(picoquic_connection_id_t * cnx_id, uint64_t va
...
 
+uint64_t picoquic_hash_addr(const struct sockaddr* addr)
+{
+    uint64_t h;
+
+    if (addr->sa_family == AF_INET) {
+        struct sockaddr_in* a4 = (struct sockaddr_in*)addr;
+        h = picohash_bytes((uint8_t*)&a4->sin_addr , 4);
+        h += 128ull * a4->sin_port;
+        h ^= 0x04040404;
```
steschu77: Why is that and the 0x06060606 below necessary?

CMakeLists.txt
```diff
@@ -81,6 +81,16 @@ set(PICOHTTP_TEST_LIBRARY_FILES
     picoquictest/http0dot9test.c
 )
 
+set(PICOQINQ_LIBRARY_FILES
```
steschu77: Thanks for making this a separate library!

👍

picoquic/util.c
huitema: The idea was to add a different salt to the hash for IPv6 and IPv4 addresses. But yes, that probably not too useful.

picoquic/util.c
huitema: Good catch!

picoqinq/qinqclient.c
huitema: OK. I fixed that here and in a set of other places.

PR #732, opened 2019-11-29
Remove unnecessary timer on cwin blocked.

huitema:
@jlaine this removes the "cwin timer" code that you flagged as unnecessary. The "packet trace" test shows no difference, which means the timer was indeed not useful.


jlaine:
Thanks for looking into this, the next person who reads the code won't have to wonder what it does :)

Issue #720, opened 2019-11-21
Satellite transmission tests 


huitema:
The IETF discussion of satellite transmission outlined issues with handling large BDP links. There are issues like long delays forcing slow start to last a long time, large sensitivity to losses, and high cost of the uplink. The first step there is to add a series of non regression tests:

1) Transmission of a large file over a high speed satellite link. Assume direct communication, very few losses. 

2) Same test over concatenation of Wi-Fi link and satellite link. Assume some level of loss due to Wi-Fi link.

3) Assume that the link fades from 100 Mbps to 10 Mbps in the middle of the session and then goes back, possibly with a high loss period during the transition from 100 to 10.

The evaluation should include the target efficiency and the corresponding completion time, and the volume of data on the up-link.

huitema:
Added basic test in PR #733.

First observation: 
* Transfer of 10MB completes in 6 seconds
* Code remains in slow start for entire period
* Zero loss
* CWIN grows to 60MB, way more than required 7.5MB
* Pacing consistent with 750 Mbps
* Cwin blocked only during early stage of connection
* Flw blocked only during early stage
* Stream block happens way more often.

huitema:
Added satellite test in PR #733 
First observation:
* Limited by flow control (stream, cnx)
* Lots of batched sends, as if pacing did not work
* stabilizes to expected CWIN after 6 seconds

huitema:
Limiting the CWIN growth to cases where the sender was CWIN-blocked in the previous RTT does bring some sanity. Switching to more aggressive increase of flow control windows if RTT is larger than 100ms reduces the duration of the satellite test by 1.5 second, from 7 to 5.5 sec.

huitema:
Remaining issues:
* Hystart test for exit slow start is too lenient when delay is high, allowing the CWIN to keep growing when the SRTT increases from 600200 to 610000.
* Still some number of batched sends.
* Too many acks
* Initial rise is too slow.

huitema:
Almost there for the non zero part, but need to decimate ACKS.

Ack concatenation is tricky:
* done during slow start, it will slow down the window ramp up.
* doing it right requires knowing sender CWIN and sender RTT.
* do not delay if FIN received.
* ideally, install a feedback loop so the ACK rate matches the capacity of the uplink.

huitema:
First batch of changes done in PR #734 . The feedback loop for ACK and the handling of non-congestion losses will have to happen in separate PR.

[...]

PR #734, opened 2019-12-03
Satellite fix 1

huitema:
Addressing the issue #720 by adding a series of improvements on flow control, congestion control and packing.

As note in #720, this is still work in progress.


huitema:
Merging at this point. There is work left to do on ack coalescing and dealing with non-congestion losses, but this will come in separate PRs.

PR #737, opened 2019-12-09
Integrate WSAStartUp into picosocks

steschu77:
Add WSAStartUp to picosocks.c and remove indicidual initialization from picoquicdemo and tests.

That helps users of picoquic to avoid debugging Windows-related network issues.

### DIFF
--- a/picoquic/picosocks.c
+++ b/picoquic/picosocks.c
@@ -284,6 +284,8 @@ int picoquic_socket_set_ecn_options(SOCKET_TYPE sd, int af, int * recv_set, int 
 SOCKET_TYPE picoquic_open_client_socket(int af)
 {
 #ifdef _WINDOWS
+    WSADATA wsaData = { 0 };
+    WSA_START(MAKEWORD(2, 2), &wsaData);
     SOCKET_TYPE sd = WSASocket(af, SOCK_DGRAM, IPPROTO_UDP, NULL, 0, WSA_FLAG_OVERLAPPED);
 #else
     SOCKET_TYPE sd = socket(af, SOCK_DGRAM, IPPROTO_UDP);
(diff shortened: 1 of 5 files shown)


picoquic/picosocks.c
```diff
@@ -284,6 +284,8 @@ int picoquic_socket_set_ecn_options(SOCKET_TYPE sd, int af, int * recv_set, int
 SOCKET_TYPE picoquic_open_client_socket(int af)
 {
 #ifdef _WINDOWS
+    WSADATA wsaData = { 0 };
+    WSA_START(MAKEWORD(2, 2), &wsaData);
     SOCKET_TYPE sd = WSASocket(af, SOCK_DGRAM, IPPROTO_UDP, NULL, 0, WSA_FLAG_OVERLAPPED);
 #else
```
huitema: Looks good, but that means calling WSA_START several times, once per socket.

huitema:
Looking at the [WSAStartUp page on MSDN](https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-wsastartup) shows the potential issue with calling that function multiple times: _An application must call the WSACleanup function for every successful time the WSAStartup function is called._ I am not sure how big a deal that is in practice, because we never call WSACleanup. According to the MSDN page, that should be OK as long as all calls are for the same Winsock version.

steschu77:
We can make sure the WSAStartUp is called really once using some global
variable. I know that is ugly, I just want to know your opinion :)

huitema:
Yes, I know that there is an ugly solution. Let's just check the code as is, and wait for someone to complain about memory leak or whatever...

steschu77:
Ok, I'll just add the double socket open test.

steschu77:
@huitema please review the test and merge if it's ok.

PR #738, opened 2019-12-09
One way delay

huitema:
Add a timestamp to the ACK format, controlled by a negotiated option. PR is not ready yet, still need to pass the measured one way delays to the congestion controller, and also probably add them to the binary log somehow.


huitema:
@steschu77 please check that the way I added one-way-delay events in the qlog is not too messed up!

huitema:
Looking at the [WSAStartUp page on MSDN](https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-wsastartup) confirms the issue with calling that function multiple times: _An application must call the WSACleanup function for every successful time the WSAStartup function is called._ I am not sure how big a deal that is in practice.

picoquic/logger.c
```diff
@@ -581,30 +590,47 @@ size_t picoquic_log_stream_frame(FILE* F, uint8_t* bytes, size_t bytes_max)
...
 
-    int ret = picoquic_parse_ack_header(bytes, bytes_max, &num_block, NULL,
-        &largest, &ack_delay, &byte_index, 0);
+    if (!has_one_way_delay) {
+        ret = picoquic_parse_ack_header(bytes, bytes_max, &num_block,
+            &largest, &ack_delay, &byte_index, 0, NULL);
+    }
+    else {
+        ret = picoquic_parse_ack_header(bytes, bytes_max, &num_block,
+            &largest, &ack_delay, &byte_index, 0, &one_way_delay);
```
steschu77: ternary operator for `..., has_one_way_delay ? &one_way_delay : NULL);`

steschu77:
Are there any documents regarding this addition of one-way delay in ACK messages?

I'd like to see a receive timestamp per acknowledged packet in the ACK message, that's why I am curious what are current drivers in this extension.

huitema:
The main driver is the work on satellite communication (issue #720 ). Traditional congestion control algorithms rely on packet losses as indication of congestion. Given the long delays and large number of packets in flight, that can be very inefficient -- especially in scenarios that couple satellite links and a lossy Wi-Fi link. It is thus much more efficient to have delay based controls, but these work much better if we can estimate one-way delays. One-way delays also enable much better estimation of bandwidth, without having to filter for ACK-compression.

In theory, it would be interesting to have one way delay estimates for every packet but in practice that not necessary. For effective congestion control, we need several signals per RTT -- maybe 8 or 16. After that, it makes no difference. The same goes for bandwidth evaluation -- it is sufficient to evaluate bandwidth once per RTT, or maybe a few times. But in the satellite experiments we can easily have 20000 packets in flight, and adding 20000 time-stamps would be a bit wasteful. Also, the links are asymmetric, and it becomes important to keep ACK size relatively small.

Ian Swett of Google had proposed a design very similar to what I implemented -- just one time stamp per ACK. I sent him a message to ask whether he has a draft out. If not, I will write a small draft describing the "one way delay extension".

PR #742, opened 2019-12-14
Cubic 1wd

huitema:
Add version of Cubic that relies on delays (one way if available) instead of losses. Verifies that it provides plausible results when testing in high random loss scenario. Fix CPU cost explosion on high random losses by using splay for managing partially received stream data.

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -2732,7 +2732,7 @@ int picoquic_prepare_ack_frame(picoquic_cnx_t* cnx, uint64_t current_time,
             /* Set the lowest acknowledged */
             lowest_acknowledged = pkt_ctx->first_sack_item.start_of_sack_range;
             /* Encode the ack blocks that fit in the allocated space */
-            while (num_block < 63 && next_sack != NULL) {
+            while (num_block < 32 && next_sack != NULL) {
                 size_t l_gap = 0;
                 size_t l_range = 0;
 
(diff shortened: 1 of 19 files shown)


picoquic/frames.c
```diff
@@ -933,44 +932,73 @@ static int picoquic_queue_network_input(picoquic_cnx_t* cnx, picoquic_stream_hea
...
 
-        if (data_length > 0) {
-            picoquic_stream_data_t* data = (picoquic_stream_data_t*)malloc(sizeof(picoquic_stream_data_t));
+        /* Check whether parts of the frame are covered by the next frames */
+        while (ret == 0 && start < length) {
+            data_length = length - start;
 
-            if (data == NULL) {
-                ret = picoquic_connection_error(cnx, PICOQUIC_ERROR_MEMORY, 0);
+            if (next != NULL && next->offset < offset + length) {
```
steschu77: When I understand correctly, this may lead to unintended data loss. Consider the following example:

A block of 6 bytes is received at offset 2. And there is already another block of 2 bytes received at offset 4:

```
           0 1 2 3 4 5 6 7 8 9
available:         xxxx
received:      xxxxxxxxxxxx
```

With this condition you would threw bytes 4 to 7, from which bytes 6 and 7 haven't been received yet.

Or maybe I just don't understand :)

steschu77: Ah, ok. Turns out, I didn't understand. Since there is a while loop around this condition, it eventually also handles bytes 6 and 7 mentioned above in a later iteration.

huitema: Sorry that this is hard to read. I will check that we have a test case covering the overlap scenario.

huitema: Good thing you flagged that, there was in fact a bug and the new test found it.

PR #744, opened 2019-12-16
Refactor Stream Data Receiver

steschu77:
A proposed rework of picoquic_queue_network_input for easier understanding.

This is mainly for discussion. Please only approve if it **really** does simplify code reading.

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -938,15 +938,17 @@ static int add_chunk_node(picosplay_tree_t* tree, uint64_t offset, size_t length
 }
 
 /* Common code to data stream and crypto hs stream */
-static int picoquic_queue_network_input(picoquic_stream_head_t* stream, uint64_t stream_ofs, uint8_t* bytes, size_t length, int * new_data_available)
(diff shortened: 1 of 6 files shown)


picoquic/frames.c
```diff
@@ -912,95 +912,85 @@ void picoquic_stream_data_callback(picoquic_cnx_t* cnx, picoquic_stream_head_t*
...
+    }
+
+    return ret;
+}
+
 /* Common code to data stream and crypto hs stream */
-static int picoquic_queue_network_input(picoquic_cnx_t* cnx, picoquic_stream_head_t* stream, uint64_t offset, uint8_t* bytes, size_t length, int * new_data_available)
+int picoquic_queue_network_input(picosplay_tree_t* tree, uint64_t consumed_offset,
+    uint64_t stream_ofs, const uint8_t* bytes, size_t length, int* new_data_available)
 {
```
huitema: Not sure what `stream_ofs` stands for. Should it be `chunk_offset`?

huitema: OK, I see chunk_offset used later. `frame_data_offset` would do.

PR #745, opened 2019-12-17
Restore call to WSAStartup in Picoquic demo

huitema:
Turns out that there are many more calls to Winsock than just socket creation -- name resolution, for example. The picoquicdemo.exe was broken.


steschu77:
Thanks for fixing this. Too bad that this is not easy to provide by picoquic.

PR #747, opened 2019-12-19
Fix hot loop in logwriter, and also add log of 1WD frames.

huitema:
Fixes a hot loop bug bug in the log writer. When parsing a packet composed of random bytes, the first byte may by random chance be the type of ACK frames. The code would parse the highest ck number, and then the number of blocks. Since that it random, it could be a very large 62 bit numbers. And the code would then loop on that, even if the parsing detected an error.


huitema:
Also found a bug in the logging of application close frames -- this is not quite the same code as connection close, connection close has 1 extra parameter.

PR #748, opened 2019-12-19
Add test for binlog to qlog conversion

steschu77:
This adds a QLOG conversion step to the binlog test. It requires some more meta data than just the frames that were logged so far.

Two minor issues have been identified with the test:

1. The ECN frames broke the QLOG JSON format.
2. The `default_spin_policy` was not initialized in `picoquic_create`


picoquic/logwriter.c
```diff
@@ -634,7 +634,7 @@ void binlog_close_connection(picoquic_cnx_t * cnx)
     bytestream_buf stream_msg;
     bytestream * msg = bytestream_buf_init(&stream_msg, BYTESTREAM_MAX_BUFFER_SIZE);
     bytewrite_cid(msg, &cnx->initial_cnxid);
-    bytewrite_vint(msg, picoquic_current_time());
```
steschu77: Unfortunately, I had to revert this to 0. Otherwise the binary file comparison would complain about changes in the file. This isn't used right now in the converter, so it doesn't hurt.

I couldn't find a way to do this right (i.e. using simulated time during tests), so this is 0 for now.

steschu77:
The issue with the "not initialized" `spin_policy` is actually not that it is not initialized, but that it is random when set to '0', i.e. basic.

```
        if (cnx->spin_policy == picoquic_spinbit_basic) {
            uint8_t rand256 = (uint8_t)picoquic_public_random_64();
            if (rand256 < PICOQUIC_SPIN_RESERVE_MOD_256) {
                cnx->spin_policy = picoquic_spinbit_null;
            }
        }
```

picoquic/logwriter.c
huitema: You want to call:
```
uint64_t picoquic_get_quic_time(picoquic_quic_t* quic); /* connection time, compatible with simulations */
```
That's will be either the simulated time during tests, or the wall time in production.

picoquic/logwriter.c
```diff
@@ -647,33 +647,40 @@ void binlog_close_connection(picoquic_cnx_t * cnx)
...
+        DBG_PRINTF("Cannot open file %s for write.\n", binlog_file);
+    }
+    else {
+        /* Write a header text with version identifier and current date  */
+        bytestream_buf stream;
+        bytestream* ps = bytestream_buf_init(&stream, 16);
+        bytewrite_int32(ps, FOURCC('q', 'l', 'o', 'g'));
+        bytewrite_int32(ps, 0x01);
+        bytewrite_int64(ps, creation_time);
+
```
huitema: For tests, be careful to set that to the simulated time of the test!

picoquic/logwriter.c
```diff
@@ -647,33 +647,40 @@ void binlog_close_connection(picoquic_cnx_t * cnx)
...
+int binlog_open(picoquic_quic_t* quic, char const* binlog_file)
 {
     int ret = 0;
-    binlog_close(quic);
 
+    binlog_close(quic);
     if (binlog_file != NULL) {
-        quic->f_binlog = picoquic_file_open(binlog_file, "wb");
+        quic->f_binlog = create_binlog(binlog_file, picoquic_current_time());
         if (quic->f_binlog == NULL) {
```
huitema: You should use the simulated time there too.

picoquictest/skip_frame_test.c
```diff
@@ -799,33 +804,105 @@ int logger_test()
...
+            ret = qlog_convert(&srce_cid, f_binlog, binlog_test_file, ".");
+            if (ret != 0) {
+                DBG_PRINTF("%s", "Cannot convert the binary log into QLOG.\n");
+            }
+            else {
+                /* When changing the reference QLOG file please verify the new file at:
+                   https://qvis.edm.uhasselt.be/#/files */
+                ret = picoquic_test_compare_text_files(qlog_test_file, qlog_test_ref);
+            }
         }
```
huitema: That code will only perform the Qlog comparison if the bin file binaries are identical. That's a good way to test qlog.c, but that does not solve the "binary debug" problem. Suppose that we change the log content for any reason: maybe we have added a new frame type, maybe we have changed the format of some frame. The binaries will be different. If they are different "in the way that we expect", then life is good, we should just replace the reference bin log by the new version. But if they are different in a bad way, we should fix the code and leave the reference untouched.

Maybe what we need is a structured comparison. Read the binary log and the reference record by record, and report on the differences. But failing that, we could run the qlog comparison all the time, even if the binaries are different. Then when the test fail we can inspect the qlog file to understand what changed.

steschu77: I am ok with removing the binary comparison all together. Binary to QLOG conversion should be bijective one day anyhow.

huitema: Works for me.

picoquic/logwriter.c
huitema: Works now.

huitema:
The PR fails the memcmp test, which is about checking constant time memory compare. The failure happens on Windows 64 bit in release mode. That's weird, because it does not repro on my machine. You can ignore that for now and check in.

PR #749, opened 2019-12-30
Bbr

huitema:
Implementation of Google's BBR congestion control algorithm, based on https://tools.ietf.org/id/draft-cardwell-iccrg-bbr-congestion-control-00.html


huitema:
This PR only implements the "common path" of BBR. The "Probe RTT" part of BBR is implemented but not tested. It kicks in if no min RTT is observed for 10 seconds, and there should be a test designed to simulate a connection lasting more than 10 seconds, maybe 20 or 30 sec, and exercising this path. The "total loss" part of BBR is partially implemented but not connected to Picoquic, because the corresponding signal is not available yet.

huitema:
Checking this as is, the "high loss" signal will come in a separate PR.

PR #828, opened 2020-02-25
Debug bin log

huitema:
Fix bin log so it can produce enough qlog to replace text logs, and enough binary stats to replace separate cc_log. Unify binlog and cc_log. Remove separate cc_log APIs and code.


steschu77:
I'll look at it this afternoon which will be in about 6 hours :)

picolog/picolog.c
```diff
@@ -235,5 +263,56 @@ int convert_svg(const picoquic_connection_id_t * cid, void * ptr)
...
+
+int filedump_binlog(FILE* bin_log, FILE* bin_dump)
+{
+    int ret = 0;
+    uint8_t head[4];
+    bytestream_buf stream_msg;
+
+    fseek(bin_log, 16, SEEK_SET);
+
+    fprintf(bin_dump, "MDG-len, I-CID, Time, ID, Comment\n");
```
steschu77: Typo: MSG-len?

picoquic/logwriter.c
```diff
@@ -852,8 +794,15 @@ void picoquic_cc_dump(picoquic_cnx_t * cnx, uint64_t current_time)
 
     bytewrite_int32(ps_head, (uint32_t)bytestream_length(ps_msg));
 
-    (void)fwrite(bytestream_data(ps_head), bytestream_length(ps_head), 1, cnx->cc_log);
-    (void)fwrite(bytestream_data(ps_msg), bytestream_length(ps_msg), 1, cnx->cc_log);
+    if (cnx->cc_log != NULL) {
```
steschu77: Since the code for opening a cc_log has been removed, is it still possible to come here?

picoquic/util.c
```diff
@@ -667,8 +667,10 @@ uint8_t* picoquic_frames_l_v_encode(uint8_t* bytes, const uint8_t* bytes_max, si
 {
     if ((bytes = picoquic_frames_varlen_encode(bytes, bytes_max, l)) != NULL &&
         (bytes + l) <= bytes_max) {
-        memcpy(bytes, v, l);
-        bytes += l;
+        if (l > 0) {
```
steschu77: I think this check is a bit superfluous since memcpy can handle copying 0 bytes.

loglib/qlog.c
```diff
@@ -84,6 +88,276 @@ int qlog_packet_start(uint64_t time, uint64_t size, const picoquic_packet_header
...
+void qlog_truncated_sting(FILE* f, bytestream* s, uint64_t l, uint64_t l_max)
+{
+    uint64_t ll = (l < l_max) ? l : l_max;
+    uint64_t x;
+    int error_found = (s->ptr + (size_t)ll > s->size);
+
+    fprintf(f, "\"");
+
+    for (x = 0; x < ll && s->ptr < s->size; x++) {
+        fprintf(f, "%02x", s->data[s->ptr++]);
```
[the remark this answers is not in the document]
huitema: Actually, the intent was "process l bytes but only print out the first l_max bytes," but looking at the use cases I ended up printing all the bytes in these instances.

My goal is to reach parity between the qlog and the current text log. One of the gaps is the "preview of stream content". When processing a stream frame, the text log displays the first bytes in the frame. That has proven to be a useful debugging feature, but qlog has not provision for that. The current "truncated log" function was meant to support that, but it is probably inadequate. We will need to extend the data logged for stream frames in the log writer, and use a qlog extension to indicate "partial content" or "start with" or something like that. Same for the datagram frame and the crypto frame.

picolog/picolog.c
huitema: Yes...

picoquic/logwriter.c
huitema: Let me check. The updated code writes the congestion control reports in the "binlog" frame. I have verified that doing that I can extract the same CSV files -- the packet trace test passes. The intent is to document congestion control events in the qlog, but I have not added that code yet.

picoquic/util.c
huitema: OK.

PR #843, opened 2020-03-09
Remove address length from packet send receieve API

### LINKED ISSUE #838: Remove address length from API
Several picoquic APIs pass both socket address and address length. This is similar to the socket API like accept or connect, but it is not strictly necessary because the only support address families are AF_UNDEF, AF_INET and AF_INET6, and in those cases the length is easy to derive from the type.

Moreover, the type of the length is inconsistent. The socket API use a type socklen_t. Some picoquic API do the same, but others use int.

huitema:
Close #838


huitema:
@bkchr This may affect the API that you use.

PR #861, opened 2020-03-21
Verify retry dcid

### LINKED ISSUE #850: Authenticate Connection ID
This is QUIC PR https://github.com/quicwg/base-drafts/pull/3499.

The PR introduces two new parameters:

* handshake_connection_id (0x0f): The value that the endpoint included in the Source Connection ID field of the first Initial packet it sends during the handshake.

* retry_connection_id (0x10): The value that the the server included in the Source Connection ID field of a Retry packet; see {{cid-auth}}.  This transport parameter is only sent by a server.

Sending the first parameter is never an issue: it is the value that the peer currently uses as source address. Servers need to remember the retry connection ID in the same way that they remember the ODCID, so they can fill the second parameter.

There is a related problem with NEW TOKEN. It could be abused if a botnet master obtains a NEW TOKEN and broadcast it to all members of the botnet. Plausible mitigation is to make sure that all connections to the same token end up matching the same connection context, or maybe to treat the token as invalid if the connection context already exists. This would mean deriving the new SCID from the token value.

huitema:
Close #850


huitema:
Merging despite the travis builds failing to start for mac os. Will investigate that later.

PR #869, opened 2020-03-30
Error if no ALPN provided at client

### LINKED ISSUE #868: picoquic_initialize_tls_stream should fail if no ALPN is provided
I am building a minimal picoquic client/server application. I noticed that to get the client connected to the server it needs to provide an ALPN. In my case only the server complains about the missing ALPN, which is difficult to find. It would be better if the client already returns an error during context initialization if no ALPN has been provided.

huitema:
The ALPN is mandatory in Quic connections. Clients can provide it in two ways:

1) By entering the "ALPN" parameter when creating a connection context
2) By responding to the "picoquic_callback_request_alpn_list" request in the client call back function and providing a list of supported ALPN.

This PR adds a check that the client provides at least on ALPN before trying to establish a connection. If no ALPN is provided, the error `PICOQUIC_ERROR_NO_ALPN_PROVIDED` is returned by the function `picoquic_initialize_tls_stream`, through the API call to `picoquic_start_client_cnx` or `picoquic_create_client_cnx`.

Close #868


steschu77:
Works perfectly! Thanks for the prompt solution, Christian!

huitema:
@steschu77 thanks for the review. I added a test.

PR #891, opened 2020-04-27
Remove dependencies from h3zero to picoquic

huitema:


larseggert:
PR works, no longer need to link against picoquic-core!

PR #893, opened 2020-04-28
Remove all dependencies on the picotls minicrypto binding

larseggert:


larseggert:
CI fails, because the openssl version included with xenial is way too old (1.0.2). You should update your CI setup to use bionic.

picoquic/tls_api.c
```diff
@@ -1033,14 +1032,14 @@ void picoquic_crypto_context_free(picoquic_crypto_context_t * ctx)
 
 /* Definition of supported key exchange algorithms */
 
-ptls_key_exchange_algorithm_t *picoquic_key_exchanges[] = { &ptls_openssl_secp256r1, &ptls_minicrypto_x25519, NULL };
+ptls_key_exchange_algorithm_t *picoquic_key_exchanges[] = { &ptls_openssl_secp256r1, &ptls_openssl_x25519, NULL };
 ptls_cipher_suite_t *picoquic_cipher_suites[] = {
```
huitema, line 1036: There should be an #ifdef here, to make sure to not attempt linking with ptls_openssl_x25519 if it is not supported. Something like:
```
#ifdef PTLS_OPENSSL_HAVE_CHACHA20_POLY1305
ptls_key_exchange_algorithm_t *picoquic_key_exchanges[] = { &ptls_openssl_secp256r1, &ptls_openssl_x25519, NULL };
#else
ptls_key_exchange_algorithm_t *picoquic_key_exchanges[] = { &ptls_openssl_secp256r1, NULL };
#endif
```

huitema:
@larseggert I am merging this now, as it provides the desired goal: remove links to mincrypto; enable ChaCha20 if the platform supports it; do not use it if it is not supported.

PR #895, opened 2020-04-28
xenial -> bionic

larseggert:
Also fix nits

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
-dist: xenial
-language: minimal
-sudo: required
+dist: bionic
+os: linux
+language: c
+
 addons:
   apt:
     sources:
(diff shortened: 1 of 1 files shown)


huitema:
Does Travis support Focal already?

I think there are two issues: test all the features in the code, and check that common platforms are supported too. For testing the code, Bionic is fine. For testing platform support, the Ubuntu supported versions are Xenial (uintil 2021), Bionic and Focal (and Eoan for a few more months). I think we should maintain at least one test for Bionic, for the same reasons that we test old versions of OSX.

huitema:
Looking at the list of tests in the config, I would suggested keeping the "cpp-check" test Xenial for now, until Ubuntu officially drops support for Xenial next year.

I just updated and merged PR #893 -- using #ifdef to remove the references to ChaCha20 if not supported in the current platform.

huitema:
Forget the suggestion to use xenial for tests. Since there is a difference between the supported algorithm, there is also a difference in the size and content of the Client Hello. This means that the QLOG tests will have different outcomes on the two systems. Not desirable. We will just have to test xenial support manually.

huitema:
And of course, the problem that I feared for xenial also occurs for MacOS xcode 10.1. SO we might just as well try xenial after all.

Issue #922, opened 2020-05-21
Error when running the multi-file test on interop-runner


huitema:
Out of 50 files, one was corrupted. Logs at: https://interop.seemann.io/logs_2020-05-20T04:20:45UTC/picoquic_picoquic/multiplexing/

huitema:
Could not reproduce the exact issue, but found a need to tighten the closure of files on the server. Added a multi-file test in PR #926, and demonstrated sending 1999 files in parallel on all supported platforms.

huitema:
There was indeed an issue hidden by the file closure problem -- see bug #935, being fixed in PR #936.

PR #926, opened 2020-05-22
Add multi-file test

huitema:
First version of a test that replicates the M test in the interop runner -- see issue #922. The test execution in VS debugger is very slow, due to need to create many files. More importantly, the test appears to fail on the Windows client with diagnostic "too many files open" when processing the 277th file. This is bizarre, because the tracking of open and close files shows that the client has only one file opened at that point.

### DIFF
--- a/UnitTest1/unittest1.cpp
+++ b/UnitTest1/unittest1.cpp
@@ -1372,6 +1372,12 @@ namespace UnitTest1
             Assert::AreEqual(ret, 0);
         }
 
+        TEST_METHOD(h3_multi_file) {
+            int ret = h3_multi_file_test();
+
+            Assert::AreEqual(ret, 0);
+        }
+
         TEST_METHOD(http_stress) {
             int ret = http_stress_test();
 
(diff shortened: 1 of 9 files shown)


huitema:
First runs of the tests shows that the Linux builds do support sending 1999 files in parallel, but the Mac OS builds do not. The test failures are difficult to diagnose, and the error found in issue #922 is not reproduced.

huitema:
Changes: do not run for a very large number of files in the non-regression test, just run for 128 files to check that the test works, and add an option in `picohttp_ct` to test with an arbitrary number of files. Also, add a binary log of both client and server side, for later debugging.

Two issues still: the creation of files takes a very long time; and, the bug #922 is not reproduced.

huitema:
The server was only closing files at the end of the connection. By closing as soon as the file is sent, we can keep the number of open files under the system's limit, not just on Linux but also on MAC OS and Windows. Restoring the max number of trials to the default 128 before checking in.

PR #939, opened 2020-06-04
Create Dockerfile

cmiller-NA:


huitema:
Thanks for contributing that!

This is great, but I wonder whether you could make this a two-pass Dockerfile, much like for example [the docker file developed for the Quic interop runner?](https://github.com/private-octopus/picoquic-interop-runner/blob/master/Dockerfile)

The problem with the single pass file is that the image will include the entire development environment, resulting in a large size, over 250MB. The Picoquic image for the interop runner is less than 50 MB, and that could be possibly lowered by having a more selective copy of picoquic binaries and files to the final image. Also, you will end up having to start from a scratch Ubuntu image, load the tools and compile Picotls after each change to Picoquic, which is somewhat slow.

huitema:
@cmiller-NA maybe I do not get what do you want to use the docker image for... If your goal is to merely use it as the first stage for an application's image, the current project is probably fine. Just add a few comments explaining how people are supposed to use it, and we will be good.

cmiller-NA:
I used that Dockerfile as a quick way to build, explore and try out picoquic. Docker caches each stage so changing the source won’t require a complete rebuild of the whole image, just the last step.

-Chris

huitema:
OK, let's check this as is. If someone expresses different requirements, we can always start a new PR!

PR #975, opened 2020-07-06
Address memory leaks detected by cpp-check

huitema:


huitema:
@alagoutte No more CPPCheck warnings. I think the leaks were theoretical -- for example, one would only happen if the code tried to open the key log file twice in the same session, and we never do that. But in any case, it is fixed.

PR #977, opened 2020-07-07
sample: Add documentation

tbarbette:
Show usage, how to generate certs, etc. We had students bumping on this so...

### DIFF
--- a/sample/README.md
+++ b/sample/README.md
@@ -5,7 +5,7 @@ The sample program is a simple QUIC client/server demo.
 
 Building
 --------
-picoquic\_sample is built as part of the compilation process of picoquic. It is
+picoquic\_sample is built as part of the compilation process of picoquic. It
 will be available in the root folder.
 
 Usage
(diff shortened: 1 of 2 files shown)


sample/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+Example
+-------
+
+Generate the certificates:
+```
+openssl req -x509 -newkey rsa:4096 -days 365 -keyout ca-key.pem -out ca-cert.pem
+openssl req -newkey rsa:4096 -keyout server-key.pem -out server-req.pem
+```
+These commands will prompt a few questions, you don't need to put actual data
```
huitema, line 26: RSA 4096 will make the initial exchange kind of slow. Using ECDSA is a bit more complex, so I get sticking with RSA for the sample is fine, but 4096 is probably overkill. 2048 would be adequate.

sample/README.md
```diff
@@ -0,0 +1,43 @@
+picoquic sample
+===============
+
+The sample program is a simple QUIC client/server demo.
+
+Building
+--------
+picoquic\_sample is built as part of the compilation process of picoquic. It is
+will be available in the root folder.
+
```
huitema, line 10: Typo, "it is will"

sample/README.md
```diff
@@ -0,0 +1,43 @@
...
+```
+mkdir server_files
+echo "Hello world!" >> ./server_files/index.htm
+```
+
+And run the server:
+```
+sudo ./picoquic_sample server 443 ./ca-cert.pem ./server-key.pem ./server_files
+```
+
```
huitema, line 39: Or run on high port number if you don't want to use sudo. Most of the QUIC interop work was done on port 4433.

sample/README.md
```diff
@@ -0,0 +1,43 @@
...
+Then, test if you can reach it using the client:
+```
+sudo ./picoquic_sample client localhost 443 /tmp index.htm
+```
```
huitema, line 43: You don't need sudo there.

tbarbette:
Should be done ;) Thanks for the review!

PR #986, opened 2020-07-11
Create a binary log file per connection

steschu77:
Currently picoquic writes one giant binary log file that contains all traffic that picoquic handles across all connections. The advantage is that even context-free information can be written and later matched to the correct connection once the connection id is known. The disadvantage is that it is hard to implement log-file rotation and it is more difficult to extract connection data from this log file when a connection is closed.

This change allows picoquic to be configured with a directory where log files are written per connection.

There are a couple of failing unittests where I'd need some help with.

* Congestion control tests app_limit_cc and pacing_update are failing
* packet_trace is failing because the connection id seems to be random which is an issue now since the log files are named after their connection id.
* qlog_trace fails since packet sizes seem to be different now and I don't have any idea why.


picoquic/logwriter.c
```diff
@@ -736,38 +742,65 @@ void binlog_picotls_ticket(FILE* f, picoquic_connection_id_t cnx_id,
...
-    /* Algorithms used */
-    bytewrite_cstr(msg, cnx->congestion_alg->congestion_algorithm_id);
-    bytewrite_vint(msg, cnx->spin_policy);
+    char log_filename[512];
+    if (ret == 0) {
+        if (picoquic_sprintf(log_filename, sizeof(log_filename), NULL, "%s%s%s.log",
+            cnx->quic->binlog_dir, PICOQUIC_FILE_SEPARATOR, cid_name) != 0) {
+            ret = -1;
+        }
+    }
```
huitema, line 769: This is good, but it also is the reason why the "packet_trace" test is failing. The test assumes that the binlog is in the file PACKET_TRACE_BIN in the current directory. With the new API, the trace would be in the folder PACKET_TRACE_BIN, with a name derived from the initial CID. Same problem for the "app_limit_cc" test: the code in `app_limit_cc_test_one` assumes that the trace is in APP_LIMIT_TRACE_BIN.

In both cases, the solution is to create the log in the current working directory -- set the folder name to "." or either "./" or ".\" depending on OS, derive a dynamic file name from the initial CID of `test_ctx->cnx_client`, and use that instead of the constant value.

There are more tests that generate bin logs today, and those are probably failing to generate these logs as well. I can deal with that later.

huitema: Looking further, there is another alternative: add a parameter to the test context creation to use a predictable initial CID, much like what is done in the `qlog_trace` test. This is done in an ad-hoc way in the qlog test, delete the created connection and then create a new one, but that could be simplified by passing the initial CID as a parameter in (tls_api_init_ctx), so it could then be used in that code when creating the client connection context.

steschu77:
> Very simple, with relatively small code changes. This is great. I added a comment explaining why the tests are failing. I can add the corresponding code if you want, but that means doing a PR on a PR. Tell me if you want me to do that.

Yes, please add the neccessary changes. You could use my branch, so there isn't another PR

> Very simple, with relatively small code changes. This is great. I added a comment explaining why the tests are failing. I can add the corresponding code if you want, but that means doing a PR on a PR. Tell me if you want me to do that.

Yes, please add the neccessary code changes. You can use my branch "one_binlog_per_connection" to do that so there won't be another PR on a PR.

Thanks, Christian!

huitema:
I did the minimal fixes to make the failing tests work. I may need to do a couple more fixes to get a log file for some of the other tests.

huitema:
Please don't check it yet, I need to also update some dependencies and docs.

steschu77:
Thanks again. Let me know when I should have a look.

huitema:
Sorry for the long delay. I had to check the demo and the sample, so people would not cut and paste the wrong code. I also added the "client" or "server" suffix in the names of the logs, because people sometimes use the same folder for client and server logs. And then I made sure that all the tests for which logs are produced use descriptive initial connection ID, so all these tests produce different binary files. I think I am done now, barring bugs.

huitema:
@steschu77 please take a look, and then merge if you like it.

steschu77:
Thanks, Christian, for fixing, variable renamings, and also updating the demo!

PR #987, opened 2020-07-13
Auto qlog

huitema:


picoquic/logwriter.c
```diff
@@ -746,7 +748,8 @@ FILE* create_binlog(char const* binlog_file, uint64_t creation_time);
 
 void binlog_new_connection(picoquic_cnx_t * cnx)
 {
-    if (cnx == NULL || cnx->quic->binlog_dir == NULL) {
+    char const* bin_dir = (cnx->quic->binlog_dir == NULL) ? cnx->quic->qlog_dir : cnx->quic->binlog_dir;
+    if (cnx == NULL || bin_dir == NULL) {
```
steschu77, line 752: If cnx was NULL then you'd have crashed the line before already :)

picoquic/logwriter.c
```diff
@@ -822,6 +829,42 @@ void binlog_close_connection(picoquic_cnx_t * cnx)
     fflush(f);
 
     cnx->f_binlog = picoquic_file_close(cnx->f_binlog);
+
```
steschu77, line 832: Would it be possible to have a #define here such that you can exclude the QLOG stuff from picoquic for people who like the "pico" in picoquic? :) This also includes all the additional files you added to the picoquic core library for the QLOG conversion.

picoquic/logwriter.c
huitema: Removed the extra test. It would mess up static analysis.

picoquic/logwriter.c
huitema: I did not use a #define. Instead, I used a function pointer. I moved the files back to the log library, and added there an autoqlog file that defines `picoquic_set_qlog` and documents the function pointer. Applications that want to use that API will have to link the log library, and those that do without get a smaller code size.

huitema: @steschu77 I think that fixes the code size issue that you mentioned. Will further reduce it once we get rid of the text log.

huitema:
Merging so I can prepare the setupfor the interop on 7/14.

picoquic/logwriter.c
steschu77: Thanks, @huitema ! That is exactly what I was looking for!

PR #1010, opened 2020-08-16
Socket loop perf

huitema:
All picoquic based applications incorporate a "packet loop" that is a variation of:
```
while (something) {
    // ask the quic context for the duration of the next timer
    // wait for arrival of packets within that timer
    if (a_packet_arrived) {
        // submit the packet to the picoquic context
    }
    // Ask picoquic if there is a packet send
    if (something_to_send){
        // Send it
    }
}
```
This PR develops a function `picoquic_packet_loop` that executes this loop.

There are two goals in this development:

1) Simplify writing of picoquic applications by removing a lot of code.
2) Allow development of platform specific optimizations, such as using WSA sockets in Windows, or sending multiple messages in some versions of Linux. These code changes can be embedded in the packet loop function, so application developers don't have to concern themselves so much with platform dependent socket code.


huitema:
The first check-ins enable the packet loop for servers, such as the demo server or the sample server. This is not ready to merge, as enabling clients is more challenging and may lead to changes in the API.

PR #1014, opened 2020-08-18
Fix casing of 1RTT/0RTT for qlog

martinthomson:
These match the values in the spec.

### DIFF
--- a/loglib/logconvert.c
+++ b/loglib/logconvert.c
@@ -34,9 +34,9 @@ const char * ptype2str(picoquic_packet_type_enum ptype)
     case picoquic_packet_handshake:
         return "handshake";
     case picoquic_packet_0rtt_protected:
-        return "0rtt";
+        return "0RTT";
     case picoquic_packet_1rtt_protected:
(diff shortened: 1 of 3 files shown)


huitema:
This is failing because of unit tests of the qlog trace functionality. The tests run a connection and compare the qlog to an expected value. If something changes, the tests break. The expectation is that if you change the trace production, you should also change the reference test files to match what you expect, and verify that something else did not break by mistake. In your case, apply s/0RTT/0rtt/ and s/1RTT/1rtt/ in the file "./picoquictest/qlog_trace_ref.txt".

huitema:
@martinthomson I understand you just wanted to make a quick fix, and that dealing with the test suite makes it more work. If that's too much, just tell me and I will fix the issue.

martinthomson:
I apologize for not running the tests.  I just wanted to get this under your nose; I hadn't had time to locate the tests.

huitema:
No problem, glad you found that. I always have doubts about the proper casing of JSON...

PR #1016, opened 2020-08-20
Log grease_quic_bit tp

martinduke:


martinduke:
Sure! The compact lots are great for quickly seeing what's going on. Qlog is very powerful but a little verbose IMO.

Issue #1022, opened 2020-08-26
Simultaneous Connection Load


victorstewart:
Do you have any data or intuition on how many simultaneous mostly-quiet connections Picoquic can tolerate? Will 100,000 cause it to fall over?

I'd need at least 50,000 (run as a containerized instance per logical core) to consider it for production workloads.

huitema:
I did not test that, and the old adage says "if it is not tested it probably does not work". How would you write that test?

victorstewart:
@huitema 

Well it would be all about monitoring the stability of the implementation's memory usage, throughput, and latency (and no crashes) with ever growing connection count (maybe in increments of 1000 connections). 

I assume packet loss and reordering rates depend on traffic volume not open connections.

So it might not be too complicated since we don't need any intermediate chaos boxes.

huitema:
I am not worried about the extra load for attributing a packet to a connection -- this is done with a hash table, and size 50000 would not be too much of an issue. I am also not too worried about the network effects, since we use a single UDP socket for all connections, although there might be something happening in the kernel implementation of UDP and of the ARP caches. I am worried about memory size, since that was not optimized for. I am also worried about other details, such as the shutting down of connections if their are idle for too long, or the extra load of maintenance traffic to avoid the idle mechanism. That tells me that any test should not just create batches of connections, but also simulate what happens when these connections remain idle for at least 10 minutes.

From a testing point of view, I am worried about having 50000 clients. Will the local machine let me create 50000 sockets? What about handling 50000 separate client processes? It seems that a loopback test is difficult to engineer. On the other hand, I do not have the hardware available to simulate a network of clients.

Most tests are done with a simulated network using simulated time, bypassing the system network stack entirely. I could imagine a "test N clients" test written like that, but I a concerned that the same thread will handle both the clients and the server. So the first testing iterations will probably need to deal with the perf of the test itself.

Of course, if you do have a test infrastructure available, I will be happy to look at the results.

huitema:
@victorstewart What is the traffic model for your applications? Some are easier than other. For example, if the transactions are initiated by the client, it is more efficient to just let the connection expire after an idle timer, and have the client set another connection when a new transaction is needed. With 0-RTT, this can be done without any extra delay. This is what is done for example for DNS over QUIC. But if the server is pushing data to the client at random intervals, then the connection needs to stay up -- and also needs to deal with issues like NATs and Firewalls.

victorstewart:
[...]

PR #1032, opened 2020-09-06
Functional test of many parallel cnx

huitema:
This is the unit test version of the connection stress. It simulates 100 connections on a single core -- in fact 100 client connections and 100 server connections. The test checks that the simulation runs faster that the real time clock -- on dev laptop, it runs in 1.1 sec, for a simulated duration of 120 sec. Two tasks left before checking this in and closing issue #1022:

1) Add an option to `picoquic_ct` to run this code in "stress" mode, with arbitrary target duration and arbitrary number of connections.

2) Run a first pass of performance analysis. If the simulation runs 100 connections and the associated load for 120 seconds in 1 seconds, it probably only runs about 10,000 connections on a single core, falling short of the 50,000 target. We need to understand first where the time is spent.

Depending on the analysis, we may be able to quickly remove some bottlenecks, or it may be more complicated, in which case individual bugs will be filed for the identified issues.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,8 +53,9 @@ set(PICOQUIC_TEST_LIBRARY_FILES
     picoquictest/ack_of_ack_test.c
     picoquictest/bytestream_test.c
     picoquictest/cleartext_aead_test.c
-    picoquictest/cplusplus.cpp
     picoquictest/cnx_creation_test.c
+    picoquictest/cnxstress.c
+    picoquictest/cplusplus.cpp
     picoquictest/hashtest.c
     picoquictest/intformattest.c
     picoquictest/multipath_test.c
--- a/PerfAndStressTest/PerfAndStressTest.cpp
+++ b/PerfAndStressTest/PerfAndStressTest.cpp
@@ -95,16 +95,20 @@ namespace PerfAndStressTest
             Assert::AreEqual(ret, 0);
         }
 
-        TEST_METHOD(fuzz)
-        {
+        TEST_METHOD(fuzz) {
             int ret = fuzz_test();
 
             Assert::AreEqual(ret, 0);
         }
 
-        TEST_METHOD(fuzz_initial)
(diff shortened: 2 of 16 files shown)


huitema:
Simulations shows that the execution time of the simulation increase more than linearly with the number of connections:


NB Clients | Wall Time (s) | `picoquic_reinsert_by_wake_time`
-- | -- | --
100 | 0.136934 |  
200 | 0.271403 |  
500 | 0.700534 |  
1000 | 1.497147 | 1.92%
2000 | 3.341739 | 4.44%
5000 | 10.46268 | 9.50%
10000 | 27.21576 | 50.47%
20000 | 93.48444 |  
50000 | 571.4684 |  

When the number of connections is small, the largest fraction of the CPU usage can be traced to the TLS stack. But when the number of connections increases, a growing fraction of the CPU time is spent in the function `picoquic_reinsert_by_wake_time` as seen on measurements for 1000, 2000, 5000 and 10000 connections. The number for 10000 connections is computed on the tail of the simulation, when all the TLS sessions are already established, and is thus probably not directly comparable to the figures for 1000, 2000 and 5000. However, this clearly points to a hot spot that should be fixed.

huitema:
Replacing the linear search in `picoquic_reinsert_by_wake_time` enables management of large sets of client connections. The cost of the simulation is now more or less linear with the number of simulated connections, as expected:


NB Clients | Wall Time (s) | picoquic_reinsert_by_wake_time | After fix
-- | -- | -- | --
100 | 0.137 |   | 0.140
200 | 0.271 |   | 0.279
500 | 0.701 |   | 0.714
1000 | 1.497 | 1.92% | 1.471
2000 | 3.342 | 4.44% | 3.010
5000 | 10.463 | 9.50% | 7.467
10000 | 27.216 | 50.47% | 14.624
20000 | 93.484 |   | 32.186
50000 | 571.468 |   | 71.735

PR #1046, opened 2020-10-01
Windows send multiple

huitema:


huitema:
Checking in, because the mechanism is solid, although the speed improvements on Windows failed to materialize.

PR #1052, opened 2020-10-02
Remove QUIC_ prefix in SSLKEYLOG

huitema:
Wireshark does not need these prefixes anymore, and will drop support in a coming release.


huitema:
@Lekensteyn can you take a look? Is this what you want?

PR #1058, opened 2020-10-12
Adding an install option to CMake

huitema:
This is a first attempt at installing executable, libraries and include files for Picoquic. There are obvious questions, e.g. where we should target generic location s, e.g., "/usr/local/bin", versus specific locations, e.g.,  "/usr/local/bin/picoquic/", or whether the file names such as "autoqlog.h" should have a more specialized name like "picoquic_autoqlog.h". Also, should we have a configuration file? But I would rather only do these changes after getting actual feedback from users.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@ endif()
 set(PICOQUIC_LIBRARY_FILES
     picoquic/bbr.c
     picoquic/bytestream.c
-	picoquic/cc_common.c
+    picoquic/cc_common.c
     picoquic/cubic.c
     picoquic/fastcc.c
     picoquic/frames.c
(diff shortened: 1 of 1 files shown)


tbarbette:
Tested and works perfectly on Ubuntu 18.04.

It would be also nice to install a pkg-config file so one can use `pkg-config --libs picoquic` to find the libs, `--exists` to check if it's installed, etc.

I can try to do something if you don't have time.

Then, doing the same for picotls, and then, eventually find the dependency automatically instead of looking for ../picotls :)

huitema:
@tbarbette For picotls, I started https://github.com/h2o/picotls/pull/327.

I am still worried about the name space issues, especially for the header files. Some of the names are not unique. Exporting "picoquic.h" seems fine, but for other files it might be better to but them in a subfolder. Picotls has the same problem, I will look at the solution adopted there and use the same for picoquic.

huitema:
The only questionable part of this PR is the naming of the include files "demoserver.h" and "democlient.h", whose names could well conflict with client and server demos developed by another application. If someone complaints, the names may need to be changed.

Issue #1057, opened 2020-10-11
Modular support of crypto suites


huitema:
Would it be possible to remove the dependency of Picoquic on OpenSSL? That dependency is somewhat annoying on Windows because it forces using a component not part of the system, and may be a blocker in some IOT deployments. Ideally, Picoquic should work just fine if OpenSSL was not present, and would only use the OpenSSL components if they are available and better than the default components.

Picoquic depends on the implementation of TLS 1.3 in Picotls, but it also depends on OpenSSL for the implementation of several algorithms: AESGCM, ChaCha/Poly, AES128_ECB, SHA256, secp256r1, x25519, SHA384. The encryption and key exchanges algorithms are provided to Picotls through an API, when the master TLS context is initialized. In Windows, the AESGCM, SHA256, SHA384 and secp256r1 components are available in the Bcrypt library. The Picotls library includes implementations of all algorithms as part of the minicrypto library. The new implementation of AESGCM in the fusion library is as fast or faster as that of OpenSSL, but the implementation of ChaCha/Poly is much slower. If we solve the performance issues, or if we decide that they are not important, we could initialize the TLS context using the Windows and/or Minicrypto libraries, removing that dependency.

These are not the only dependencies. The code calls "ptls_openssl_init_sign_certificate" to set up the public key of the server and "ptls_openssl_dispose_sign_certificate" to dispose it, "ptls_openssl_init_verify_certificate" to create and "ptls_openssl_dispose_verify_certificate" to remove the certificate verification code, "ptls_openssl_random_bytes" to initialize the random number generator. There are also error tracing functions that access open_ssl error messages. The Picotls code allows for calling substitutes for these functions, but apart from "random_bytes" these alternatives are limited to the support of "PTLS_SIGNATURE_ECDSA_SECP256R1_SHA256" in the minicrypto library: servers could only support ECDSA certificates and not RSA certificates; clients do not have good alternative for verifying certificates. Those would have to be developed.

Most dependencies are concentrated in "tls_api.c" for the Picoquic-core library, and in a few files for the test library.

Removing OpenSSL dependencies is a large body of work. We should decided whether we want to simply remove OpenSSL and always use a different code, use compile time #ifdef to compile with OpenSSL or an alternative crypto provider, or use an API to let application select among existing providers, or bring their own. Out of these, the "single option" has the advantage of simplicity, while the compile option makes testing hard. An API design would allow testing the "matrix" of providers, and could be the best solution, if the complexity can be mastered.

[...]

PR #1059, opened 2020-10-12
Propose new structure and modular support implementation [WIP]

BitFis:
This pull request contains a possible future implementation of a modularized way addressing #1057.

API parts:
- [ ] Certificate verification backend
- [ ] Key loading procedures
- [ ] Crypto alg. backend / calls to algorithm / implementation
- [ ] Encrypt / Decrypt functions (AEAD and one way functions)
- [ ] platform backend
  - [ ] entropy / RND functions

Current state:
- [ ] Define slim crypto API containing all picoQuic calls
  - [ ] handle the "cipher suite" issues (AEAD and one way functions)
  - [ ] handling of certificates
- [ ] Develop crypto implementations examples
  - [ ] openssl
  - [ ] minicrypto or fusion
- [ ] Enable defining endpoint during compilation or linking


picoquic/pictls_openssl_crypto.c
```diff
@@ -0,0 +1,25 @@
+/*
+* Copyright (c) 2020, The PicoQuic Authors.
+* All rights reserved.
```
huitema, line 3: I am OK with copyright to a specific author, but I also need a copyright to Private Octopus Inc., maybe delegated. Otherwise, source management could become very messy.

huitema: Also, please check the name. Cannot be "pictls_". Should be either "ptls_", the convention in picotls, or "picoquic". Or maybe just "openssl_crypto.c".

picoquic/pictls_openssl_crypto.c
```diff
@@ -0,0 +1,25 @@
...
+* LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+* ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+* (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+* SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+/**
+ *
+ */
+
```
huitema, line 24: We obviously need to define the API before populating that. For test purposes, it would be good to have not just one implementation, but at least two -- say one with openssl and one with minicrypto. That's a good way to find out what is really specific to an implementation, and what should stay in the common code in tls_api.c.

picoquictest/crypto_api_test.c
```diff
@@ -0,0 +1,39 @@
...
+    (void)loss_mask;
+    (void)proposed_version;
+    (void)sni;
+    (void)alpn;
+
+    return 0;
+}
+
+int crypto_api_test() {
+    return crypto_api_test_endpoint(NULL, PICOQUIC_INTERNAL_TEST_VERSION_1, PICOQUIC_TEST_SNI, PICOQUIC_TEST_ALPN);
```
huitema, line 38: That's where we need to consider testing at least 2 implementations, using execution time selection.

picoquictest/crypto_api_test.h
```diff
@@ -0,0 +1,38 @@
+/*
+* Copyright (c) 2020, The PicoQuic Authors.
+* All rights reserved.
```
huitema, line 3: Same remark as above on copyright to Private Octopus.

picoquictest/h3zerotest.c
```diff
@@ -74,7 +74,7 @@ static h3zero_test_integer_case_t h3zero_int_case[] = {
```
huitema, line 74: It seems that the only changes in these file are formatting changes. If that's the case, maybe we should leave it untouched. But then, we may indeed need a change somewhere, for example to set the default crypto.

huitema (commented):
Glad to see that you are starting this. I have a few comments, but that boils down to:

1) First define the API, probably based on a "crypto_context" structure containing pointers to data and functions.

2) In a first milestone, just handle the "cipher suite" issues (AEAD and one way functions)

3) Leave the handling of certificates to a  further milestone

4) Develop 2 crypto implementations, not just one, so we can test the switching of API, and also get a feeling about which code shall remain common.

picoquic/pictls_openssl_crypto.c
BitFis: True, I thought we could change it to a more generic Copyright since I am not part of Private Octopus Inc., I dont care to much about copyright, what about:

```c
/*
* Copyright (c) 2020, Private Octopus Inc., The PicoQuic Authors.
* All rights reserved.
```

This was inspired by https://github.com/openthread/openthread/blob/master/LICENSE.

picoquic/pictls_openssl_crypto.c
BitFis: got you, will go for `ptls_fusion_backend.c` and `ptls_openssl_backend.c`.

picoquictest/crypto_api_test.h
BitFis: 👍 I will add the files like this right now, but we can change it as soon as we defined what you prefer

picoquictest/h3zerotest.c
BitFis: still unsure what triggered it, will remove it

picohttp_t/picohttp_t.c
```diff
@@ -29,7 +29,7 @@
 #include <stdlib.h>
 
 extern size_t picohttp_nb_stress_clients;
-size_t picohttp_test_multifile_number;
+static size_t picohttp_test_multifile_number;
```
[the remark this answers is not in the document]
BitFis: #1061

picoquic/pictls_openssl_crypto.c
huitema: Private Octopus is my one person company. Not that I am making any money with the code, but i would rather have a clean copyright there. At some point I could set up something like a non-profit picoquic foundation, but that requires a fair bit of lawyers fee, so I would only do it if there is actual demand.

What you propose here, " Copyright (c) 2020, Private Octopus Inc., The PicoQuic Authors",  is fine.

picoquictest/picotls_openssl_crypto_backend.c
```diff
@@ -0,0 +1,23 @@
+/*
```
huitema, line 1: Is this a file containing test code of actual code?

picoquictest/h3zerotest.c
BitFis: should be moved now to #1061

BitFis: I proposed the usage of clang-format and editorconfig #1062, so future issues like those can be handled fluently during upload. Await your feedback there

picoquic_crypto/ptls_fusion_backend.c
```diff
@@ -0,0 +1,23 @@
+/*
+* Copyright (c) 2020, Private Octopus, Inc., The PicoQuic Authors.
```
huitema, line 2: I started working on integration of fusion -- the first step was done inside the Picotls project, to add the fusion code to a benchmark program. This requires changes in the generic AEAD implementation in tls_api.c. I think those changes need to be done before we start splitting some of the code in a crypto library. I will push a PR for that soon.

CMakeLists.txt
```diff
@@ -34,11 +34,14 @@ set(PICOQUIC_LIBRARY_FILES
     picoquic/spinbit.c
     picoquic/ticket_store.c
     picoquic/token_store.c
-    picoquic/tls_api.c
     picoquic/transport.c
```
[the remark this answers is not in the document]
huitema: @BitFis please revert this change. The "tls_api.c" code should stay in the "core" library for now.

picoquic_crypto/ptls_fusion_backend.c
BitFis: Great, I am starting with a basic skeleton of the crypto api as a suggestion, It then can be changed at a later state. Also be aware that I will force push this branch

picoquictest/picotls_openssl_crypto_backend.c
BitFis: This was ment to have only test code

CMakeLists.txt
```diff
@@ -34,11 +34,14 @@ set(PICOQUIC_LIBRARY_FILES
...
     picoquic/util.c
 )
 
+set(PICOQUIC_CRYPTO_LIBRARY_FILES
+    picoquic/tls_api.c
+)
+
```
huitema, line 44: Please don't do that. The glue to the picotls library really belongs in the core library.

I am not convinced that we need to handle the crypto implementations in a separate library, by opposition to keeping the files in the core library. Having the files in separate libraries just makes it harder for application. They will always need the crypto functions, so now they have to load two libraries instead of one.

Having a single library or two libraries has no effect on the size of the executable. Individual object files in the library only get loaded if the application uses functions or data from that file. If the crypto variant have different headers, the application will get to chose by inserting the chosen header and calling the "setup" function in the API.

See for example how "autoqlog" works.

CMakeLists.txt
```diff
@@ -117,8 +134,7 @@ add_library(picohttp-core
 )
 
 target_link_libraries(picohttp-core
-    ${PTLS_LIBRARIES}
-    ${OPENSSL_LIBRARIES}
+    picoquic-crypto-picotls-openssl
 )
```
huitema, line 138: I don't like the idea of separate libraries for each crypto variant. That is making the project more complex, and requires a cascade of changes in applications.

huitema:
As I said in a few comments, I don't like the idea of creating separate libraries for crypto variants. That creates cascading changes in a set of unrelated projects. It also creates recurring issues with visual studio build. I think that we should keep it as a single "core" project for now, and roll back the creation of separate crypto projects.

We should absolutely have separate files for each crypto implementation, and a code structure that allows the linker to only pick the object files required for build. But let's start with a small PR, whose only focus is the crypto API.

CMakeLists.txt
BitFis: I thought of statically linking the crypto implementation. This should help seperating the core implementation from the crypto backend implementation and prevent any acidental usage of direct crypto calls. But i would have kept all crypto implementation in the same library for now. What do you think?

BitFis:
> As I said in a few comments, I don't like the idea of creating separate libraries for crypto variants. That creates cascading changes in a set of unrelated projects. It also creates recurring issues with visual studio build. I think that we should keep it as a single "core" project for now, and roll back the creation of separate crypto projects.
> 
> We should absolutely have separate files for each crypto implementation, and a code structure that allows the linker to only pick the object files required for build. But let's start with a small PR, whose only focus is the crypto API.

Sounds good with me, I actually started to struggle with the vs project files. I will undo those change

picoquic/ptls_openssl_backend.c
```diff
@@ -0,0 +1,23 @@
...
+* (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+* LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+* ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+* (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+* SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+/**
+ * PicoTLS with openssl backend [WIP]
+ */
```
huitema, line 23: Fusion is interesting from an API design point of view. The code in Picotls only runs on x64 platforms that support the AES instruction set. The application could try to use #ifdef to only add if fusion if supported, but that's error prone. Picotls has a function "int ptls_fusion_is_supported_by_cpu(void)" that can be called at run-time, and that's much more reliable.

In practice, that function will have to run during the "setup" call to the API. If the function returns False, the AESGCM suites will have to be provided differently -- minicrypto, openssl, maybe bcrypt on Windows.

There are two choices for the crypto-API: either have setup return an error, in which case the application will have to try a different provider; or require the "fusion" provider to do the fallback internally. I think letting the application decide is better.

huitema:
@BitFis I just completed PR #1064, to negotiate the fusion implementation of AESGCM when available. The good news is that the fusion code works well.

huitema:
I have started PR #1066, to isolate the crypto ops at the top of tls_api.c. I think this will facilitate the design of the crypto API.
The current code resulting from that PR is not great. There are too many functions, reflecting the incremental development of the API calls over several years.

Some of the calls are redundant, and should be replaced by "core" routines orchestrating simple functions. Also, calling conventions should be enforced, in order to align the resulting API with the look and feel of picotls. But this is probably a simpler start for the crypto API development than the previous code.

BitFis:
I would leave the PR for now while I do changes as reference. But you are right, it makes more sence to break it down in smaller parts. So for finished small parts, I will make clean small PRs. But you can also close it you feel it is inconvenient.

huitema:
We need to continue this effort. I did the work on PR #1066 to better understand what was involved, and I think that I see it much better now. As you said, we need to proceed by small parts, for two reasons. First, big parts are harder and the risk of destabilizing the code is much higher. Second, we are learning as we proceed, and it is much easier to learn step by step.

One of the lesson I learned in this exercise is that the dependency on OpenSSL can be split in two issues: the crypto library proper, and the management of certificates. The crypto library part covers AEAD algorithms and Key Exchange suites, and also random number generation. I think that part is almost ready, and we should start there. We may need to simplify a bit after PR #1066, but we are very close.

Doing a crypto API will solve some issues like picking the best algorithms for a given platform, but will not completely solve the dependency on OpenSSL. Picoquic uses the OpenSSL library for certificate verification: the "certificate_verifier" function used in the client, and the "sign_certificate" function used in the server. These functions are only partially implemented in the Picotls library. Picotls does not support RSA, its internal implementation of "sign certificate" can only support ECDSA certificates. Picotls has not equivalent of the "verify_certificate" function, and the only practical alternative is to have a null verifier.

To fully remove the OpenSSL dependency, we need first to address gaps in Picotls. For example, Picotls does have a function for loading private keys (ptls_minicrypto_load_private_key), but that alternative does not yet handle the Elliptic Curve key format used by default in OpenSSL, forcing deployment to convert to the "standard" format. Picotls does not have a "certificate verify" function, and that would need to be developed. The loading of certificate lists would have to be tested. Some support for RSA would have to be added. All of that is doable, but requires work and time.

We may also want to develop alternatives to the "verify_certificate" approach, which is based on PKI. This is adequate for Web interactions, but other approaches might be needed if we want P2P or IOT applications: a "trust on first use" strategy similar to SSH for P2P, and the use of raw public keys instead of certificates for IOT. The API will need to handle that in the future.

But let's not just stop in front of this work. The code is ripe for the implementation of a basic crypto provider API. Let's start there.

BitFis:
Sadly I have to put my work on this on ice for now, since priorities changed. I will reopen this issue when I get back to it at work or during the weekends.

PR #1062, opened 2020-10-13
Propose usage of clang-format and editorconfig [WIP]

BitFis:
I would like to propose the usage of [clang-format](https://clang.llvm.org/docs/ClangFormat.html) and [editorconfig](https://editorconfig.org/).

This pull requests contains:
- [ ] `clang-format` in cicd
- [ ] format whole project
- [ ] check files to be `.editorconfig` correct

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,17 @@
+---
+# We'll use defaults from the LLVM style, but with 4 columns indentation.
+# BasedOnStyle: LLVM
+IndentWidth: 4
+---
+Language: Cpp
+# Force pointers to the type for C++.
+DerivePointerAlignment: false
+PointerAlignment: Left
+AlignAfterOpenBracket: AlwaysBreak
+AllowAllParametersOfDeclarationOnNextLine: true
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterFunction: true
+ColumnLimit: 110
+AllowShortFunctionsOnASingleLine: false
+SortIncludes: false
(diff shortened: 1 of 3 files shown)


huitema:
Asking @bkchr to take a look at this if he has time. Bastian implemented the "make clangformat" option in CMake, which I believe does something pretty close to what this PR suggests. My only problem is that the Visual Studio editor keeps fighting against the formatting, which is a bit tiring.

BitFis:
> Asking @bkchr to take a look at this if he has time. Bastian implemented the "make clangformat" option in CMake, which I believe does something pretty close to what this PR suggests. My only problem is that the Visual Studio editor keeps fighting against the formatting, which is a bit tiring.

Jep, saw that. Seems we use [`-style=Webkit`](https://webkit.org/code-style-guidelines/). I will change this pull request accordingly and would like to suggest adding  the clang-format command to the cicd to be checked and run, well, continuously. The `.editorconfig` contains configuration which can not be checked by the clang-format; whitespaces / new lines at the end of file etc. . This would be an extension. It is also supported by vs and most IDEs. so not much work for integration. I could split up into two pull requests if you are interested.

huitema:
You don't need to split in two, just ensure that the "make clangformat" is consistent with the editing rule. If you can do it in such a way that VS understand it, then that's great.

BitFis:
> You don't need to split in two, just ensure that the "make clangformat" is consistent with the editing rule. If you can do it in such a way that VS understand it, then that's great.

Wont get to it this week anymore, will remove WIP after completion

BitFis:
Sadly I have to put my work on this on ice for now, since priorities changed. I will reopen this issue when I get back to it at work or during the weekends.

PR #1072, opened 2020-10-26
Unified logging

### LINKED ISSUE #811: Unified logging?
There are now two logging systems: text file and binary file. The binary file can be processed to retrieve almost the same text as the text file, but in qlog format, which is unified and standard. The binary file also contains congestion events, in a format specific to picoquic. It would not take much to make the binary file is a complete superset of the text file. Mostly:

* Add '**datagram** events as specified in qlog
* Add qlog logging of transport parameters, SNI, ALPN
* Make sure that the qlog parser displays the frame contents with the same details as the text file
* Add a way to stick debug printf into the log.
* Find a rational way to log congestion events.

Unifying the logging will reduce the logging overhead. It will also help the next action, switch from an always increasing log to a rotating log.

huitema:
Close #811 

Make sure that different types of logs are defined by a table of functions, and thus allow applications to only link with the logging code that they use. Also ensure that the code itself only calls into a unified logging API.


huitema:
@steschu77 this is work in progress, inspired by your comments on the "autoqlog" design. The idea is to make all the logging code optional. The application has to call "picoquic_set_textlog" (defined in picoquic_logger.h) to enable text logging; if it doesn't, the text logging code will not be linked. Same for "picoquic_set_binlog".

picoquic/logger.c
```diff
@@ -1983,29 +1982,205 @@ void picoquic_txtlog_message_v(picoquic_quic_t* quic, const picoquic_connection_
...
+    textlog_cc_dump
+};
+
+int picoquic_set_textlog(picoquic_quic_t* quic, char const* textlog_file)
+{
+    int ret = 0;
+    FILE* F_log;
+
+    if (quic->F_log != NULL && quic->should_close_log) {
+        (void)picoquic_file_close(quic->F_log);
```
steschu77, line 2167: The idea was to use it as `quic->F_log = picoquic_file_close(quic->F_log);`

picoquic/picoquic_unified_log.h
```diff
@@ -0,0 +1,189 @@
...
+
+/* log the start of a connection */
+typedef void (*picoquic_log_new_connection_fn)(picoquic_cnx_t* cnx);
+/* log the end of a connection */
+typedef void (*picoquic_log_close_connection_fn)(picoquic_cnx_t* cnx);
+
+/* log congestion control parameters */
+typedef void (*picoquic_log_cc_dump_fn)(picoquic_cnx_t* cnx, uint64_t current_time);
+
+typedef struct st_picoquic_unified_login_t {
```
steschu77, line 106: Should probably read `st_picoquic_unified_logging_t` :)

steschu77 (approved):

picoquic/logger.c
huitema: Good catch. Fixing that.

picoquic/picoquic_unified_log.h
huitema: Yes! Fixing that too.

Issue #1088, opened 2020-11-08
Migration with zero-length CID on client


huitema:
This should be supported. Need to at a minimum add a test to verify that it works for both "client driven migration" and "server preferred address". This supposes that server has non-zero-length CID.

huitema:
Fixed in PR #1089

PR #1089, opened 2020-11-09
Zero cid migration

huitema:
Adding tests for migration with zero-length client CID.


huitema:
@martinthomson This PR adds three tests of migration, NAT rebinding and preferred address migration with the constraint that the client uses zero-length CID. The tests pass. Does that alleviate your concerns about these scenarios?

huitema:
Close #1088

martinthomson (approved):
More testing 👍 

It seems like you had close to the some bug as everyone else though.

huitema:
@martinthomson The only changed I made in the actual code was to allow the client to engage in migrations even if the client CID length was set to 0. The previous code would just consider setting "cid length=0" to imply "no migrations". With that, the tests verify that NAT rebinding, explicit migration and preferred address migration work. I also verified manually that the client modified with this PR can do NAT rebinding and explicit migration against the not-yet-modified QUIC server.

During your previous tests, were you testing the neqo client against the picoquic server, or the picoquic client against the neqo server? If there is an issue that I did not capture, I need to reproduce it now.

By the way, I saw an interesting issue when testing from home. The client's PATH CHALLENGE are fully padded. The ACKs sent on the old port prior to the challenge response are not. The network manages to send all the short packets before the long ones, leading to "far away" our of order delivery of the path challenge!

huitema:
Close #1088

martinthomson:
I didn't test picoquic.  I was using public services.  I can try to find time to test picoquic if you think that would help.

huitema:
Also fixed the migration code, to only migrate after receiving non-migration packets. The previous code was migrating immediately after receiving a challenge response. That could cause a "migration bounce" if packets were received out of order. The rule is only enforced on the server side, since the client intention is to migrate. It is not enforced for NAT rebinding, because that causes too many packet losses.

PR #1097, opened 2020-11-21
Fix printing of picoquic options

IvanNardi:


huitema:
Thanks for the fixes, @IvanNardi . I tested this on Windows, but not on Linux -- sorry about that. I suppose that I should add an unit test to make sure this stay fixed!

IvanNardi:
> Thanks for the fixes, @IvanNardi . I tested this on Windows, but not on Linux -- sorry about that. I suppose that I should add an unit test to make sure this stay fixed!

And I tested it on Linux only... :-)

PR #1098, opened 2020-11-22
Linux udp gso

huitema:
Enabling UDP GSO on Linux


huitema:
Using Reno as congestion control (-G Reno), the observed performance on Linux loopback on a VirtualBox VM was 980Mbps, significant improvement.

PR #1099, opened 2020-11-22
fix crash-on-exit bug in server

huitema:


huitema:
The crash on exit happened if the server was launched without specifying cert or key. In that case, the config was set to the default value... and freed on exit, because that default value was created on the stack.

Issue #8, opened 2017-10-06
Receive TLS resume session token


huitema:
Interop with Winquic failed because the TLS code choked on the incoming resume token.

Kazuho suggests looking at code in picotls key/cli.c for examples.

huitema:
Fixed. The Picotls code provides two API: ptls_handshake is only valid for handshake messages, and now picoquic properly calls ptls_receive for post handshake messages.

PR #1108, opened 2020-12-18
Mutipath option

huitema:
Work in progress.


huitema:
Close #8
Close #1110
Close #1111

PR #1113, opened 2021-01-02
picoquic_get_wake_delay: when to wake for a given cnx

tbarbette:
Allows to know when to wake-up for a particulalry given connection and
not the whole QUIC stack (as opposed to get_next_wake_delay). This is useful
for first party apps that wrap each cnx inside structures with their own
time management

### DIFF
--- a/picoquic/picoquic.h
+++ b/picoquic/picoquic.h
@@ -585,6 +585,9 @@ picoquic_cnx_t* picoquic_get_next_cnx(picoquic_cnx_t* cnx);
 int64_t picoquic_get_next_wake_delay(picoquic_quic_t* quic,
     uint64_t current_time,
     int64_t delay_max);
+int64_t picoquic_get_wake_delay(picoquic_cnx_t* cnx,
+    uint64_t current_time,
+    int64_t delay_max);
 picoquic_cnx_t* picoquic_get_earliest_cnx_to_wake(picoquic_quic_t* quic, uint64_t max_wake_time);
 
 uint64_t picoquic_get_next_wake_time(picoquic_quic_t* quic, uint64_t current_time);
--- a/picoquic/quicctx.c
+++ b/picoquic/quicctx.c
@@ -1057,6 +1057,33 @@ int64_t picoquic_get_next_wake_delay(picoquic_quic_t* quic,
     return wake_delay;
 }
 
+static uint64_t picoquic_get_wake_time(picoquic_cnx_t* cnx, uint64_t current_time)
+{
+    uint64_t wake_time = UINT64_MAX;
+
+    if (cnx->quic->pending_stateless_packet != NULL) {
+        wake_time = current_time;
+    } else {
+        wake_time = cnx->next_wake_time;
+    }
+
+    return wake_time;
+}
+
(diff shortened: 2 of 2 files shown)


huitema:
This is good and probably useful, but I am a bit concerned about the documentation. (Yes, I know, could be much better overall.) I just spent time helping a developer write a robust wait loop. Part of the issue was the confusion between access per connection and global wait. The robust structure for a global loop is:

* Call `picoquic_get_next_wake_delay` to get the next overall delay
* do select or equivalent to wait for that delay, or a packet arrival, or maybe an app event
* process the packet through `picoquic_incoming_packet`
* check whether packets are ready to send through `picoquic_prepare_next_packet_ex` or `picoquic_prepare_next_packet`, and if something is ready, send it.
* repeat.

Maybe we need some comments in the header explaining when to use the per connection API versus the global API.

tbarbette:
In my case, I couldn't use that because it's a client that has many connections (a load generator). Each client has its own timing and socket, also according to the rate.
The generator uses its own event loop, basically, I need to know when to wake every FD independently. Maybe the thing a little bit different here is that I send connections through different sockets?

huitema:
Sorry for trying to dig in the issue, but synchronization issues are interesting. The current APIs are not thread safe. I wonder whether a new set of thread safe APIs would simplify a number of use cases. I also wonder how exactly such API should be designed. If I was writing a load generator like yours, I would handle each client connection as a separate thread, with its own Quic context. The common context forces synchronization of multiple clients. In a multi-threaded environment it requires using some kind of semaphore before calling `picoquic_incoming_packet` or `picoquic_prepare_next_packet`, because otherwise queue pointers in the Quic context may be overwritten. What do you gain from having a single Quic context?

huitema:
Opened issue #1114 to track real time programming considerations.

tbarbette:
I use one "quic" object per thread. Performance-wise we can't have one connection per thread, it would be way too slow to context switch all the time.  So each thread (one per core) handles many connections. There is also a bit of legacy because I'm modifying the generator (WRK) that is not from me. If I was building it from scratch, I guess I could use the normal get_next_wake_delay. However, when waking up for a transmit event I would have to lookup again to know which one is the next one, with the proposed API I wake up for the right connection. I guess other big projects may have their own timing+select management that imposes to know the next wake up time for each connection.

My personal opinion is I wouldn't care about thread safeness. You'll impact performance with locks everywhere, while the most efficient solution for network-driven workload has been shown many times to be a sharding approach where each core is almost an independent server.
But for a simple multi-threaded client that is not so much about performance, it may be easier to have a thread-safe API from the developer's perspective... Still, I'm not sure a browser is really using multi-threaded in the handling of a single web page/single connection anyway.

Fyi now an 8-core 3Ghz Xeon can do around 10Gbps both for client and server (HTTP "0.9", serving the generated file /10000, pretty convenient btw). The biggest issue is I/O. We could use DPDK and/or AF_XDP to mostly solve that, but in our case, we just wanted to have some QUIC traffic to design a load-balancer.
The server simply launched one picoquicdemo server per core, using network namespaces to separate them. SRIOV can ensure packets go to the right core in hardware so it's quite efficient.

huitema:
@tbarbette can we have this discussion in issue #1114 ?

PR #1123, opened 2021-01-26
Add github action for CI tests

huitema:


.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          # We must fetch at least the immediate parents so that if this is
+          # a pull request then we can checkout the head.
+          fetch-depth: 2
+          submodules: 'recursive'
+
+      - run: |
+          sudo apt-get install -y libssl-dev
```
IvanNardi, line 22: Not familiar with GitHub Action, but at very least you should install gdb too:
sudo apt-get install -y libssl-dev gdb

huitema:
@larseggert could you take a look at the CITests github action? As designed, it seems to run as expected on Ubuntu, but it might be useful to also run it on IOS. Any suggestion for that?

PR #1130, opened 2021-02-07
picoquicdemo: fix support for 0 length cids

IvanNardi:


IvanNardi:
Not sure if the proposed PR is the best solution for the issue (because it is a little invasive), but we need to be able to tell "we don' care about cid lengths; use the default values" from "use 0 length cid". Right now the value 0 is used for both situation, AFAIK

huitema (approved):
Looks fine. I read your comment about the change being invasive, but it looks minimal. You introduce the `picoquic_config_init` API, which is reasonable, and then the changes just follow.

PR #1133, opened 2021-02-09
picoquicdemo: fix multipath option, i.e. "-M"

IvanNardi:
Use the same logic already used by loss-bit" configuration.

### DIFF
--- a/picoquic/picoquic_internal.h
+++ b/picoquic/picoquic_internal.h
@@ -510,6 +510,7 @@ typedef struct st_picoquic_quic_t {
     uint32_t sequence_hole_pseudo_period; /* Optimistic ack defense */
     picoquic_spinbit_version_enum default_spin_policy;
     picoquic_lossbit_version_enum default_lossbit_policy;
+    uint32_t default_multipath_option;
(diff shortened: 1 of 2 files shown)


IvanNardi:
Is multipath already supported in picoquicdemo?
I am not able to trigger any "multipath behavior", i.e. simply using "-M" option I don't  see the new frame types; am I missing something?
Anyway, with this patch, client and server send the new TPs, at least

huitema:
The multipath option is supported, but only in a passive way: if the peer creates a new path, the demo code will use it. That's OK on the server side. On the client side, we are missing the code to add a new path. 

For testing migration, we open a new path by creating a new socket with a new port. That's not ideal for testing multipath because just changing the port means the two "path" will be routed exactly the same way by the network. We would need another option to add a path and specify the source IP address -- but for that the client needs to be multihomed.

huitema (approved):
It is OK to check that now, or we may want to add he code to actually open a new path on the client.

huitema:
I opened issue #1134 to track the demo client issue. Will address that in a separate PR.

PR #1135, opened 2021-02-12
Implement the quicperf protocol in picoquicdemo

### LINKED ISSUE #1126: Implement the perf protocol
The simple perf protocol is defined in https://datatracker.ietf.org/doc/draft-banks-quic-performance/?include_text=1.

Summary:
* ALPN is "perf"
* Runs on port 443
* Client driven.
* Client opens streams in which the first 8 bytes encode the size of the desired response (network order).
* If client opens bidir stream, server replies on bidir stream.
* If client opens unidir stream, server replies on next unidir stream.
* If stream is less than 8 bytes, or required length is zero, no response needed. No unidir stream needed.
* If length is set to 0xFFFFFFFFFFFFFFFF (almost infinite stream), client will close the stream eventually using reset stream.

huitema:
Close #1126

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -96,6 +96,7 @@ set(PICOHTTP_LIBRARY_FILES
     picohttp/democlient.c
     picohttp/demoserver.c
     picohttp/h3zero.c
+	picohttp/quicperf.c
 	picohttp/siduck.c
 )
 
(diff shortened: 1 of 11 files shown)


CMakeLists.txt
```diff
@@ -96,6 +96,7 @@ set(PICOHTTP_LIBRARY_FILES
     picohttp/democlient.c
     picohttp/demoserver.c
     picohttp/h3zero.c
+	picohttp/quicperf.c
```
alagoutte, line 99: indent ?

huitema: Tab/space. Sorry. Fixed.

PR #1136, opened 2021-02-13
Fix local computation of ack gap

huitema:
When doing performance testing between msquic client and picoquic server, we found that the throughput was about 500 Mbps, while we observed more than 2 Gbps between picoquic client and picoquic server on the same platform. Investigation shows that the picoquic client was sending way to many acknowledgements, which diminished the performance during high speed tests. This PR fixes that.

### DIFF
--- a/picoquic/config.c
+++ b/picoquic/config.c
@@ -85,6 +85,7 @@ static option_table_line_t option_table[] = {
     "send a large client hello in order to test post quantum readiness" },
     { picoquic_option_Ticket_File_Name, 'T', "ticket_file", 1, "file", "File storing the session tickets" },
     { picoquic_option_Token_File_Name, 'N', "token_file", 1, "file", "File storing the new tokens" },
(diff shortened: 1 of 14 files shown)


huitema:
@alagoutte this change causes the program to set the socket buffer sizes (SO_RCVBUF, SO_SNDBUF) to 640KB, for about 20% performance gain. The "-B" argument can be used to reverse that and use the default size for the socket buffers. I like the idea of best performance by default, but I can imagine different opinions. Do you think that's OK?

alagoutte:
> 
> 
> @alagoutte this change causes the program to set the socket buffer sizes (SO_RCVBUF, SO_SNDBUF) to 640KB, for about 20% performance gain. The "-B" argument can be used to reverse that and use the default size for the socket buffers. I like the idea of best performance by default, but I can imagine different opinions. Do you think that's OK?

Bonjour,

I think there is an error, it is not me ;-)

PR #1137, opened 2021-02-16
Add support for asan/ubsan build

IvanNardi:
Usage: "cmake -DENABLE_ASAN=ON -DENABLE_UBSAN=on ."

Tested on Ubuntu 20.04, x86_64 (cmake 3.16.3)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,12 +3,45 @@ cmake_policy(SET CMP0003 NEW)
 project(picoquic C CXX)
 find_package (Threads REQUIRED)
 
+option(ENABLE_ASAN "Enable AddressSanitizer (ASAN) for debugging" OFF)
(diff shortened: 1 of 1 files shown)


huitema:
That's great. How hard would it be to add a github action similar to `.github\workflows\ci-test.html` so the compile checks are performed for every push request?

IvanNardi:
> That's great. How hard would it be to add a github action similar to `.github\workflows\ci-test.html` so the compile checks are performed for every push request?

It shouldn't be too hard. Though, right now picoquic_ct triggers a lot of memory leaks:

``
SUMMARY: AddressSanitizer: 12929584 byte(s) leaked in 2330 allocation(s).
``

If we enable asan compilation on GitHub Action now, every future PR will fail until all these leaks are fixed (AFAIK)

huitema:
I opened issue #1138 so we do not forget the github action.

PR #1139, opened 2021-02-17
github action to  check memory leak

huitema:
Close #1138

PR #1140, opened 2021-02-17
Fix some memory leaks

IvanNardi:


IvanNardi:
Some fixes seem pretty easy, but other ones are not so obvious...

picoquictest/tls_api_test.c
```diff
@@ -7155,6 +7155,9 @@ int bad_cnxid_test()
             ret = -1;
         }
         else {
+            for (size_t i = 0; i < test_ctx->nb_test_streams; i++) {
+                test_api_delete_test_stream(&test_ctx->test_stream[i]);
+            }
```
IvanNardi, line 7160: These frees are needed, but they seems out of place here...

picoquictest/tls_api_test.c
```diff
@@ -8616,6 +8619,9 @@ int ddos_amplification_test_one(int use_0rtt)
             picoquic_delete_cnx(test_ctx->cnx_server);
             /* Delete the old client connection */
             picoquic_delete_cnx(test_ctx->cnx_client);
+            for (size_t i = 0; i < test_ctx->nb_test_streams; i++) {
+                 test_api_delete_test_stream(&test_ctx->test_stream[i]);
+            }
```
IvanNardi, line 8624: Same as above

picoquictest/tls_api_test.c
huitema: Yes. They belong in the deletion of the test context.

picoquictest/tls_api_test.c
huitema: Yes. And that's hiding the missing code in the test context deletion.

huitema (commented):
Looks good. I was going though some of that in PR #1139. I will check you PR first.

PR #1141, opened 2021-02-18
Fix some memory leaks

IvanNardi:
These fixes are not present in #1139.

### DIFF
--- a/picoquic/config.c
+++ b/picoquic/config.c
@@ -818,6 +818,10 @@ void picoquic_config_clear(picoquic_quic_config_t* config)
     {
         free((void*)config->server_key_file);
     }
+    if (config->esni_key_file != NULL)
+    {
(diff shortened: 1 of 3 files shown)


huitema:
Thanks for the fixes. In 1139, I am going after the remaining leaks one by one. I wrote a script that loops over all the tests in `picoquic_ct` and stops if there is a leak. Then, I go debug that test. I found mostly two kinds of issues: some tests were coded quickly and forgot to free memory used to set up the test; and, some tests exercise rarely used parts of the code in which there was a leak. Debugging that second kind of issue takes some time.

huitema:
A minor issue: "show all checks" only shows the results of the github actions for the pull requests that I create, but does not show them to the reviewers. That's a bit annoying. Something to fix later.

huitema:
The remaining leaks are found during the fuzz test.

PR #1148, opened 2021-02-23
Ease detecting of UBSAN Runtime errors.

IvanNardi:
If such an error is found, print a verbose error report and exit the program.
This behavior might be useful in GitHub Action scripts


huitema:
Thanks for the PR!

PR #1149, opened 2021-02-23
CI tests on pull request

huitema:


huitema:
@IvanNardi Running the tests on pull requests makes the test results visible to reviewers. I also fixed the ubsan-asan test to return an error if picoquic_ct fails, and thus benefit from PR #1148

PR #1153, opened 2021-02-26
picoquic: fix frame comparison

IvanNardi:
Frame types are not only 1 byte long anymore

Warning:
/home/ivan/svnrepos/picoquic/picoquic/frames.c:2640:41: error: result of
comparison of constant 'picoquic_frame_type_ack_mp' (764832) with
expression of type 'uint8_t' (aka 'unsigned char') is always false
[-Werror,-Wtautological-constant-out-of-range-compare]
        } else if (p->bytes[byte_index] == picoquic_frame_type_ack_mp) {
                   ~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~

### DIFF
--- a/picoquictest/multipath_qlog_ref.txt
+++ b/picoquictest/multipath_qlog_ref.txt
@@ -193,7 +193,7 @@
 [67657, 0, "transport", "datagram_received", { "byte_length": 55, "addr_from" : {"ip_v4": "10.0.0.2", "port_v4":1234}}],
 [67657, 0, "transport", "packet_received", { "packet_type": "1RTT", "header": { "packet_size": 55, "packet_number": 5, "dcid": "0a09080706050403" }, "frames": [{ 
     "frame_type": "time_stamp", "time_stamp": 8576}, { 
-    "frame_type": "ack_mp", "path_id": 0, "ack_delay": 0, "acked_ranges": [[0, 8]]}, { 
+    "frame_type": "ack_mp", "path_id": 0, "ack_delay": 0, "acked_ranges": [[2, 8]]}, { 
     "frame_type": "ack_mp", "path_id": 1, "ack_delay": 0, "acked_ranges": [[0, 0]]}, { 
     "frame_type": "padding"}]}],
 [67657, 0, "recovery", "metrics_updated", {"pacing_rate": 7084870,"bytes_in_flight": 4555,"smoothed_rtt": 21680,"min_rtt": 18078,"latest_rtt": 23560}],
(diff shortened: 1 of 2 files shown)


IvanNardi:
While this patch seems fine to me, it breaks multipath_qlog test.
No sure if the culprit is the patch iself or something in the test...

huitema (changes requested):
This is a good fix, but we need to fix the test `multipath_qlog` before checking it in.

The code change fixes the handling of ACK_MP frames. This is good, but it will automatically changes the flow of packets in the test connection logged in the `multipath_qlog` frame. The normal steps there are:

1) Run the `multipath_qlog` test. It will fail because the log that it produces, `multipath_qlog_test.qlog`, is different from the expected value in `picoquictest/multipath_qlog_ref.txt`.

2) Verify that the changes in the log are what we expect. The first change ought to be on a frame containing an ACK_MP. Due to improved ack-of-ack processing, the ACK_MP frame ought to carry fewer ranges or shorter ranges than the reference version.

3) If the changes are good, checkout `picoquictest/multipath_qlog_ref.txt` and replace its content by the value of `multipath_qlog_test.qlog`.

4) Run the test again. It should succeed. If it does, commit the change in your PR branch.

picoquictest/multipath_qlog_ref.txt
```diff
@@ -208,7 +208,7 @@
 [70893, 0, "transport", "spin_bit_updated", { "state": true }],
 [70893, 0, "transport", "packet_sent", { "packet_type": "1RTT", "header": { "packet_size": 1424, "packet_number": 14, "dcid": "0908070605040302" }, "frames": [{ 
     "frame_type": "time_stamp", "time_stamp": 8861}, { 
-    "frame_type": "ack_mp", "path_id": 0, "ack_delay": 0, "acked_ranges": [[0, 6]]}, { 
+    "frame_type": "ack_mp", "path_id": 0, "ack_delay": 0, "acked_ranges": [[4, 6]]}, { 
     "frame_type": "ack_mp", "path_id": 1, "ack_delay": 0, "acked_ranges": [[0, 2]]}, { 
     "frame_type": "stream", "id": 8, "offset": 0, "length": 1390, "fin": false , "has_length": false, "begins_with": "0001020304050607"}]}],
```
huitema, line 213: Nice. That's exactly the changes I would have expected. It shows that the code fix is having the intended effect.

huitema (approved):

IvanNardi:
Thanks for your suggestion! I confirm that ack ranges are smaller in ACK_MP frame now

huitema:
Thanks for finding and fixing this bug!

Issue #1155, opened 2021-02-27
Runtime error on migration tests


IvanNardi:
Pretty much every \*migration\* test trigger the same Runtime Error.

For example: 

```
Starting test number 135, migration
quictest/tls_api_test.c:1562 [tls_api_synch_to_empty_loop]: Exit synch loop after 49 rounds, backlog or not enough paths (1 & 1).

quic/picoquic/quicctx.c:1504 [picoquic_find_path_by_address]: Could not find path
/home/ivan/svnrepos/picoquic/picoquic/sender.c:3726:66: runtime error: member access within null pointer of type 'picoquic_local_cnxid_t' (aka 'struct st_picoquic_local_cnxid_t')
    #0 0x61b359 in picoquic_prepare_packet_ready /home/ivan/svnrepos/picoquic/picoquic/sender.c:3726:66
    #1 0x61bdcc in picoquic_prepare_segment /home/ivan/svnrepos/picoquic/picoquic/sender.c
    #2 0x61dbfd in picoquic_prepare_packet_ex /home/ivan/svnrepos/picoquic/picoquic/sender.c:4179:27
    #3 0x529e71 in tls_api_one_sim_round /home/ivan/svnrepos/picoquic/picoquictest/tls_api_test.c:1301:23
    #4 0x52c20c in tls_api_data_sending_loop /home/ivan/svnrepos/picoquic/picoquictest/tls_api_test.c:1504:15
    #5 0x54aa21 in migration_test_scenario /home/ivan/svnrepos/picoquic/picoquictest/tls_api_test.c:5378:15
    #6 0x4e36b7 in do_one_test /home/ivan/svnrepos/picoquic/picoquic_t/picoquic_t.c:301:15
    #7 0x4e25fe in main /home/ivan/svnrepos/picoquic/picoquic_t/picoquic_t.c:560:29
    #8 0x7f5e0f6b20b2 in __libc_start_main /build/glibc-eX1tMB/glibc-2.31/csu/../csu/libc-start.c:308:16
    #9 0x434b8d in _start (/home/ivan/svnrepos/picoquic/picoquic_ct+0x434b8d)

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/ivan/svnrepos/picoquic/picoquic/sender.c:3726:66 in
```

huitema:
This was fixed by PR #1158.

huitema:
[...]

PR #1158, opened 2021-02-28
Fix references to p_local_cnxid

huitema:
@IvanNardi this is a fix for issue #1155. The code verifier complains about dereferencing a NULL reference to p_local_cnxid. That reference is not actually needed in the specific code path in which the error is detected. The PR rewrites the calls to the packet protection and header formatting functions, so that p_local_cnxid is only deferenced when formatting long header packets (initial, handshake, 0-RTT) and not when formatting 1-RTT packets. The code guarantees that path[0]->p_local_cnxid will always be a valid reference during the handshake.


IvanNardi (approved):
LGTM, thank you very much!

PR #1165, opened 2021-03-05
Allow build on FreeBSD

adventureloop:


huitema:
Thanks for the fix. One question: we use to have CI runs testing the FreeBSD (and IOS) ports. we lost that when switching from Travis to GitHub Actions. Any suggestion for running at least one of the tests on FreeBSD in a GitHub Action?

PR #1167, opened 2021-03-06
GitHub action macos latest

huitema:


huitema:
@adventureloop this PR adds a PI action to verify that picoquic builds and runs properly in MacOS. I would like to add something similar for FreeBSD, but I am not sure how to do that with GitHub Actions.

adventureloop:
It seems the path to running FreeBSD ci from a github action requires using a mac os image and spinning up a vm

see:
https://github.com/actions/runner/issues/385

huitema:
The "macos-latest" action should largely verify that picoquic works fine under FreeBSD. If it doesn't, then we should open specific issues and fix them.

Issue #1012, opened 2020-08-17
Prevent Request forgery attacks


huitema:
See QUIC issue https://github.com/quicwg/base-drafts/issues/3995.

Should be some heuristics about what migrations are OK and what are not, with particular for Server Preferred Address, but also client driven redirect:

* Consider proximity levels as remote / local / host, avoid redirects that lower the level.
* Consider not redirecting towards unexpected port numbers?

huitema:
Draft 30 suggests specific fixes:

1) Do not redirect to loopback (127.0.0.1 or ::1) if the current server address is not loopback.
2) Limit use of "NEW TOKEN" to same address server

huitema:
Protection against request forgery should also consider version negotiation, see discussion in security section of QUIC invariants.

huitema:
The final text has two level of suggestions:
1) protection against migration to loopback, which should probably be on by default;
2) protection against migration to other style of addresses, such as link local or NAT, which is more problematic.

This suggests some gradual protection, based on the potential drawback:

Protection | level | checks
----------- | ----- | -----
no restriction | 0 | do not apply any protection
loopback | 1 | very little risk of site effect
IPv6 link local | 2 | not much risk, but may be annoying
IPv4 NAT | 3 | limit migrations to that same NAT, risk breakage if multiple levels of NAT
IPv6 unique local | 4 | Do not migrate to a unique local number, risk breakage if multiple prefixes are used

The increasing levels of protection make it harder to justify. For example, the "IPv4 NAT" protection is based on the idea that such addresses may not be accessible from outside a firewall. But if there are several levels of NAT and routers, this is not clear. Similarly, there may be situations in which an enterprise deploys a set of different unique local prefixes. It seems that anything that relies on specific sets of addresses is better dealt with by firewalls. So the only levels that appear practical are 0, 1 and 2.

huitema:
Should we have different policies at servers and clients? Servers can direct clients to issue a probe using the "server preferred address" mechanism. An antagonistic server could use that to target devices on the local network of the client. Would that justify protection against IPv4 NAT sort of attacks? What would exactly be the gain?

huitema:
There was a suggestion on the WG Chat to apply special treatment for some "well known" port numbers, preventing reflection attacks against, for example, NTP servers. Need to retrieve the list that they used.

huitema:
The blocking protection applies to most "initial" attacks. There are however two holes in these protections:

[...]

Issue #956, opened 2020-06-22
Avoid reset blowback under stress


huitema:
The stateless reset should be rate limited. It should probably not be sent when the server is under DOS attack. In any case, the reset should only be sent in response to valid connection ID. If the connection ID is encrypted, it should be possible to test its validity in a stateless manner.

huitema:
The simplest method by far is to just enforce a min delay between two consecutive reset packets. We can think of something more complicated later.

PR #1168, opened 2021-03-07
Fix parsing of coalesced packets

IvanNardi:
Is very unlikely that a coalesced packet with an unexpected version
is a valid QUIC packet. It is probably some kind of unencrypted random
padding data added after a previously valid QUIC packet.

Note that we don't perform any checks on CIDs presence/values for any
packets with an unsupported version; if the coalesced packet is very
short , we might read invalid data when preparing the VN reply.

### DIFF
--- a/picoquic/packet.c
+++ b/picoquic/packet.c
@@ -2015,7 +2015,10 @@ int picoquic_incoming_segment(
     else {
         if (ret == 0 && picoquic_compare_connection_id(previous_dest_id, &ph.dest_cnx_id) != 0) {
             ret = PICOQUIC_ERROR_CNXID_SEGMENT;
-        }
+        } else if (ret == PICOQUIC_ERROR_VERSION_NOT_SUPPORTED) {
+            /* A coalesced packet with unknown version is likely some kind of padding */
+            ret = PICOQUIC_ERROR_CNXID_SEGMENT;
+	}
     }
 
     /* Store packet if received in advance of encryption keys */


IvanNardi:
Original error found with `./picoquic_ct -f 1000`, with a slightly modified code where some debug messages were been added:

```
oquic/picoquic/packet.c:60 [picoquic_parse_long_packet_header]: Version is not recognized: 0xab678586

picoquic_incoming_segment:2056 packet_length 1263 length 11
=================================================================
==46581==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x61c004d56798 at pc 0x7fc812ad0480 bp 0x7ffcd3fbeb30 sp 0x7ffcd3fbe2d8
READ of size 190 at 0x61c004d56798 thread T0
    #0 0x7fc812ad047f in __interceptor_memcpy ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:790
    #1 0x55a31a76b29f in memcpy /usr/include/x86_64-linux-gnu/bits/string_fortified.h:34
    #2 0x55a31a76b29f in picoquic_prepare_version_negotiation /home/ivan/svnrepos/picoquic/picoquic/packet.c:876
    #3 0x55a31a774633 in picoquic_incoming_segment /home/ivan/svnrepos/picoquic/picoquic/packet.c:2059
    #4 0x55a31a7763f6 in picoquic_incoming_packet /home/ivan/svnrepos/picoquic/picoquic/packet.c:2252
    #5 0x55a31a6aa73e in stress_handle_packet_arrival /home/ivan/svnrepos/picoquic/picoquictest/stresstest.c:541
    #6 0x55a31a6af447 in stress_loop_poll_context /home/ivan/svnrepos/picoquic/picoquictest/stresstest.c:796
    #7 0x55a31a6af447 in stress_or_fuzz_test /home/ivan/svnrepos/picoquic/picoquictest/stresstest.c:1030
    #8 0x55a31a6b0388 in fuzz_test /home/ivan/svnrepos/picoquic/picoquictest/stresstest.c:1166
    #9 0x55a31a66904e in do_one_test /home/ivan/svnrepos/picoquic/picoquic_t/picoquic_t.c:303
    #10 0x55a31a6663e9 in main /home/ivan/svnrepos/picoquic/picoquic_t/picoquic_t.c:538
    #11 0x7fc811c040b2 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x270b2)
    #12 0x55a31a668e9d in _start (/home/ivan/svnrepos/picoquic/picoquic_ct+0x1e3e9d)

0x61c004d56798 is located 0 bytes to the right of 1816-byte region [0x61c004d56080,0x61c004d56798)
allocated by thread T0 here:
    #0 0x7fc812b42bc8 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cc:144
    #1 0x55a31a7bdcf1 in picoquictest_sim_link_create_packet /home/ivan/svnrepos/picoquic/picoquic/sim_link.c:83
    #2 0x55a31a6aa7fd in stress_handle_packet_prepare /home/ivan/svnrepos/picoquic/picoquictest/stresstest.c:560
    #3 0x55a31a6af316 in stress_loop_poll_context /home/ivan/svnrepos/picoquic/picoquictest/stresstest.c:808
    #4 0x55a31a6af316 in stress_or_fuzz_test /home/ivan/svnrepos/picoquic/picoquictest/stresstest.c:1030
    #5 0x55a31a6b0388 in fuzz_test /home/ivan/svnrepos/picoquic/picoquictest/stresstest.c:1166
    #6 0x55a31a66904e in do_one_test /home/ivan/svnrepos/picoquic/picoquic_t/picoquic_t.c:303
    #7 0x55a31a6663e9 in main /home/ivan/svnrepos/picoquic/picoquic_t/picoquic_t.c:538
    #8 0x7fc811c040b2 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x270b2)
    #9 0x55a31a668e9d in _start (/home/ivan/svnrepos/picoquic/picoquic_ct+0x1e3e9d)

```

picoquic/packet.c
```diff
@@ -2015,7 +2015,10 @@ int picoquic_incoming_segment(
     else {
         if (ret == 0 && picoquic_compare_connection_id(previous_dest_id, &ph.dest_cnx_id) != 0) {
             ret = PICOQUIC_ERROR_CNXID_SEGMENT;
-        }
+        } else if (ret == PICOQUIC_ERROR_VERSION_NOT_SUPPORTED) {
+            /* A coalesced packet with unknown version is likely some kind of padding */
+            ret = PICOQUIC_ERROR_CNXID_SEGMENT;
```
IvanNardi, line 2020: Is this the right return code?

huitema: The two tests are likely to detect pretty much the same root cause, so reusing the same error code makes sense.

huitema (commented):
Good fix. How would we add a unit test for that?

huitema (changes requested):
I think we need a bigger fix. If you look at the code in `picoquic_parse_long_packet_header`, the VN verification happens very early, and there is no verification that the packet is otherwise valid. The code in `picoquic_prepare_version_negotiation` parses the incoming packet to find the dcid, but does so without checking the segment length. If the segment length is just short enough and the DCID length is just long enough (up to 256 bytes is legal), then the code will access unitialized memory -- what ASAN has detected. We need to fix that too.

huitema:
I think I understand what happened. The version negotiation code was designed before packet coalescing was added to the spec. It verifies that the packet is big enough. The max size of a type + version + CID is 519 bytes, much less than the 1200 enforced size. But coalescing breaks that assumption. The packet might be 1200 bytes, but the segment could be much less.

IvanNardi:
> I think we need a bigger fix. If you look at the code in `picoquic_parse_long_packet_header`, the VN verification happens very early, and there is no verification that the packet is otherwise valid. The code in `picoquic_prepare_version_negotiation` parses the incoming packet to find the dcid, but does so without checking the segment length. If the segment length is just short enough and the DCID length is just long enough (up to 256 bytes is legal), then the code will access unitialized memory -- what ASAN has detected. We need to fix that too.

That's exactly what I meant with" `if the coalesced packet is very short , we might read invalid data when preparing the VN reply.`

I am no sure we need further checks.
If I am reading correctly the code, we prepare the VN reply only if the UDP payload is longer than `PICOQUIC_ENFORCED_INITIAL_MTU`(1200) bytes.
We have the following cases:
* if the packet with the unknown version is the first (or the only) coalesced packet there is always enough data in the initial message for the CIDs
* if the packet with the unknown version is the second/third/... in the UDP payload with my patch we treat it as padding and we don't prepare the VN anymore

Am I missing somethnig?

huitema (approved):
You are right. There are changes needed in the version negotiation code, but they are unrelated to this issue: main change is that connection ID are not limited to 20 bytes for the "unknown" versions -- the limit is 255 bytes. These changes require fixing issues #1012 and perhaps also #956. But restricting VN negotiation to the first segment in a packet will make the code safe, and that's good enough.

huitema:
@IvanNardi by the way, I really appreciate the combination of fuzzing and asan. That's a very powerful way to find potential issues before some hacker forces us to fix some security bug! I need to apply that to the http code as well, maybe using the `http_corrupt` test.

PR #1171, opened 2021-03-08
Add test of coalescing random bytes to initial packet

huitema:


huitema:
@IvanNardi this adds a formal test of packet coalescing with random bytes, so we don't rely so much on the fuzz test. Also, one can manually verify that the logs are correct.

IvanNardi:
> @IvanNardi this adds a formal test of packet coalescing with random bytes, so we don't rely so much on the fuzz test. Also, one can manually verify that the logs are correct.

Thanks for adding some kind of test!

PR #1173, opened 2021-03-10
Delay file opening on server and close quickly

### LINKED ISSUE #1170: Crash during interop runner multiplex test
marten-seemann  1:23 AM
@huitema Looks like the multiplexing test is killing your implementation: https://interop.seemann.io/logs/2021-03-08T00:05/picoquic_quic-go/multiplexing/output.txt

huitema  7:13 AM
Do you have a stack trace (gdb, bt) by any chance?

huitema  7:29 AM
Looking at the traces to see "what is different". I see that your client is repeating a bunch of stream frames containing get commands that were already sent by the server. I did some aggressive clean-up of used memory recently, and maybe that was too aggressive... Will check that.

huitema:
The cause of the crash was "too many open files". Files were opened as soon as the request was received, but were served in order. By piling a large number of file opening, the client could cause the server to exceed the available number of file descriptors, which triggered a buggy code path. That path has been fixed too.

Close #1170

### DIFF
--- a/picohttp/democlient.c
+++ b/picohttp/democlient.c
@@ -485,11 +485,12 @@ static int picoquic_demo_client_open_stream(picoquic_cnx_t* cnx,
     return ret;
 }
 
-static int picoquic_demo_client_close_stream(
+static int picoquic_demo_client_close_stream(picoquic_cnx_t * cnx,
     picoquic_demo_callback_ctx_t* ctx, picoquic_demo_client_stream_ctx_t* stream_ctx)
 {
     int ret = 0;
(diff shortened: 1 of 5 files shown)


huitema:
The repro involved adding tests in which a lot of packets were losts. Analysis showed that this caused a large number of file descriptors to stay open, because more command were accepted than served. The file descriptors were opened when the command was parsed, so that process could cause more than 500 FD to be opened. After that, crashed could occur due to poor management of errors.

The big fix is to delay file opening until starting to transmit the stream. The client was doing that, but not the server. When the server does that, the number of simultaneous open remains low, and the error does not occur.

PR #1179, opened 2021-03-22
Add performance log

huitema:
The performance log writes data to a CSV file after the end of the current connections. It is meant to help tracking and  debugging performance issues.


huitema:
Feedback from end-to-end test: need ta add min-rtt, srtt, cwin. For cwin, may want to add max-cwin, because cwin may drop at the end of the connection.

PR #1183, opened 2021-03-24
Add tests of certificate verification

huitema:


huitema:
@victorstewart Can you take a look? We probably need more test scenarios.

victorstewart:
since picotls has checks of the certification verification and signature code paths, in theory i think all that's required is to make sure the picoquic paths are tested.

picoquic is agnostic to whether the certificate uses RSA or ECC, but we might as well add both. also missing a `picoquic_set_client_authentication(engine, 1)` path. and the custom certificate verification path... pending resolution on the picotls side.

otherwise the tests you added cover all the input vectors that come to mind at the moment.

i attached a zip file of ed25519 certs keys and a root chain. (though we need to wait until picotls is updated to run a test on them).

[Uploading ed25519.tls.zip…]()

huitema:
OK. I am going to save the 25519 certificates for now, and do an additional PR when the new version of picotls is available. I notice that you provided client certs. Should we add client certificate verification tests? We do have a test like that, `request_client_authentication_test`. But it is only run on windows. First step may be to also run it on Linux builds. Then I can think of a version that exercises 25519.

I see two version of keys and certs in you zip, and ed25519 folder, and a copy in a MacOS folder. Are these different?

I do not see a specific "root CA" file. What should I use?

victorstewart:
the `chain.cert.pem` file has an intermediate CA on top and a root CA on bottom, 2 in one! so load the chain as the root. 

haha maybe the files unpacked weirdly i zipped them on my mac os laptop (github wouldn't let me attach files with random extensions). there should only be 1 copy of each file by each name. i can just email you the files if that's easier.

yes i think we should add a client authentication test. I'll be using that mutual authentication pattern always. it already works on linux i've been using it.

huitema:
Just added the client auth tests to picotls_ct -- the goal here is to make sure we have end to end verification, so for example we detect potential issues in file formats, usage of API, or maybe memory leaks :-). I saw the "chain" file. I do not think I can use it as is, but I will try.

victorstewart:
the chain file will work. these are the certs and keys i've been using with picoquic already and my handshakes complete fine.

huitema:
We have to thank @bkchr  for the work on client authentication.

victorstewart:
oh but right none of these certs keys or the chain will work until picoquic is using the latest picotls version, since all of them are ed25519.

huitema:
OK, checking in this one. My next step is to rework the way certificate verifiers are plugged in picoquic. The current method is a bit hacky.

PR #1187, opened 2021-03-27
Fix discrepancy between H3 HQ and Quicperf

huitema:
Better receive loop on winsock.

Decrease ack gap during slow start to try reduce losses. 
Not ready yet, chaisng bizarre memory errors


huitema:
This brings in two important fixes:

1) In the windows and linux clients, exhaust the read queue before attempting to send data.
2) In the engine, track the state of congestion control and use smaller ack gaps if doing slow start.

PR #1197, opened 2021-04-04
Expose Connection Limits

victorstewart:
Given that ~5% of QUIC connections on the Internet still get blocked by middleware, it is still necessary to have a TCP fallback. Thus in order to enforce a unified application wide connection limit over TCP **and** QUIC connections, these connection limits need to be fluid. 

An alternative design might be to "listen" for new QUIC connections on one port, and then transition to a different port via Server Preferred Address. But unfortunately that is not an option here since using any port besides 443 sometimes also results in traffic being dropped or throttled by middleware.

So the new behavior would be... setting `max_nb_connections` during `picoquic_create` would initialize data structures to the maximum possible number of quic connections, and then `picoquic_adjust_max_connections` allows this maximal limit to be titrated up or down as the server open and closes backup TCP connections with clients unable to connect over UDP.

### DIFF
--- a/picoquic/picoquic.h
+++ b/picoquic/picoquic.h
@@ -410,6 +410,12 @@ int picoquic_esni_load_key(picoquic_quic_t * quic, char const * esni_key_file_na
 /* Set the ESNI RR. Must be called after setting the ESNI key at least once. */
 int picoquic_esni_server_setup(picoquic_quic_t * quic, char const * esni_rr_file_name);
 
+/* Adjust maximum connections allowed */
+void picoquic_adjust_max_connections(picoquic_quic_t * quic, uint32_t max_nb_connections);
+
+/* Get number of open connections */
+uint32_t picoquic_current_number_connections(picoquic_quic_t * quic);
+
 /* Obtain the reasons why a connection was closed */
 void picoquic_get_close_reasons(picoquic_cnx_t* cnx, uint64_t* local_reason,
     uint64_t* remote_reason, uint64_t* local_application_reason,
--- a/picoquic/picoquic.h
+++ b/picoquic/picoquic.h
@@ -410,8 +410,11 @@ int picoquic_esni_load_key(picoquic_quic_t * quic, char const * esni_key_file_na
 /* Set the ESNI RR. Must be called after setting the ESNI key at least once. */
 int picoquic_esni_server_setup(picoquic_quic_t * quic, char const * esni_rr_file_name);
 
-/* Adjust maximum connections allowed */
-void picoquic_adjust_max_connections(picoquic_quic_t * quic, uint32_t max_nb_connections);
+/* Adjust maximum connections allowed to the specified value.
+ * The maximum number cannot be set to a value higher than the limit set when the context was
+ * created. Trying higher values has no effect.
+ */
+int picoquic_adjust_max_connections(picoquic_quic_t * quic, uint32_t max_nb_connections);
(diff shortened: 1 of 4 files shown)


picoquic/quicctx.c
```diff
@@ -330,6 +330,16 @@ void picoquic_registered_token_clear(picoquic_quic_t* quic, uint64_t expiry_time
     } while (!end_reached);
 }
 
+void picoquic_adjust_max_connections(picoquic_quic_t * quic, uint32_t max_nb_connections)
+{
+    quic->max_number_connections = max_nb_connections;
+}
```
huitema, line 336: That's probably not sufficient. During the initial context creation, "max_nb connections" is used to set the dimension of the hash tables. If we increase the hash number without resizing the hash tables, the system might become slow.

huitema (changes requested):
We need to be very clear about the effect on performance. Currently, sizing the hash tables based on number of connections ensures that performance remains stable up to that point; refusing further connections enforces that. Should we resize the hash tables when the max number of connections changes? Should we design a two-part system with a target hash table size and a separate max connections? Should we add a separate API to resize the hash table?

picoquic/quicctx.c
victorstewart: oh very good point about technically being able to raise that number higher. we could store another variable called "working maximum" something like that. and restrict it to never raising higher than the maximum set in `picoquic_create`.

victorstewart:
> We need to be very clear about the effect on performance. Currently, sizing the hash tables based on number of connections ensures that performance remains stable up to that point; refusing further connections enforces that. Should we resize the hash tables when the max number of connections changes? Should we design a two-part system with a target hash table size and a separate max connections? Should we add a separate API to resize the hash table?

exactly for performance reasons, I think that the hash tables should be statically sized at initialization time to the maximum number of QUIC connections ever possible... while the working ceiling connections can grow into is allowed to oscillate at or below that level.

I imagine something where as each TCP connection is made, decrementing the working ceiling down by one. and as each closes incrementing it up by one.

as long as the working ceiling is never allowed to grow higher than the initial maximum set, i think all is fine?

victorstewart:
this 2nd commit enforces that maximum ceiling

picoquic/picoquic.h
```diff
@@ -410,6 +410,12 @@ int picoquic_esni_load_key(picoquic_quic_t * quic, char const * esni_key_file_na
 /* Set the ESNI RR. Must be called after setting the ESNI key at least once. */
 int picoquic_esni_server_setup(picoquic_quic_t * quic, char const * esni_rr_file_name);
 
+/* Adjust maximum connections allowed */
+void picoquic_adjust_max_connections(picoquic_quic_t * quic, uint32_t max_nb_connections);
```
huitema, line 414: Maybe update the comment to explain the behavior:
```
/* Adjust maximum connections allowed to the specified value.
 * The maximum number cannot be set to a value higher than the limit set when the context was
 * created. Trying higher values has no effect.
 */
 ```
Consider returning an int, 0 if OK, -1 if failed.

huitema (approved):
Approved, but fix the comment in the header file. Or consider "trying higher values will reset the
value to the initial maximum." Note that in general API that may fail return an error when they do.

picoquic/picoquic.h
victorstewart: done

PR #1201, opened 2021-04-20
Add preemptive repeat logic and test

### LINKED ISSUE #924: Explores per stream FEC for stream delivery latency
On 5/20/2020 2:27 AM, Carsten Bormann wrote:
> On 2020-05-13, at 17:49, Christian Huitema <huitema@huitema.net> wrote:
>> I am running simulations of that lossy configuration as part of the Picoquic test suite. When using BBR, I observe that the large transfer over a high speed link takes a couple seconds more than in the absence of losses. The issue is not congestion, and not even the 1.6% bandwidth consumed by error correction. The problem is the cascade: some of the corrections of lost packets will be lost again, which causes several rounds of retransmission at the end of the transfer.
> Hi Christian,
>
> Since you have data about these cascades, I’ll ask you my curious question:  What is the timing of these?  How long would the WiFi segment had time to do a local retransmission?

The simulation sets a satellite link, 250 Mbps down, 3 Mbps up, 300ms latency each way. The packet size is 1500 bytes. The 600ms RTT translates to a 1sec timer, approximately. In 1 second, the link carries more than 20,000 packets. If 1.6% are lost, that means 320 need to be retransmitted. Out of those 320, 5 or 6 will need to be retransmitted a second time, incurring an additional 2 seconds delay.

Note that this does not consume much bandwidth. The link is available for transmission of further documents. It is strictly a "latency of document transfer" issue. If you are piping a large set of large files over the network, that's hardly a problem, but in some cases it is important.

> Of course, I’m asking this in the context of LOOPS (loops@ietf.org), which is meeting next week in order to nail down its initial charter proposal; support for WiFi “last miles” has always been proposed as a possible subject.

A lot could be done in improving the behavior of Wi-Fi, that's sure. But I really don't like the idea of putting proxies everywhere -- the cost in management and decreased reliability would be enormous. Besides, the deployment issues would be daunting -- who is going to set up all these proxies in all these Wi-Fi networks? And, if they are only deployed in a fraction of the Wi-Fi network, then the transport folks will still need an end-to-end solution.

I have not yet explored the end-to-end solutions to the latency problem, but I could. FEC would cure the latency issue, but FEC trades bandwidth for latency, so you don't want that in the general case. If I had to do something, I would explore "selective FEC". Detect when the document is almost done, when there is only 1 second of transmission left. During that 1 second, apply something like 1/8 FEC for the remaining stream frames. That would make the last 1 second 12.5% less efficient, but would limit the "cascade' effect to 125ms.

huitema:
Add a flag to perform "preemptive repeat" of not yet acknowledged packets if there is no application data to send. In the "satellite test with 1% packet loss", this reduces the transmission time from 7.36 seconds to 6.79 sec, i.e. almost 1 rtt.

Close #924 

There is still work to do before merging this PR:

- Add an option to turn on this function in the demo program
- Verify that the demo program does not wait acknowledgement of redundant data before closing
- Add redundant transmission in the "almost ready" states, subject to connection validation
- Consider redundant transmission of handshake packets
- Consider variant of "many losses" connection test with preemptive repeat turned on

### DIFF
--- a/PerfAndStressTest/PerfAndStressTest.cpp
+++ b/PerfAndStressTest/PerfAndStressTest.cpp
@@ -62,6 +62,13 @@ namespace PerfAndStressTest
             Assert::AreEqual(ret, 0);
         }
 
+        TEST_METHOD(satellite_preemptive)
+        {
+            int ret = satellite_preemptive_test();
+
+            Assert::AreEqual(ret, 0);
+        }
+
         TEST_METHOD(satellite_small)
         {
             int ret = satellite_small_test();
--- a/picoquic/config.c
+++ b/picoquic/config.c
@@ -89,6 +89,7 @@ static option_table_line_t option_table[] = {
     { picoquic_option_Token_File_Name, 'N', "token_file", 1, "file", "File storing the new tokens" },
     { picoquic_option_Socket_buffer_size, 'B', "so_buf_size", 1, "number", "Set buffer size with SO_SNDBUF SO_RCVBUF" },
     { picoquic_option_Performance_Log, 'F', "log_file_name", 1, "file", "Append performance reports to performance log" },
+    { picoquic_option_Preemptive_Repeat, 'V', "preemptive_repeat", 0, "", "enable preemptive repeat" },
     { picoquic_option_HELP, 'h', "help", 0, "This help message" }
 };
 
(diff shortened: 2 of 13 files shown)


huitema:
It appears that adding repetition capability in the "almost ready" state creates a regression. Maybe some interference with slow start or hystart. We will not do that.

huitema:
The option "-V" turns on preemptive repeat in the demo program. The loopback tests initially showed a performance regression, dropping from 2.2 Gbps on Windows to 1.5 Gbps. This was due to interference between preemptive repeat and the "final ack" feature. The final ack feature causes immediate acknowledgement once the requested stream have been delivered (FIN bit); the goal is to expedite the acknowledgement of the fin, and thus the closure of the connection. But the same feature was causing an immediate acknowledgment of each "preemptive repeat" packet, which caused a drop in performance. The fix is to send the "final ack" just once, instead of once for every packet received after fin.

PR #1204, opened 2021-04-30
Remember RTT and CWIn from previous connections

huitema:
This PR implements most of the recommendations from the [0RTT BDP draft](https://github.com/NicoKos/QUIC_HIGH_BDP/blob/master/ietf-document/draft-0rtt-bdp/draft-kuhn-quic-0rtt-bdp-09.txt). However, this PR does not make use of the BDP extension frames defined in the draft, and as such does not require a specific extension from QUIC V1. Instead, the PR lets both client and server remember their BDP parameters independently:

* During a 1RTT session, the client stores the client side RTT and CWIN together with the session resume ticket obtained from the server. The client also store the IP address of the server.
* The server maintains a table of previously issued tickets, indexed by the same random ticket identifier that is used to guarantee uniqueness of the AEAD encryption. Old tokens are removed from the table using LRU logic. For each ticket identifier, the table holds the RTT and CWIN, and also the IP address of the client.

During the 0-RTT session, the nodes wait for the first RTT measurement from the peer's IP address to verify that the parameters have not changed. If the RTT matches expectation, the CWIN is set to the remembered value.

Testing with 250 Mpbs satellite connection, the simulation shows that a 100MB upload time drops from just under 6 seconds with the optimized picoquic code to just under 4.8 seconds when remembering the CWIN and RTT.

### DIFF
--- a/PerfAndStressTest/PerfAndStressTest.cpp
+++ b/PerfAndStressTest/PerfAndStressTest.cpp
@@ -41,6 +41,13 @@ namespace PerfAndStressTest
             Assert::AreEqual(ret, 0);
         }
 
+        TEST_METHOD(satellite_seeded)
+        {
+            int ret = satellite_seeded_test();
+
+            Assert::AreEqual(ret, 0);
+        }
+
         TEST_METHOD(satellite_loss)
         {
             int ret = satellite_loss_test();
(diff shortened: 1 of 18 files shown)


victorstewart:
as we discussed before i think this no extra frame way is definitely the best way to go, as it allows independent evolution and advantage for non-picoquic peers.

nothing catches my eye, looks great!

victorstewart (approved):
as we discussed before i think this no extra frame way is definitely the best way to go, as it allows independent evolution and advantage for non-picoquic peers.

nothing catches my eye, looks great!

PR #1209, opened 2021-05-06
Add support for 0RTT-BDP extension

francklinSIMO:
This PR implements the [0RTT BDP draft](https://github.com/NicoKos/QUIC_HIGH_BDP/blob/master/ietf-document/draft-0rtt-bdp/draft-kuhn-quic-0rtt-bdp-09.txt). It extends this PR [Remember RTT and CWIn from previous connections](https://github.com/private-octopus/picoquic/pull/1204) to allow server to offload previous connections data on the client as defined by the draft. 

The **BDP option** added to picoquicdemo ("-j") and picoquic API ("bdp_option" ) allows to choose :

-  (0) to disable BDP extension
-  (1) to enable BDP extension and store connections data locally in unilateral way as proposed in [Remember RTT and CWIn from previous connections](https://github.com/private-octopus/picoquic/pull/1204)
-  (2) to enable BDP extension and store both client and server connections data on the client using BDP extension frame

The **TP enable_bdp** is defined at both client and server side to negociate the extension behavior

The **BDP Frame** is added to carry BDP parameters

Other related mecanisms especially to hold BDP parameters and to store them, to verify that the parameters have not changed and to interact with CCs are reused from [Remember RTT and CWIn from previous connections](https://github.com/private-octopus/picoquic/pull/1204)  

Testing with 250 Mpbs satellite connection, the simulation shows that a 100MB upload time with bdp_option = 0 drops from just under 6 seconds  to just under 4.8 seconds when remembering the CWIN and RTT (bdp_option = 1 or 2).

Since the draft is still in progress, we (@NicoKos : https://github.com/NicoKos) thought this PR can be useful.


picoquic/sender.c
```diff
@@ -3809,6 +3820,12 @@ int picoquic_prepare_packet_ready(picoquic_cnx_t* cnx, picoquic_path_t* path_x,
                             bytes_next = picoquic_format_available_stream_frames(cnx, bytes_next, bytes_max, &more_data, &is_pure_ack, &stream_tried_and_failed, &ret);
                         }
 
+                        /* Send bdp frames if there are no stream frames to send 
+                         * and if peer wishes to receive bdp frames */
+                        if(!cnx->client_mode && cnx->quic->default_bdp_option == 2 && (cnx->remote_parameters.enable_bdp == 2 || cnx->remote_parameters.enable_bdp == 3)) {
+                           bytes_next = picoquic_format_bdp_frame(cnx, bytes_next, bytes_max, path_x, &more_data, &is_pure_ack);
+                        }
+
```
huitema, line 3828: There is a better way to do that. Inside `picoquic_seed_ticket`, when new values of the CWIN are learned, the code places them in the server ticket hash table. At the same time, it could post a `miscellaneous frame` to queue transmission of the BDP frame.

huitema (commented):
Nice work. We are on the right track. You will find lots of detailed comments in the review, but the main one is that this should be a "bdp_frame_option", leaving the current code that remembers data at the server side unchanged.

I appreciate that you added the tests!

huitema:
Now that we have added the IP parameters, there are only a few issues left:

1) Should tie sending the BDP frame to the evaluation of the CWIN, not send it before that.
2) Should align processing of incoming DP frame at server to processing of "local knowledge"
3) Should add a test of BDP frame (satellite is not enough, because in that test only the client sends data)
4) Should add test verifying that the BDP frame is ignored if the client IP changes
5) Should add test verifying that the BDP frame is ignored if the link RTT changed.

huitema:
The logical tests are now all working, but there is was a flow control issue. The rate was limited by the value of the initial flow control parameter, which prevented taking full advantage of the increased window. I am doing a test in which this initial flow control parameter is set to a large value, but then I find another issue. The CWIN is set at a value that is too large, which causes a lot of losses and in turn slows the link:

image

huitema:
Added a check of the CWIN value before storing it for BDP frame transmission. The CWIN value can overshoot by the end of slow start. Using that value will cause a wild overshoot during the next connection. Instead, we cap it to estimated_bandwidth*rtt_min, which produces better results. But there is still a significant overshot.

huitema:
Fixed the overshot issues for RENO and BBR. There are still too many losses in the case of Cubic, but that investigation will have to happen later.

PR #1216, opened 2021-07-07
Update Build Settings

victorstewart:
i submitted a PR to picotls to enable installing it to a target directory. then added a `PTLS_PREFIX` macro to FindPTLS. so there's now an additional build option to specify where to look for picotls. i think this is the best option.

for example...

`cmake -S . -B build -DPTLS_PREFIX=/root/quicperf/picofast`

i also updated the optimization flags and dropped the debug information. getting about 100Mb/s throughput benefit in picoperf over loopback.


huitema:
There seem to be a couple of issues:

1) CILeakTest is failing, complains about null pointer access.
2) CITestClang is failing, test returns non zero despite all tests passing
3) CITestMacOS is failing, test returns non zero despite all tests passing

These are probably latent code issues that were hidden by running at O2, or with -g.  We need to fix them.

huitema:
We also need to make sure that there is an option to work with O2, -g, if only when debugging is required.

huitema:
The clang based tests are failing. Investigating.
Also, building with clang causes errors:
image

huitema:
Somehow, the optimization flags, when applied to clang, cause a difference of 1 microsecond in the timing recorded for the ECN test:
image
This is using a virtual clock, so it points to some difference in a computation. Maybe the O3 flags caused a different rounding mode?

victorstewart:
easy enough to add a debug build flag. i can do so

oh that's fascinating. ya it must be -ffast-math turned on by Ofast. it does a few borderline things in the name of performance, so maybe let's just go with O3.

the rename registers works on my clang, but probably because i'm on clang 12. might as well just drop that too. no reason to belabor these points.

Issue #1230, opened 2021-08-31
PMTUD creates too much overhead in some applications


huitema:
Path MTU Discovery operates by sending "MTU probe" packets that contain just PING and PAD frames. If a probe of length L gets acknowledged, the code learns that setting the PMTU to L is probably fine. There is logic in Picoquic to try different lengths. The connection only works if the minimum packet size is supported (1232 for IPv6, 1252 for IPv4). If that does not work, the path will not be used at all. Otherwise, the initial PMTU is set to that value. After that, Picoquic will try to send a succession of probes until one works: first try the maximum packet size authorized by the peer, then if that does not work try something intermediate, etc., until either a good size has been found, or it does not make sense to further try.

The problem with that is overhead. The PING+PAD MTU probes are pure overhead. On one hand that's good, because sending those probes does not affect the logic of the application. On the other hand, the overhead creates additional traffic on the links, more packets to process by the peers, etc. Some applications just send a few short messages -- DNS over QUIC would be an example, text messaging would be another. The MTU probes can end up being a big fraction of the total traffic, for little benefit.

The logic in the code checks whether to send PMTUD or not:

* If the transport knows that it will send lots of data, PMTUD is deemed a good investment and the value is set to "required"
* Else it is set to "optional"

If required, the probes are sent as soon as the bandwidth is available. If "optional", they are only sent if there is nothing else to send. But that logic does not help the low-traffic applications, because the probes will still be sent as soon as some network bandwidth becomes available, creating overhead. The reverse problem happens for some high-bandwidth applications, because it might take some time for the transport to decide that the "the application will send lots of data". 

It would probably be nice to have a way to set PMTUD policy, with a few defined behaviors:

* Do PMTUD always, asap -- whether the logic says optional or required.
* Do PMTUD according to the current logic, i.e. send optional packets if the is space.
* Only do PMTUD is there is lots of data to send.
* Never do PMTUD, stick with 1232.

We have one API already defined:
```
void picoquic_cnx_set_pmtud_required(picoquic_cnx_t* cnx, int is_pmtud_required);
```
This API takes a binary parameter. If set to 1, the code will do PMTUD always, which is great for high bandwidth applications. If set to 0, the code will use the current logic. It seems that we need to use two other values for "only if there is lots of data to send" and "never".

The API operates at the individual connection level. That's useful, as it allows changing the behavior of ongoing connections, but in many applications it might be simpler to define the behavior at the QUIC context level.

huitema:
[...]

PR #1231, opened 2021-09-01
Revise pmtud negotiation

huitema:
Implement the solutions discussed in issue #1230


picoquic/quicctx.c
```diff
@@ -2806,6 +2807,11 @@ picoquic_cnx_t* picoquic_create_cnx(picoquic_quic_t* quic,
             cnx->local_parameters.enable_loss_bit = quic->default_lossbit_policy;
             cnx->local_parameters.enable_multipath = quic->default_multipath_option & 1;
             cnx->local_parameters.enable_simple_multipath = (quic->default_multipath_option >> 1) & 1;
+            /* Apply the defined MTU MAX instead of default, if specified */
+            if (cnx->quic->mtu_max > 0)
+            {
+                cnx->local_parameters.max_packet_size = cnx->quic->mtu_max;
+            }
```
victorstewart, line 2814: isn't this always incorrect, since the MTU max to be probed includes the UDP + IPv4/IPv6 header sizes and the max_packet_size is just the maximum quic packet size? and when set per connection, we already know the address of the path thus whether it's IPv4 or IPv6 and can derive the accurate maximum quic packet size accepted?

victorstewart (changes requested):

victorstewart:
all looks great besides that one point of clarity!

picoquic/quicctx.c
huitema: The "MTU MAX" in the code is always "UDP max payload size", consistent with the definitions in QUIC transport. So maybe we should set the name to "udp_payload_max".

Yes, there is an issue of whether this should be optimized for V4 or V6. Here we have a one-size-fits-all, which is a limitation. We could have two parameters per QUIC context, one for V6 and one for V4, and pick the one for the address family of the path's destination address.

victorstewart: oh okay! i misunderstood that until now. guess that means there's a problem with the name for sure.

given that `max_udp_payload_size`, here `max_packet_size`, is a transport parameter and thus must be provided by both applications. i think the cleanest way to do this is to do away with `mtu_max` and expect the opposing client provided the maximum ipv4-derived (because that truly would be the maximum) quic packet size, and then per path once we know if it's ipv4 or 6, we derive the correct maximum IP + UDP + QUIC MTU to probe for.

i think having these 2 opposing inputs is confusing and unnecessary, and also requires defining their interplay. but if necessary for backwards compatibility in the api, if both are set, we clamp to the minimum. 

when `mtu_max` is larger than the opposing party's `max_packet_size`, there's no sense in probing for an MTU higher because the opposing party already said it can't accept packets any larger. but if `mtu_max` is less than the opposing `max_packet_size`, maybe there's some sensible reason the application wants to underutilize the available packet size.

huitema: I am thinking of the people using the API. We make the API more complex, as in:
```
void picoquic_set_mtu_max(picoquic_quic_t* quic, uint32_t mtu_max_ipv4, uint32_t mtu_max_ipv6);
```
I can see people puzzling their head about that. Another possibility is to slightly change the semantics of `picoquic_set_mtu_max`, so it sets the `max packet size`, including the IP and UDP headers, with the implication that `picoquic_set_mtu_max(quic, mtu_max)` is equivalent to:
```
payload_max_ipv4 = mtu_max - 20 - 8;
payload_max_ipv6 = mtu_max - 40 - 8;
```
So the default would be `picoquic_set_mtu_max(quic, 1500)`, or for 5G `picoquic_set_mtu_max(quic, 1420)`.

Would that work?

huitema: The PR #1231 revises the implementation of `mtu_max` to be interpreted as the conventional definition of the MTU.

I think it is important to have a parameter that may override the values proposed by the peer. In the general case, the two peers are independent, not the same admin. It makes sense for the server to limit the packet size that the peer can send, because the server will have to deal with the incoming packets.

victorstewart: i think that's the perfect solution

victorstewart (approved):

PR #1238, opened 2021-10-13
Packet path number for unified loss detection

huitema:
Add a per path packet number to the packet metadata. This enables RACK logic independently of whether QUIC operates with a single active path as in V1, or uses simple multipath, or uses multipath with sequence number spaces per path or per CID.


huitema:
Checking in despite the regressions because cubic for satellite is not a core scenario, and the multipath repair issue will be addressed in an upcoming PR.

PR #1260, opened 2021-11-25
Add (basic) support for QUICv2

IvanNardi:
See: https://datatracker.ietf.org/doc/html/draft-ietf-quic-v2-00


IvanNardi:
This low-level encryption stuff is not my area of expertise, but the unit tests are OK and the client works fine with the server (using `picoquicdemo`)
Two big issues:
* is that the best/simplest  way to "pass around" the tls labels (depending on the version)?
* main point: what about the labels used for the server tickets? I didn't find any reference to them in the QUIC RFCs, and since the tickets are used only by the server, I **think** that we can use any label. If that is not the case, probably this entire MR might be dropped...

TODO:
* extend unit tests
* how v2 works with VN?

huitema:
Thanks for the PR. This is the Thanksgiving weekend here, I will not be able to properly review this for a couple of days. On the "ticket encryption" part, yes this is purely local to the server, There is no need to do anything specific to a QUIC version.

IvanNardi:
> Thanks for the PR. This is the Thanksgiving weekend here

I forgot, sorry! Happy Thanksgiving.

huitema (approved):
Looks good. Thank you for the PR.

PR #1264, opened 2021-12-03
Include chosen version in client's other list.

### LINKED ISSUE #1263: VN: missing Chosen Version in the Client-Sent Other Versions list
It seems that the, in the TP sent by the client, the Chosen Version is not present in the Other Versions list.
According to the Editor Copy of the draft:
```
Note that the version in the Chosen Version field MUST be included in this list to allow the client to communicate the chosen version's preference
```
Am I missing something?

Screenshot from 2021-12-03 15-55-06

huitema:
Close #1263


IvanNardi (approved):
It works, thanks!
Now updating from 1 to 0xFF020000 works with `picoquicdemo` -> mwe.org:443, AFAIK

PR #1268, opened 2021-12-24
High latency tests

huitema:
First test, check that the connection can be established.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,6 +131,7 @@ set(PICOQUIC_TEST_LIBRARY_FILES
     picoquictest/cplusplus.cpp
     picoquictest/datagram_tests.c
     picoquictest/hashtest.c
+    picoquictest/high_latency_test.c
     picoquictest/intformattest.c
     picoquictest/multipath_test.c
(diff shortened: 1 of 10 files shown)


huitema:
Added transmission tests using BBR and Cubic. The traces shows that in both cases, the hystart code exits slow start too soon, estimating the bandwidth at 1.9 Mbps instead of 10 Mbps.

huitema:
Found a first issue in Hystart. Fixing that, the 10MB test succeeds. However, the code remains in slow start for the whole duration.
Need to do increase the transfer size to check that the code properly exits slow-start.

huitema:
Increased the file size to 100MB. BBR test completes in 141 seconds, Cubic test in 160 seconds.

huitema:
Still to do: check that BBR "probe RTT" test does not fire in presence of long delays. Maybe to that by running a long duration test with a 1 Mbps link.

huitema:
Verified that probe RTT does not fire on long duration transfer, with new "probe RTT" test. Verified that code already has safeguards to not fire probe RTT if latency is high. Cleaned up test code.

PR #1273, opened 2022-01-02
Fix binlog crash on bad stream frame

### LINKED ISSUE #1272: Fuzzer causes binlog to crash
Fuzzer can produce faulty frames. A faulty "stream" frame can cause the binlog write to crash, when the "byte" position returned by the parser is NULL.

huitema:
Close #1272


huitema:
@steschu77 Checking this now, because fuzzing app crashes otherwise. But please take a look...

Issue #1274, opened 2022-01-03
Fuzzing the client shows traces of reverse connections.


huitema:
Client fuzzing is possible with [fuzi_q](https://github.com/private-octopus/fuzi_q). One test showed the presence of an extra trace, for a *server* connection. The explanation is simple: fuzzing changed the DCID of an initial message sent by the server, and the received message was interpreted as the first message of a connection started by the server. It did stop there, but it shows a potential issue if a "client only" context is allowed to process incoming connections.

PR #1275, opened 2022-01-03
Block connections to client-only contexts

huitema:


huitema:
@victorstewart see issue #1274. I added a flag to protect a client context against random incoming connection attempts. This closes a potential attack surface. I understand that you are using client auth, you may want to look at the additional API.

Issue #1313, opened 2022-03-11
File corrupted during L1 interop test with MSQUIC


huitema:
An interop runner test between MSQUIC client and Picoquic server reports the following issue:
```
2022-03-11 11:02:19,024 Using the client's key log file.
2022-03-11 11:02:22,752 File contents of /tmp/download_24qp7s48/bwwkgkeord do not match.
```
Analysis of the [server qlog](https://interop.seemann.io/logs/2022-03-11T08:09/picoquic_msquic/handshakeloss/server/qlog/92f3c66ef161b1c7.server.qlog) shows that the server retransmits the packet content multiple times, but at some point that packet becomes corrupted.

Initial transmission in packet number 3 has:
```
{
        "frame_type": "stream",
        "id": 0,
        "offset": 0,
        "length": 1024,
        "fin": true,
        "begins_with": "273d79ef8d4fafce",
        "stream_id": 0
}
```
The same frame is sent in packet numbers 6, 8 and 10, but packet 11 has:
```
      {
        "frame_type": "stream",
        "id": 0,
        "offset": 0,
        "length": 1024,
        "fin": true,
        "begins_with": "78e294403c3c7716",
        "stream_id": 0
      }
```

PR #1314, opened 2022-03-12
Try repro file corruption

huitema:
Trying to reproduce the issue #1313. About the same sequence is replayed, but the memory corruption did not reproduce.


huitema:
Trying to simplify the trace, so add an option to have fewer repeat of the "connection close" message.

Issue #1319, opened 2022-03-15
New Connection ID frame repeated with different content


huitema:
I am looking at the interop runner failed test C1 between neqo client and picoquic server. One of the connections failed, see [client log](https://interop.seemann.io/logs/2022-03-15T08:08/picoquic_neqo/handshakecorruption/client/qlog/server4-24bfa124caa8180a.qlog) and [server log](https://interop.seemann.io/logs/2022-03-15T08:08/picoquic_neqo/handshakecorruption/server/qlog/24bfa124caa8180a.server.qlog). What seems to happen is:

* Picoquic server sends 7 new connection ID frames in 1 RTT packet number 0, with CID number 1 to 7
* Packet number 0 is marked lost after some time
* Content of packet #0 is repeated in packet number 3
* Packet number 3 is acked by client.
* Picoquic repeats New CID number 1 to 7 in packet number 6 but with different values.
* Neqo client detects the issue and closes the connection.

huitema:
Same issue reproes in a test between quic-go client and pricoquic server. See [client log](https://interop.seemann.io/logs/2022-03-23T08:08/picoquic_quic-go/handshakeloss/client/qlog/430292ad14cb182e7fd82d63.qlog) and [server log](https://interop.seemann.io/logs/2022-03-23T08:08/picoquic_quic-go/handshakeloss/client/qlog/430292ad14cb182e7fd82d63.qlog).

huitema:
Issue is hard to repro, points to some random behavior.

The creation of local connection ID is done in `picoquic_create_local_cnxid`. The current code manages two counters: `cnx->local_cnxid_sequence_next` and `cnx->nb_local_cnxid`.

The `new_connection_id` frames are added to outgoing packets in the function `picoquic_format_new_local_id_as_needed`. The code uses `cnx->nb_local_cnxid` to check whether new local CID should be created:
```
        cnx->nb_local_cnxid < ((int)(cnx->remote_parameters.active_connection_id_limit) + cnx->nb_local_cnxid_expired) &&
        cnx->nb_local_cnxid <= (PICOQUIC_NB_PATH_TARGET+cnx->nb_local_cnxid_expired)
```
This code assumes that `cnx->nb_local_cid` describes the total number of connection ID sent to the peer since the beginning of the connection. If the test passes, the connection ID is created by calling `picoquic_create_local_cnxid`. If the function fails, it has no side effect. If is succeeds, it creates a new CID, with the CID sequence number set as:
```
[...]

PR #1335, opened 2022-03-29
Predict header test

huitema:


huitema:
This fixes the buffer overflow that probably caused issue #1319.

Issue #1338, opened 2022-03-30
Extra repeat of bidir streams frame


huitema:
Observed in [interop tests between quic client and quinn server](https://interop.seemann.io/logs/2022-03-30T08:08/picoquic_quinn/multiplexing/). The client is stuck in a loop repeating 

In the trace, the packet is first transmitted as packet number 401.
![image](https://user-images.githubusercontent.com/7464128/160908223-8d555f3b-0a7a-44cc-a292-dc626a037b8a.png)

Packet number 401 is acknowledged later, but transmission does not stop.
![image](https://user-images.githubusercontent.com/7464128/160907815-f6d25b73-12ed-45c7-ad95-520ccd169c64.png)

It will go on almost forever, despite packets being acknowledged and not marked lost. We can also notice that the repeat interval is very short, only a few milliseconds. Wonder whether such very frequent use of preemptive repeat is wise!

huitema:
Verified fixed on interop runner after pr #1339

PR #1339, opened 2022-03-30
Streams bidir loop

huitema:
Fix bug in checking whether to repeat max streams frame


huitema:
Should fix issue #1338

PR #1357, opened 2022-05-19
Discard stream

### LINKED ISSUE #1354: Forget Stream
when tearing down an entire connection, we have the ability to use `picoquic_delete_cnx` and instantly forget the connection without waiting for synchronization with the opposing party. this allows for optimal simple one-step cleanup of memory and objects surrounding the connection and its streams.

but there is no way to do this for a stream when closing it. thus whether we use the stop sending or FIN pathway, there will be some future callback for the stream on the connection (because the connection will persist). this requires the implementer to leave the memory/object representing the stream in limbo, and finish cleaning it up later.

at present i've implemented this by creating a dummy pointer and assigning it as the stream_ctx with `picoquic_set_app_stream_ctx`, then checking in the callback if the stream_ctx is this pointer and bouncing out if so. but it would be more efficient to have this inside of picoquic so the callbacks are never called, any data received dropped, etc.

huitema:
Close #1354

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -283,9 +283,11 @@ const uint8_t* picoquic_decode_stream_reset_frame(picoquic_cnx_t* cnx, const uin
         picoquic_update_max_stream_ID_local(cnx, stream);
 
         if (cnx->callback_fn != NULL && !stream->reset_signalled) {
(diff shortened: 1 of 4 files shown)


victorstewart (approved):
all looks great!

Issue #1363, opened 2022-05-23
Connection Close Without Chatting


victorstewart:
it can be advantageous sometimes to close a connection without informing the opposing party. for example in an application server when malformed messages are discovered. this could be due to a client bug, or a malicious party proding the server. either way spending the resources to gracefully close the connection not only costs something, but can greatly increase the bad traffic volume: a buggy client might get into a connect -> close -> reconnect loop, or an attacker is now provided clarity as to boundaries and provided signals with which to increase the rate of their attacks.

this can be accomplished currently by using `picoquic_delete_cnx` but the problem is this function can not be called from within `picoquic_stream_data_cb_fn`, so the application operation code is then forced to be aware whether it's running within the callback or elsewhere plus (if running within the callback) discard all streams (to make sure no more data is read in) and persist the  cnx object for deletion after the callback cycle completes.

so it would be great to have a single consistent path here, so that one function can always be called and all this state awareness and persisting and cleanup can be avoided. something like `picoquic_close` with a silent flag.

huitema:
Solved in PR #1364

PR #1364, opened 2022-05-23
Adding an immediate close API.

huitema:
This should address issue #1363 

@victorstewart I understand that part of your problem regards callbacks and using the API from within a callback. This is very hard to test. I could do a simulation, but the simulation would only prove that the particular code in the simulation does not cause problem. I do believe that it works, because the state variable is never overwritten, and because the actual clearing of resource will only happen after a timeout. But...

### DIFF
--- a/UnitTest1/unittest1.cpp
+++ b/UnitTest1/unittest1.cpp
@@ -626,6 +626,13 @@ namespace UnitTest1
             Assert::AreEqual(ret, 0);
         }
 
+        TEST_METHOD(immediate_close)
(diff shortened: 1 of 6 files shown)


victorstewart:
so this will invoke the `picoquic_callback_close` event after a short wait?

you could merge it and put experimental on the immediate close, and then i can live with it for a while and see if i notice any problems.

victorstewart (approved):

huitema:
Close #1363

PR #1365, opened 2022-05-27
fix sni when server address is IP

clarkzjw:
When `sni == NULL` and `ip_address_text` is a valid IP format, `is_name` would be `0`, thus `sni` should be `ip_address_text` either. So checking whether `is_name != 0` is redundant or is there any other consideration?

### DIFF
--- a/picoquicfirst/picoquicdemo.c
+++ b/picoquicfirst/picoquicdemo.c
@@ -494,7 +494,7 @@ int quic_client(const char* ip_address_text, int server_port,
 
     if (ret == 0) {
         ret = picoquic_get_server_address(ip_address_text, server_port, &loop_cb.server_address, &is_name);
-        if (sni == NULL && is_name != 0) {
+        if (sni == NULL) {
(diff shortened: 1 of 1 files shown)


huitema (commented):
I have two reservations:

1) Do we want this for all ALPN, or just for H3?
2) What are the side effects, and how do we test them?

We need for a "Host" header in HTTP transactions. The H3 code fills that with the SNI, so filling the SNI will indeed solve the H3 issue. However, the SNI is also the input to the TLS certificate verification algorithm. I am concerned that setting the SNI to the address text might cause certificate verification to fail.

I am specifically concerned with the scenario in which a researcher starts a server using test certificates such as provided in the distribution. Have we verified that the connection does succeed? How hard would it be to add a variation of `h3zero_server_test()` that tested `sni = ip_address_text`?

clarkzjw:
@huitema Thanks for the comment and info. I did miss the certificate issues you mentioned. 
Closing this pr for now, I'll try to look into them later.

PR #1366, opened 2022-06-01
WIP: add multipath support in demo client

### LINKED ISSUE #1134: Multipath support in demo client
Just enabling the multipath option does not cause the client to actually use multiple paths. Logically, it should. But we need to design that well.

The first idea would be to add an option to specify an additional path. This is simple to design, but it requires in practice adding support for multiple sockets, which implies a pretty large change in the socket loop code. Also, it is difficult to use, because the user has to find out which of the local addresses the client will chose by default, and thus which one should be the other address.

Another design would be to do that automatically inside the client if multipath is enabled. For example, list the IP addresses available on the client using a system API, then automatically create paths for those that are not in use. The new path could in theory use the same port and UDP socket as the other paths.

clarkzjw:
This is a work in progress pull request. The code is still quite rough right now, I just would like to know whether I'm in the right direction. Eventually, it should fix #1134. 

Notes:
1. Only IPv4 and H3 are considered so far. 
2. Added two command line options for the client, `-d <default ip>` to bind the interface where the IP is on as the default path, `-A <alternative ip>` for the second IP/alternative path. (The letters are all temporary choices)
3. I'm not quite sure where to put `picoquic_probe_new_path`.
4. I'm testing it on Mininet.

<details>
  <summary>The topology used for testing</summary>
  
```python
#!/usr/bin/python3

from mininet.net import Mininet
from mininet.link import TCLink
from mininet.cli import CLI
from mininet.log import info

max_queue_size = 10000

net = Mininet(cleanup=True)
c0 = net.addController('c0')

h1 = net.addHost('h1', ip='192.168.1.1/24')
h2 = net.addHost('h2', ip='192.168.1.2/24')

s1 = net.addSwitch('s1')
s2 = net.addSwitch('s2')

net.addLink(h1, s1, cls=TCLink)
net.addLink(s1, h2, cls=TCLink)
net.addLink(h1, s2, cls=TCLink)
net.addLink(s2, h2, cls=TCLink)

h1.setIP('192.168.2.1/24', intf='h1-eth1')
h2.setIP('192.168.2.2/24', intf='h2-eth1')

net.start()

info('*** Running CLI\n')
CLI(net)

info('*** Stopping network')
net.stop()
```
</details>

<details>
  <summary>Example commands I used for client and server</summary>
  
```bash
# on the server
./picoquicdemo -c ./ca-cert.pem -k ./server-key.pem -p 4433 -w ./static/ -M 1
# on the client
./picoquicdemo -M 1 -A 192.168.2.2 -o /tmp 192.168.1.1 4433 index.html
# for the above code to work, I do apply https://github.com/private-octopus/picoquic/pull/1365 locally, 
# which is not included in the PR.
```
</details>

However, I can only see the traffic on the default path where `-d` is specified. 
@huitema I'm not quite familiar with socket programming and QUIC yet. Any suggestions are much appreciated, e.g., whether the Mininet topology/testing methodology is correct, or whether I'm still on the right track. Thanks!


picoquicfirst/picoquicdemo.c
```diff
@@ -293,6 +294,16 @@ int client_loop_cb(picoquic_quic_t* quic, picoquic_packet_loop_cb_enum cb_mode,
             fprintf(stdout, "Waiting for packets.\n");
             break;
         case picoquic_packet_loop_after_receive:
+            if (cb_ctx->cnx_client->is_1rtt_acked) {
+                // TODO
+                uint64_t simulated_time = 0;
```
clarkzjw, line 299: I'm pretty sure `simulated_time` here is not correct.

picoquicfirst/picoquicdemo.c
```diff
@@ -293,6 +294,16 @@ int client_loop_cb(picoquic_quic_t* quic, picoquic_packet_loop_cb_enum cb_mode,
             fprintf(stdout, "Waiting for packets.\n");
             break;
         case picoquic_packet_loop_after_receive:
+            if (cb_ctx->cnx_client->is_1rtt_acked) {
```
clarkzjw, line 297: Is probing a new path after the 1RTT packet is acked the right timing?

picoquic/picosocks.c
```diff
@@ -22,7 +22,7 @@
 #include "picosocks.h"
 #include "picoquic_utils.h"
 
-int picoquic_bind_to_port(SOCKET_TYPE fd, int af, int port)
+int picoquic_bind_to_port(char *ip, SOCKET_TYPE fd, int af, int port)
 {
```
huitema, line 26: You probably do not need that. The picoquic loop uses socket options to specify the outgoing IP address. Binding one socket to a port number is sufficient.

picoquic/picosocks.c
```diff
@@ -303,7 +308,7 @@ int picoquic_open_server_sockets(picoquic_server_sockets_t* sockets, int port)
             }
             ret = picoquic_socket_set_pkt_info(sockets->s_socket[i], sock_af[i]);
             if (ret == 0) {
-                ret = picoquic_bind_to_port(sockets->s_socket[i], sock_af[i], port);
+                ret = picoquic_bind_to_port(NULL, sockets->s_socket[i], sock_af[i], port);
             }
```
huitema, line 312: Please don't do that. Different applications already use "bind_to_port". If you really believe that it is necessary to bind sockets to addresses, you should create a different call, "bind_to_port_and_ip". You could then rewrite "bind_to_port" to use the new call with a NULL argument. But see the other comments, I don't think this new function is necessary.

picoquic/picosocks.h
```diff
@@ -172,7 +172,7 @@ typedef struct st_picoquic_server_sockets_t {
     SOCKET_TYPE s_socket[PICOQUIC_NB_SERVER_SOCKETS];
 } picoquic_server_sockets_t;
 
-int picoquic_bind_to_port(SOCKET_TYPE fd, int af, int port);
+int picoquic_bind_to_port(char *ip, SOCKET_TYPE fd, int af, int port);
```
huitema, line 176: See discussion.

picoquic/sockloop.c
```diff
@@ -134,11 +135,11 @@ int picoquic_packet_loop_open_sockets(int local_port, int local_af, SOCKET_TYPE
...
         int recv_set = 0;
         int send_set = 0;
-        
+
         if ((s_socket[i] = socket(sock_af[i], SOCK_DGRAM, IPPROTO_UDP)) == INVALID_SOCKET ||
             picoquic_socket_set_ecn_options(s_socket[i], sock_af[i], &recv_set, &send_set) != 0 ||
             picoquic_socket_set_pkt_info(s_socket[i], sock_af[i]) != 0 ||
-            picoquic_bind_to_port(s_socket[i], sock_af[i], local_port) != 0 ||
+            picoquic_bind_to_port(ip, s_socket[i], sock_af[i], local_port) != 0 ||
             picoquic_get_local_address(s_socket[i], &local_address) != 0)
```
huitema, line 143: see discussion.

picoquic/sockloop.c
```diff
@@ -237,9 +240,15 @@ int picoquic_packet_loop(picoquic_quic_t* quic,
...
+    if ((nb_sockets = picoquic_packet_loop_open_sockets(config->multipath_default_ip, local_port, local_af,
+                                                        s_socket, sock_af, sock_ports, socket_buffer_size,
+                                                        PICOQUIC_PACKET_LOOP_SOCKETS_MAX)) == 0) {
         ret = PICOQUIC_ERROR_UNEXPECTED_ERROR;
+    } else if (config->multipath_alternative_ip != NULL) {
+        nb_sockets += picoquic_packet_loop_open_sockets(config->multipath_alternative_ip, local_port, local_af,
+                                                        s_socket+nb_sockets, sock_af+nb_sockets,
+                                                        sock_ports+nb_sockets, socket_buffer_size,
+                                                        PICOQUIC_PACKET_LOOP_SOCKETS_MAX);
     }
```
huitema, line 252: see discussion.

picoquic/sockloop.c
```diff
@@ -440,8 +449,9 @@ int picoquic_packet_loop(picoquic_quic_t* quic,
             int sock_ret;
             int testing_nat = (ret == PICOQUIC_NO_ERROR_SIMULATE_NAT);
 
-            sock_ret = picoquic_packet_loop_open_sockets(0, sock_af[0], &s_mig, &s_mig_af,
-                &next_port, socket_buffer_size, 1);
+            sock_ret = picoquic_packet_loop_open_sockets(config->multipath_default_ip, 0,
+                                                         sock_af[0], &s_mig, &s_mig_af,
+                                                         &next_port, socket_buffer_size, 1);
             if (sock_ret != 1 || s_mig == INVALID_SOCKET) {
```
huitema, line 455: See discussion. You don't need that. In fact, it is probably harmful. The system knows how to select the most appropriate IP for a destination. For example, if you call loopback, it will select loopback address as source. Binding sockets to fixed IP addresses will change that behavior.

picoquicfirst/picoquicdemo.c
huitema: It is probably better to wait until the connection is in state "picoquic_state_ready", which means both client and server are synchronized.

picoquicfirst/picoquicdemo.c
huitema: What do you use it for? The socket loop is designed to use real time, not simulated time. The "current_time" variable normally tracks that.

picoquic/sockloop.c
```diff
@@ -109,7 +110,7 @@ static int udp_gso_available = 0;
 #endif
 #endif
 
-int picoquic_packet_loop_open_sockets(int local_port, int local_af, SOCKET_TYPE * s_socket, int * sock_af, 
+int picoquic_packet_loop_open_sockets(char *ip, int local_port, int local_af, SOCKET_TYPE * s_socket, int * sock_af,
     uint16_t * sock_ports, int socket_buffer_size, int nb_sockets_max)
 {
```
huitema, line 115: I am really not sure that passing an IP parameter at that point is useful. The selected source IP normally depends on the destination and address family. It would probably be much simpler to concentrate on the selection of the second source address during the "probe" call.

huitema (commented):
I think this is a bit too complicated. The picoquic loop uses socket options to specify the outgoing IP address. You can definitely have several different source addresses for the same socket, you don't need to bind sockets to addresses.

The sequence is:

1) Picoquic client start, parameter includes the socket address (protocol, IP, port) of the destination.

2) Loop polls the quic client context, with the "prepare" call. Output parameter includes "addr_from" and "addr_to"

3) When sending to a server address for the first time, the source address is not specified; AF_FAMILY is set to zero; the code will not set the "source IP" argument in the socket call; the system stack will pick the outgoing address based on local configuration, routing tables, etc.

4) When the server responds, the code reads the source and destination address of the incoming packet and passes it to picoquic (incoming packet). Picoquic finds the connection from the connection ID and maybe the incoming address, and matches the packet with a "path". It might either select an existing path if source and dest address match, update an existing path if the server address matches the remote address of the path and the local address of the path is unspecified, in which case the local address; or create a new path. There is also code to handle NAT traversal.

5) The next time picoquic prepares a packet, both source and destination address will be specified.

Note that the current behavior implicitly supports multiple paths. For example, a server might accept connections from IPv4 or IPv6; in IPv4, it will use different source addresses depending on whether the client comes from 127.0.0.1 or from a remote address; in IPv6, the server source address could be ::1 for local loop clients, a link local address for clients coming through link local; or a global address. If the server is multi-homed, the global address will be that of the local interface used to communicate to the client.

In theory, you just have to specify the alternative address as a source address when sending the "probe". The client will create a path with that source address as local address and the server address as remote. There is of course a little subtlety there, because you really want this second address to not be the one that the server picked by default. But that's simple to avoid:

1) Have two or more "ip addresses"
2) Read the address already used as local address on path[0] for the connection.
3) Compare that local address to the list of alternative addresses and select the first alternative address that does not match
4) Set the sockaddr for the source address of the probe from that selected address and the current port, set the destination address to the server address, and issue the probe.

picoquic/picosocks.c
huitema: Also, if you do that, consider something like:
```
int picoquic_bind_to_address(SOCKET fd, struct sockaddr * local_addr)
```
That will be more generic than passing IP addresses as text strings.

clarkzjw:
I updated and removed some redundant socket binding code, and if I got it correctly, assume the following scenario:

image012

`h1` is a multi-homed client with two different IPs, and `h2` is a video server. 
When the `picoquicclient` on `h1` starts and connects to `h2`, the system stack picks a default path (probably by default routing policy). After the client and server negotiation finishes (e.g., `cnx_state==picoquic_state_ready`), the client calls `picoquic_probe_new_path`, registering an alternative IP and a new path is created. Then probing packets are sent to the server on the alternative physical path. After the server responds to probing packets, the multipath is actually established. 

My questions are:
1. When does the client actually send the probing packets and how to confirm the probing succeeds and multipath is actually established? Do we need something similar to [`wait_multipath_ready`](https://github.com/private-octopus/picoquic/blob/master/picoquictest/multipath_test.c#L354) from the tests?
2. After the multipath is created, how does the packet/path scheduling work? i.e., what happens after (4):
> Set the sockaddr for the source address of the probe from that selected address and the current port, set the destination address to the server address, and issue the probe.

3. Though this question sounds a bit early, how do we extend and add new schedulers?
4. This is not related to picoquic, however, while debugging on Mininet or testbed platforms, do we need extra routing configurations like MPTCP? e.g., http://csie.nqu.edu.tw/smallko/sdn/mptcp-test.htm

Thanks.

picoquicfirst/picoquicdemo.c
```diff
@@ -679,6 +689,12 @@ int quic_client(const char* ip_address_text, int server_port,
 
     /* Wait for packets */
     if (ret == 0) {
+        struct sockaddr_in alt_addr;
+        memset(&alt_addr, 0, sizeof(alt_addr));
+        alt_addr.sin_family = AF_INET;
+        alt_addr.sin_addr.s_addr = inet_addr(config->multipath_alternative_ip);
+        memcpy(&loop_cb.client_alt_address, &alt_addr, sizeof(alt_addr));
+
```
huitema, line 697: This code uses "inet_addr(), which is not portable. We get the following warning treated as error in Windows builds:
```
c:\projects\picoquic\picoquicfirst\picoquicdemo.c(695): error C2220: warning treated as error - no 'object' file generated [C:\projects\picoquic\picoquicfirst\picoquicfirst.vcxproj]
    21>c:\projects\picoquic\picoquicfirst\picoquicdemo.c(695): warning C4996: 'inet_addr': Use inet_pton() or InetPton() instead or define _WINSOCK_DEPRECATED_NO_WARNINGS to disable deprecated API warnings [C:\projects\picoquic\picoquicfirst\picoquicfirst.vcxproj]
         c:\program files (x86)\windows kits\10\include\10.0.18362.0\um\winsock2.h(1831): note: see declaration of 'inet_addr'
```
Please use the utility function from the library:
```
int picoquic_store_text_addr(struct sockaddr_storage* stored_addr, const char* ip_address_text, uint16_t port)
```
It is portable, not obsoleted, and will also automatically deal with IPv4/IPv6 issues. Set port to 0 if you only care about the IP address.

picoquic/sockloop.c
huitema: Not a real problem, but as a rule it is better to avoid cosmetic changes in a PR...

picoquicfirst/picoquicdemo.c
```diff
@@ -980,6 +996,11 @@
             case '1':
                 just_once = 1;
                 break;
+            case 'A':
+                config.multipath_alternative_ip = malloc(sizeof(char) * strlen(optarg));
```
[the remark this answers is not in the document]
huitema: Please fix this to `malloc(sizeof(char) * (strlen(optarg)+1))`, to account for the NULL character that C places at the end of strings.

huitema:
To answer your question on tests and scheduling. The current support for scheduling in picoquic is limited. Packets get scheduled on the next path that has capacity (congestion control) and is deemed functional, i.e., did not recently experience packet losses. We could do some more complex scheduling, such as marking affinity between streams and paths, or sending multiple redundant copies of the same data packets on several streams for reliability. We could also do something about the datagram API, to provide affinity between a datagram `stream` and a path. If you want to do that, that probably could let you write and publish interesting papers. Separate PR, though.

We do not have picoquic callbacks that indicate that a path has been validated, or . That might be a nice addition. Again, let's consider that in a separate PR.

The code keeps statistics of how much data was sent on a given path. You can find that in the "path" context, visible if the header `picoquic_internal.h` is included. You can probably use that in this PR to print statistics.

huitema (approved):
Looks good! Thank you for this work.

PR #1369, opened 2022-06-09
add ifindex option to picoquicdemo cmdline for multipath

clarkzjw:
Per the discussion from https://github.com/private-octopus/picoquic/issues/1134#issuecomment-1146838889, this PR adds `ifindex` to `-A` option on the client side of `picoquicdemo`. 

Option description for `-A`:

```
-A ifindex,ip         Interface index and ip for multipath alternative path, e.g. 3,10.0.0.2
```

Since `-A` only exists for `picoquicdemo`, we cannot utilize `option_table_line_t.nb_params_required` and `static int config_set_option` to control the number of parameters. Therefore, the only solution I can come up with is to use ',' as the delimiter and manually parse the parameter string by the function `picoquic_parse_client_multipath_config` I added. 

Only the scenario of IPv4 with H3 has been tested, although IPv6 is expected to work as well. 
The current implementation only supports one alternative path.

### DIFF
--- a/picoquicfirst/picoquicdemo.c
+++ b/picoquicfirst/picoquicdemo.c
@@ -320,7 +320,7 @@ int picoquic_parse_client_multipath_config(char *mp_config, int *src_if, struct 
     struct sockaddr_storage ip;
     while ((token = picoquic_strsep(&str, "/"))) {
         if (picoquic_store_text_addr(&ip, token, 0) == 0) {
-            *alt_ip = ip;
+            memcpy(alt_ip, &ip, sizeof(struct sockaddr_storage));
         }
         if (atoi(token) > 0) {
             *src_if = atoi(token);
--- a/picoquicfirst/picoquicdemo.c
+++ b/picoquicfirst/picoquicdemo.c
@@ -281,8 +281,10 @@ typedef struct st_client_loop_cb_t {
     picoquic_connection_id_t client_cid_before_migration;
 } client_loop_cb_t;
 
+
+char * picoquic_strsep(char **stringp, const char *delim)
+{
 #ifdef _WINDOWS
-char *strsep(char **stringp, const char *delim) {
     if (*stringp == NULL) {
         return NULL;
     }
@@ -293,11 +295,13 @@ char *strsep(char **stringp, const char *delim) {
         (*stringp)++;
     }
     return token_start;
-}
+#else
+    return strsep(stringp, delim);
 #endif
+}
 
 /*
- * mp_config is consisted with src_if and alt_ip, seperated by ","
+ * mp_config is consisted with src_if and alt_ip, seperated by "/"
  * where src_if is an int, and alt_ip is a string.
  * alt_ip is the ip of the alternative path
  * src_if is the index of the interface where the alt_ip is bounded with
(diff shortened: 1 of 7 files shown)


picoquicfirst/picoquicdemo.c
```diff
@@ -276,10 +276,52 @@ typedef struct st_client_loop_cb_t {
...
+}
+#endif
+
+/*
+ * mp_config is consisted with src_if and alt_ip, seperated by ","
+ * where src_if is an int, and alt_ip is a string.
+ * alt_ip is the ip of the alternative path
+ * src_if is the index of the interface where the alt_ip is bounded with
+ */
+int picoquic_parse_client_multipath_config(char *mp_config, int *src_if, struct sockaddr_storage *alt_ip)
```
clarkzjw, line 305: Is there any better way? I really don't want to introduce a self-implemented `strsep` for Windows.

picoquicfirst/picoquicdemo.c
```diff
@@ -276,10 +276,52 @@ typedef struct st_client_loop_cb_t {
...
+    char *token_start = *stringp;
+    *stringp = strpbrk(token_start, delim);
+    if (*stringp) {
+        **stringp = '\0';
+        (*stringp)++;
+    }
+    return token_start;
+}
+#endif
+
```
huitema, line 298: I understand your concern about "writing strsep for windows", but i think it is one of those cases in which we have to write our own code. My only concern with your implementation is that you reuse the name "strsep", which may cause problems if some other implementer creates their own copy of strsep in a different module, or in their application. I would prefer something like:

```
char * picoquic_strsep(char **stringp, const char *delim)
{
#ifdef _WINDOWS
    if (*stringp == NULL) {
        return NULL;
    }
    char *token_start = *stringp;
    *stringp = strpbrk(token_start, delim);
    if (*stringp) {
        **stringp = '\0';
        (*stringp)++;
    }
    return token_start;
#else 
    return strsep(stringp, delim);
#endif
}
```

and then use `picoquic_strsep` in your parsing code.

picoquicfirst/picoquicdemo.c
huitema: I am not sure. I am looking at the textual representation of IP addresses and interfaces on different systems. These are commonly used with IPv6 link local addresses. On windows, I see for example:
```
Wireless LAN adapter Wi-Fi:
   Link-local IPv6 Address . . . . . : fe80::e019:8019:6e0c:695c%16
   IPv4 Address. . . . . . . . . . . : 192.168.1.106
```

But on Mac/OS I see something like
```
llw0:
   inet6 fe80::e019:8019:6e0c:695c%llw0 prefixlen 64 scopeid 0xf
```
So, sadly, there is no standard. I would not have picked "," as a separator, but the only strong rule is that we must not use ".", ":" or "&". Maybe "/". Also, it might be a bit more conventional to write the interface index after the address than before it. Something like:
```
192.168.1.106/16
```
or
```
2001:23:45::ead5:6789:def0/16
```

huitema (commented):
Thanks for doing this! I entered two minor comments below, do what you want with them.

picoquicfirst/picoquicdemo.c
clarkzjw: Alright, I changed it to '/'.

huitema (approved):
Looks great, thanks.

clarkzjw:
Oops, CI failed. Let me fix it.

huitema:
Good fix for the memory leak, apparently! Also, good thing that the test suite catches those issues.

PR #1370, opened 2022-06-10
support multiple alt paths in picoquicdemo

clarkzjw:
This PR adds support to specify multiple alternative paths in picoquicdemo by extending #1369.

The syntax for option `-A` changes to

```
-A ip/ifindex[,ip/ifindex]  IP and interface index for multipath alternative path, e.g. 10.0.0.2/3,10.0.0.3/4
```

The maximum number of alternative paths allowed is specified by reusing `PICOQUIC_NB_PATH_TARGET`.

### DIFF
--- a/picoquicfirst/picoquicdemo.c
+++ b/picoquicfirst/picoquicdemo.c
@@ -409,8 +409,8 @@ int client_loop_cb(picoquic_quic_t* quic, picoquic_packet_loop_cb_enum cb_mode,
 
                 if (picoquic_get_cnx_state(cb_ctx->cnx_client) == picoquic_state_ready && cb_ctx->multipath_probe_done == 0) {
                     for (int i = 0; i < cb_ctx->nb_alt_paths; i++) {
-                        if (picoquic_probe_new_path_ex(cb_ctx->cnx_client, (struct sockaddr *)&cb_ctx->server_address,
-                                                       (struct sockaddr *)&cb_ctx->client_alt_address[i], cb_ctx->client_alt_if[i], picoquic_get_quic_time(quic), 0)) {
+                        if ((ret = picoquic_probe_new_path_ex(cb_ctx->cnx_client, (struct sockaddr *)&cb_ctx->server_address,
+                                (struct sockaddr *)&cb_ctx->client_alt_address[i], cb_ctx->client_alt_if[i], picoquic_get_quic_time(quic), 0)) != 0) {
                             picoquic_log_app_message(cb_ctx->cnx_client, "Probe new path failed with exit code %d\n", ret);
                         } else {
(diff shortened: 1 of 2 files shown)


picoquicfirst/picoquicdemo.c
```diff
@@ -391,13 +408,15 @@ int client_loop_cb(picoquic_quic_t* quic, picoquic_packet_loop_cb_enum cb_mode,
...
-                    if (picoquic_probe_new_path_ex(cb_ctx->cnx_client, (struct sockaddr *)&cb_ctx->server_address,
-                            (struct sockaddr *)&cb_ctx->client_alt_address, cb_ctx->client_alt_if, picoquic_get_quic_time(quic), 0)) {
-                        picoquic_log_app_message(cb_ctx->cnx_client, "Probe new path failed with exit code %d\n", ret);
-                    } else {
-                        picoquic_log_app_message(cb_ctx->cnx_client, "New path added, total path available %d\n", cb_ctx->cnx_client->nb_paths);
+                    for (int i = 0; i < cb_ctx->nb_alt_paths; i++) {
+                        if (picoquic_probe_new_path_ex(cb_ctx->cnx_client, (struct sockaddr *)&cb_ctx->server_address,
+                                                       (struct sockaddr *)&cb_ctx->client_alt_address[i], cb_ctx->client_alt_if[i], picoquic_get_quic_time(quic), 0)) {
+                            picoquic_log_app_message(cb_ctx->cnx_client, "Probe new path failed with exit code %d\n", ret);
+                        } else {
```
huitema, line 415: I should have caught that when reviewing the previous PR, but the code does not read the return code of `picoquic_probe_new_path_ex`. The corrected code should be:
```
                        if ((ret = picoquic_probe_new_path_ex(cb_ctx->cnx_client, (struct sockaddr *)&cb_ctx->server_address,
                                                       (struct sockaddr *)&cb_ctx->client_alt_address[i], cb_ctx->client_alt_if[i], picoquic_get_quic_time(quic), 0)) != 0) {
                            picoquic_log_app_message(cb_ctx->cnx_client, "Probe new path failed with exit code %d\n", ret);
                        } else {
```

huitema (commented):
Looks very good. very useful too, as a way to test multipath. Please fix the issue with testing the return code.

Note that the success of "trying a new path" depends on both the local client and the remote server:
* the local client is telling the remote server that it can accept up to `PICOQUIC_NB_PATH_TARGET` connection IDs
* the remote server cannot send more connection IDs than that, but it could send less
* the remote server will also advertise the number of connection ID that they can support, and that may well be lower than `PICOQUIC_NB_PATH_TARGET`

So, the probing will sometime fail, and it is important to provide feedback!

clarkzjw:
I see. I will fix it.

picoquicfirst/picoquicdemo.c
clarkzjw: Fixed.

huitema (approved):

PR #1373, opened 2022-06-21
Fix picoquic_addr_text port to host byte order

robertsong2019:
addr text port is in network byte order.
fix it.


huitema (commented):
Good fix! Thank you!

PR #1380, opened 2022-07-13
insert newline at the eof - needed when integrating picoquic as source

suhasHere:


huitema (commented):
Yes, that's good. I wonder whether there is an easy way to test for that in the CI tests.

PR #1381, opened 2022-07-13
fail if eoline missing

huitema:


huitema:
The test is broadly incompatible with clang-format

PR #1386, opened 2022-07-16
CMake changes

paulej:
I revisited this when I noted one (more) of the downstream projects required explicit inclusion of Threads::Threads, which should have been automatic when linking to target "picoquic-core".  This change does address that issue.

But since I was in here ... :)

I could not see a need for OpenSSL here explicitly since picotls is referenced and satisfies that dependency.

I also noticed that there were multiple and conflicting build flags in the dependency chain, so I introduced code to address that.

Many things were referenced that didn't seem to be needed, so I commented those out.

As I did with quicrq, I think it would be nice to introduce ctest to allow "make test" to work and also introduce an option to disable building any tests when included in downstream projects.

Anyway, let me know if you're interested in any of these kinds of changes.  I'm happy to do additional work on this if it is of value.


paulej:
I can see I was clearly wrong about the OpenSSL dependencies.  Those apparently do exist when not using FetchContent.

huitema:
There are two issues with openssl. First, picotls provides libraries like "picotls-core", and those libraries have references to the "libcrypto" from the openssl project. Which means the application needs to be linked with "libcrypto". Second, picoquic itself uses some of the openssl libraries for handling certificates -- see for example the code of the function `picoquic_openssl_get_certificate_verifier` in `tls_api.c`. That's why we have a line `find_package(OpenSSL)` in `CMakeList.txt`, and the references to openssl in the build. 

The nice part about "fetch" is that the same operation will fetch picotls, find openssl for picotls, the find openssl for picoquic. That pretty much guarantees that the libraries and header files will be consistent, which is nice.

huitema:
Maybe related. When I "make" picoquic on my iMac, I get lots of messages like:

```

[ 97%] Built target picohttp_ct
Consolidate compiler generated dependencies of target picoquic_sample
[ 98%] Building C object CMakeFiles/picoquic_sample.dir/sample/sample.c.o
clang: warning: optimization flag '-frename-registers' is not supported [-Wignored-optimization-argument]
[ 98%] Building C object CMakeFiles/picoquic_sample.dir/sample/sample_client.c.o
clang: warning: optimization flag '-frename-registers' is not supported [-Wignored-optimization-argument]
[ 99%] Building C object CMakeFiles/picoquic_sample.dir/sample/sample_server.c.o
clang: warning: optimization flag '-frename-registers' is not supported [-Wignored-optimization-argument]
```

That was not happening before the previous change.

paulej:
I guess this is a difference between "clang" and "appleclang".  I'll change the generator expression to differentiate between the two.  I assume that's not on master, but just this branch (since I only touched that on this branch).

paulej:
Or, it might have been a copy/paste error. 😳

I put back the stuff required to build against OpenSSL with not using `FetchContent`.  All seems well.  However, some stuff is still commented out, as I didn't see an explicit need for it (e.g., picoquicdemo).  Also, I did not see a need for `CMAKE_DL_LIBS`.  I left those in place, but commented out.

I still noted that some of the compile settings are being set in `ENABLE_ASAN` and `ENABLE_UBSAN`.  Ideally, we'd put those into the target build options.  However, there might be good reason for these to be separate.  What is the preference?  My objective with those changes was to prevent things from trickling down to picotls when it is fetched.  That will definitely happen.  I'm not sure if CMake will trickle those setting up or not.

The last thing I've not revisited it CTest.  Would you like this to work via `make test` similar to how I did it Quicrq?

huitema:
The clang/appleclang issue was probably there before. The Clang test used to catch it, but CMAKE_CXX_COMPILER_ID is now set to "AppleClang". Not related with these changes, but still very annoying. I could fix it in a separate PR.

huitema:
The picoquic build has a set of potential targets. They were all built together in the original version. That would be:

* core library (picoquic)
* http library
* demo application
* code sample
* test library
* test app -- core
* test app -- http

Users building applications on top of QUIC may need only the picoquic library. These users will benefit from an integrated build/fetch process. Some users also need the picohttp library

Lots of users need the combination of core library, http library, and demo application.

Testing with `picotest_ct` requires building the picoquic library, test library and test app; testing with `picohttp_ct` requires the picoquic and picohttp libraries, and the picohttp_ct test app. There is probably not much to gain by treating those as separate targets.

The sample code is meant to be that, a sample that people read but do not actually use. It could be compiled separately.

I suppose that in a perfect make we would have several targets. I can think of `picoquic`, `demo`, `tests`, `sample` and `all`. The path of least surprise would be to make `all` the default, and have users who only care for one target specify that target.

CMakeLists.txt
```diff
@@ -1,20 +1,21 @@
...
-    set(CMAKE_C_FLAGS "-std=c99 -Wall -O3 -fno-exceptions -fno-signed-zeros -fno-trapping-math ${CC_WARNING_FLAGS} ${CMAKE_C_FLAGS}")
-else()
-    set(CMAKE_C_FLAGS "-std=c99 -Wall -O3 -fno-exceptions -fno-signed-zeros -frename-registers -fno-trapping-math ${CC_WARNING_FLAGS} ${CMAKE_C_FLAGS}")
-endif()
+# Set compile options below
+#if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+#    set(CMAKE_C_FLAGS "-std=c99 -Wall -O3 -fno-exceptions -fno-signed-zeros -fno-trapping-math ${CC_WARNING_FLAGS} ${CMAKE_C_FLAGS}")
+#else()
+#    set(CMAKE_C_FLAGS "-std=c99 -Wall -O3 -fno-exceptions -fno-signed-zeros -frename-registers -fno-trapping-math ${CC_WARNING_FLAGS} ${CMAKE_C_FLAGS}")
+#endif()
```
huitema, line 19: Would be nice to test for "Clang OR AppleClang". The issues come from `-frename-registers`, which I think is GCC only.

CMakeLists.txt
```diff
@@ -24,6 +25,8 @@ include(CheckCCompilerFlag)
 include(CheckCXXCompilerFlag)
 include(CMakePushCheckState)
 
+# PEJ - I noted there are CFLAGS set here. These will trickle down to picotls
+#       unless moved to target compile options.  What's preferred?
 if(ENABLE_ASAN)
```
huitema, line 30: Having the flags down to Picotls is probably fine. The ASAN and UBSAN flags are typically used in test builds, before running a set of tests, stress or fuzz cases to detect possible bad pointers, leaks, etc. Not a bad idea to also detect potential issues in Picotls.

CMakeLists.txt
```diff
@@ -198,124 +193,197 @@ else()
 	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DPTLS_WITHOUT_FUSION")
 endif()
 
-find_package(OpenSSL )
+find_package(OpenSSL)
 message(STATUS "root: ${OPENSSL_ROOT_DIR}")
```
huitema, line 197: Should this be REQUIRED ?

CMakeLists.txt
```diff
@@ -198,124 +193,197 @@ else()
...
+            CXX_STANDARD 11
+            CXX_STANDARD_REQUIRED YES
+            CXX_EXTENSIONS YES)
+    target_compile_options(${ARGV0}
+        PRIVATE
+            $<$<OR:$<C_COMPILER_ID:Clang>,$<C_COMPILER_ID:AppleClang>>: -O3 -Wall -fno-exceptions -fno-signed-zeros -fno-trapping-math>
+                   $<$<C_COMPILER_ID:GNU>: -O3 -Wall -fno-exceptions -fno-signed-zeros -frename-registers -fno-trapping-math>
+                   $<$<C_COMPILER_ID:MSVC>: >)
+endmacro()
+
```
huitema, line 234: If the macros make the overall text easier to read and maintain, we should go for it.

I am not sure that I remember why we have C99 in the current build. Probably copied from the Picotls CMakeList.txt 5 years ago.

CMakeLists.txt
```diff
@@ -198,124 +193,197 @@ else()
...
-    ${OPENSSL_LIBRARIES}
-    ${CMAKE_DL_LIBS}
-    ${CMAKE_THREAD_LIBS_INIT}
-)
+set_picoquic_build_options(picolog_t)
 
-target_include_directories(picoquic_ct PRIVATE loglib)
+add_executable(picoquic_ct
+    picoquic_t/picoquic_t.c
+    ${PICOQUIC_TEST_LIBRARY_FILES})
```
huitema, line 326: I know it is written like that in the current CMakeList, but this leads to compiling the test libraries twice, for `picoquic_ct` and for `picohttp_ct`.

CMakeLists.txt
```diff
@@ -198,124 +193,197 @@ else()
...
+        picoquic-core)
 
 target_include_directories(picoquic_sample PRIVATE loglib picoquic)
 
+set_picoquic_build_options(picohttp_ct)
+
+# PEJ - Should we introduce ctest and a test build option so that
+# downstream code will not have these tests built and so that
+# "make test" works as expected?
 set(TEST_EXES picoquic_ct picohttp_ct)
```
huitema, line 387: Yes.

But also, consider that the test programs of the upstream code may end up using the test libraries of picoquic.

huitema (commented):
Looks very promising, thanks for working on that.

paulej:
PS - I modified many of the workflow scripts this time, but not ci-asan-ubsan.yml.  I didn't want to break anything messing with that one.

CMakeLists.txt
```diff
@@ -198,125 +206,208 @@ else()
...
+        ${OPENSSL_INCLUDE_DIR}
+    PUBLIC
+        picoquic)
+
+set_picoquic_build_options(picohttp-core)
+
+include(CTest)
+
+if(BUILD_TESTING AND picoquic_BUILD_TESTS)
+
```
huitema, line 304: I wonder whether we need to do this kind of selection using CMake options, or whether we should instead use CMake to define a series of individual "make" targets, such as defining a "core" target with the essential libraries, a "test" target, a "demo", a "sample" target, and maybe an "all" target for compatibility with the current behavior.

When building through fetch, don't we have the opportunity to do something like this?
```
cmake build
cd build
make core
```

CMakeLists.txt
```diff
@@ -198,125 +206,208 @@ else()
...
+            #loglib
+            picoquictest)
+
+    set_picoquic_build_options(picoquic_ct)
+
+    add_executable(picohttp_ct
+        picohttp_t/picohttp_t.c
+        ${PICOQUIC_TEST_LIBRARY_FILES}
+        ${PICOHTTP_TEST_LIBRARY_FILES})
+
```
huitema, line 370: We should replace " ${PICOQUIC_TEST_LIBRARY_FILES}" by a "picoquic-test" library, which would be compiled once, rather than compiling the test code twice.

huitema (commented):
This is progressing nicely!

CMakeLists.txt
paulej: It is possible to do "make picoquic-core".  It will build it and any dependencies.

CMakeLists.txt
paulej: Good idea. I'll work on that.

paulej:
I made those additional enhancements.  I also moved the compile definitions and options out of `ENABLE_UBSAN` and `ENABLE_ASAN` into the compiler settings macro.  I left the original stuff still in the file commented out so you can see what I did.

It is normal to use both `-fsanitize=undefined` and `-fsanitize=address` at the same time?  That is how it was and so I kept the same order.

Sadly, I ended up staying up the whole night when suddenly things broke.  I thought I broke things, though I don't know how I could have broken things by just modifying the CMake file.  I finally discovered that a picotls commit made a few hours ago broke things. :(

I changed the FetchContent routine to pull commit 74feba44643ed905317be7a64a689f0d49120860.  With that, the tests passed again.

IvanNardi:
> It is normal to use both `-fsanitize=undefined` and `-fsanitize=address` at the same time? That is how it was and so I kept the same order.

It is *convenient* to use these two sanitizers at the same time since they are compatible with each other (unlike same others sanitizers) and so you can check both classes of errors with only one run.

paulej:
@huitema I think this is done.  I cleaned it up a bit more (removing my comments, adjusting spacing, etc.); let me know if there is more (or less) you'd like me to do.

huitema (approved):

huitema:
Thanks for all that, Paul!

paulej:
@huitema I made a mistake when I went to update a downstream project.  I now see that picolog_t and picoquicdemo are in tie install targets list.  I either need to move them out of the install targets list _or_ I need to have them built even if `picoquic_BUILD_TESTS` is `OFF`.  Which would you prefer?

huitema:
`picolog_t` is widely used. If a test crashes, the output dir will contain logs in binary ".log" format, which `picolog_t` can convert in readable `qlog` format. Most users who install `picoquic` will also want to install `picolog_t` and `picoquicdemo`. Having them in the install target is fine. They do not depend on the test library.

paulej:
ACK.  I'll prepare one more (much simpler) PR.

PR #1389, opened 2022-07-22
Align to picotls as of 7/21/22

huitema:
One of the recent Picotls commit reduces the size of the "Client Hello", by removing the negotiation of obsolete draft versions.
The change of size causes a minor change in the log file. This PR updates the reference files accordingly.


paulej (approved):
I built this branch stand-alone and as part of a downstream project. Code builds.  Tests pass on my machine, too.

PR #1390, opened 2022-07-22
Move certain targets so they build when tests are disabled

paulej:
This will ensure `picoquicdemo` and `picolog_t` are built even if tests are not built.  This is important since they are part of the INSTALL TARGETS.


huitema (approved):
Looks good. I should have caught that before.

PR #1391, opened 2022-07-26
Linux IP_MTU_DISCOVER/IPV6_MTU_DISCOVER

msvoelker:
On Linux, the default value for socket option IP_MTU_DISCOVER (given that sysctl net.ipv4.ip_no_pmtu_disc=0) or IPV6_MTU_DISCOVER is IP_PMTUDISC_WANT. With that, Linux's IP implementation fragments outgoing packets larger than the route MTU (as shown by ip route get DST-IP). Linux sets the route MTU upon an incoming ICMP Packet Too Big (PTB) message. Since picoquic does not process PTBs, a PTB reporting a smaller MTU leads to IP fragmentation for QUIC packets.

As picoquic already does PMTUD, I kind of disable PMTUD on IP level by setting IP_MTU_DISCOVER or IPV6_MTU_DISCOVER to IP_PMTUDISC_PROBE. With that, Linux ignores the route MTU.

### DIFF
--- a/picoquic/picosocks.c
+++ b/picoquic/picosocks.c
@@ -318,6 +318,9 @@ int picoquic_open_server_sockets(picoquic_server_sockets_t* sockets, int port)
             if (ret == 0) {
                 ret = picoquic_bind_to_port(sockets->s_socket[i], sock_af[i], port);
             }
+            if (ret == 0) {
+                ret = picoquic_socket_set_pmtud_options(sockets->s_socket[i], sock_af[i]);
+            }
         }
     }
 
--- a/picoquic/picosocks.c
+++ b/picoquic/picosocks.c
@@ -240,7 +240,7 @@ int picoquic_socket_set_ecn_options(SOCKET_TYPE sd, int af, int * recv_set, int 
     return ret;
 }
 
-int picoquic_socket_set_pmtud_options(SOCKET_TYPE sd, int af)
+int picoquic_socket_set_pmtud_options(SOCKET_TYPE sd)
 {
     int ret = 0;
 #if defined(IP_MTU_DISCOVER) && defined(IP_PMTUDISC_PROBE)
@@ -270,7 +270,7 @@ SOCKET_TYPE picoquic_open_client_socket(int af)
         if (picoquic_socket_set_ecn_options(sd, af, &recv_set, &send_set) != 0) {
             DBG_PRINTF("Cannot set ECN options (af=%d)\n", af);
         }
-        if (picoquic_socket_set_pmtud_options(sd, af) != 0) {
+        if (picoquic_socket_set_pmtud_options(sd) != 0) {
             DBG_PRINTF("Cannot set PMTUD options (af=%d)\n", af);
         }
     }
@@ -319,7 +319,7 @@ int picoquic_open_server_sockets(picoquic_server_sockets_t* sockets, int port)
                 ret = picoquic_bind_to_port(sockets->s_socket[i], sock_af[i], port);
             }
             if (ret == 0) {
-                ret = picoquic_socket_set_pmtud_options(sockets->s_socket[i], sock_af[i]);
+                ret = picoquic_socket_set_pmtud_options(sockets->s_socket[i]);
             }
         }
     }
--- a/picoquic/picosocks.c
+++ b/picoquic/picosocks.c
@@ -240,12 +240,17 @@ int picoquic_socket_set_ecn_options(SOCKET_TYPE sd, int af, int * recv_set, int 
     return ret;
 }
 
-int picoquic_socket_set_pmtud_options(SOCKET_TYPE sd)
+int picoquic_socket_set_pmtud_options(SOCKET_TYPE sd, int af)
 {
     int ret = 0;
-#if defined(IP_MTU_DISCOVER) && defined(IP_PMTUDISC_PROBE)
+#if (defined(IP_MTU_DISCOVER) || defined(IPV6_MTU_DISCOVER)) && defined(IP_PMTUDISC_PROBE)
     int val = IP_PMTUDISC_PROBE;
-    ret = setsockopt(sd, IPPROTO_IP, IP_MTU_DISCOVER, &val, sizeof(int));
+    if (af == AF_INET6) {
+        ret = setsockopt(sd, IPPROTO_IPV6, IPV6_MTU_DISCOVER, &val, sizeof(int));
+    }
+    else {
+        ret = setsockopt(sd, IPPROTO_IP, IP_MTU_DISCOVER, &val, sizeof(int));
+    }
 #endif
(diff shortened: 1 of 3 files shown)


picoquic/picosocks.c
```diff
@@ -240,6 +240,21 @@ int picoquic_socket_set_ecn_options(SOCKET_TYPE sd, int af, int * recv_set, int
...
+#if (defined(IP_MTU_DISCOVER) || defined(IPV6_MTU_DISCOVER)) && defined(IP_PMTUDISC_PROBE)
+    int val = IP_PMTUDISC_PROBE;
+    if (af == AF_INET6) {
+        ret = setsockopt(sd, IPPROTO_IPV6, IPV6_MTU_DISCOVER, &val, sizeof(int));
+    }
+    else {
+        ret = setsockopt(sd, IPPROTO_IP, IP_MTU_DISCOVER, &val, sizeof(int));
+    }
+#endif
+    return ret;
```
huitema, line 255: This does not work. If the `#if` test is not successful, the function becomes empty, the variables `sd` and `af` are not used, and the compiler creates a warning treated as error:
```
2>c:\projects\picoquic\picoquic\picosocks.c(243): error C2220: warning treated as error - no 'object' file generated [C:\projects\picoquic\picoquic\picoquic.vcxproj]
     2>c:\projects\picoquic\picoquic\picosocks.c(243): warning C4100: 'af': unreferenced formal parameter [C:\projects\picoquic\picoquic\picoquic.vcxproj]
     2>c:\projects\picoquic\picoquic\picosocks.c(243): warning C4100: 'sd': unreferenced formal parameter [C:\projects\picoquic\picoquic\picoquic.vcxproj]
```
Also, the test says `(defined(IP_MTU_DISCOVER) || defined(IPV6_MTU_DISCOVER))`, but then you call both the IPV6 and IPv4 variants. What if only one of those is defined?

huitema: Since the problem only seems to appear on Windows, one way to solve this would be:
```
    int ret = 0;
#if (defined(IP_MTU_DISCOVER) || defined(IPV6_MTU_DISCOVER)) && defined(IP_PMTUDISC_PROBE)
    int val = IP_PMTUDISC_PROBE;
    if (af == AF_INET6) {
        ret = setsockopt(sd, IPPROTO_IPV6, IPV6_MTU_DISCOVER, &val, sizeof(int));
    }
    else {
        ret = setsockopt(sd, IPPROTO_IP, IP_MTU_DISCOVER, &val, sizeof(int));
    }
#else
#ifdef _WINDOWS
    UNREFERENCED_PARAMETER(af);
    UNREFERENCED_PARAMETER(sd);
#endif
#endif
```

huitema (commented):
Thanks for the PR. There is just a minor fix needed to make it compile properly on Windows

picoquic/picosocks.c
msvoelker: > This does not work. If the #if test is not successful, the function becomes empty, the variables sd and af are not used, and the compiler creates a warning treated as error:

Oh, right. That's ugly.

> Also, the test says (defined(IP_MTU_DISCOVER) || defined(IPV6_MTU_DISCOVER)), but then you call both the IPV6 and IPv4 variants. What if only one of those is defined?

That would be a problem. I tried to save lines. If you don't mind to have more if statements, I would write:

```
#ifndef __linux__
    UNREFERENCED_PARAMETER(sd);
    UNREFERENCED_PARAMETER(af);
    return 0;
#else
    if (af == AF_INET6) {
#if defined(IPV6_MTU_DISCOVER) && defined(IP_PMTUDISC_PROBE)
        int val = IP_PMTUDISC_PROBE;
        return setsockopt(sd, IPPROTO_IPV6, IPV6_MTU_DISCOVER, &val, sizeof(int));
#else
        UNREFERENCED_PARAMETER(sd);
        return -1;
#endif
    }
    else {
#if defined(IP_MTU_DISCOVER) && defined(IP_PMTUDISC_PROBE)
        int val = IP_PMTUDISC_PROBE;
        return setsockopt(sd, IPPROTO_IP, IP_MTU_DISCOVER, &val, sizeof(int));
#else
        UNREFERENCED_PARAMETER(sd);
        return -1;
#endif
    }
#endif /* #ifndef __linux__ */
```

> Since the problem only seems to appear on Windows

I believe IP_MTU_DISCOVER is a special Linux thing.

huitema: The code compiles on MacOS, according to the tests reports, but I don't know whether it executes correctly. If we don't care about MAC OS, then we know that on Linux both IP_MTU_DISCOVER and IPV6_MTU_DISCOVER will be defined -- or not, depending on version. So without losing generality, we can simplify the code further. Something like:
```
    int ret = 0;
#if  defined __linux && defined(IP_MTU_DISCOVER) && defined(IPV6_MTU_DISCOVER) && defined(IP_PMTUDISC_PROBE)
    int val = IP_PMTUDISC_PROBE;
    if (af == AF_INET6) {
        ret = setsockopt(sd, IPPROTO_IPV6, IPV6_MTU_DISCOVER, &val, sizeof(int));
    }
    else {
        ret = setsockopt(sd, IPPROTO_IP, IP_MTU_DISCOVER, &val, sizeof(int));
    }
#else
    UNREFERENCED_PARAMETER(af);
    UNREFERENCED_PARAMETER(sd);
#endif  /* #ifndef __linux__ && ... */
```
That version would be much shorter.

msvoelker: Done

picoquic/picosocks.c
```diff
@@ -240,6 +240,24 @@ int picoquic_socket_set_ecn_options(SOCKET_TYPE sd, int af, int * recv_set, int
...
+    if (af == AF_INET6) {
+        ret = setsockopt(sd, IPPROTO_IPV6, IPV6_MTU_DISCOVER, &val, sizeof(int));
+    }
+    else {
+        ret = setsockopt(sd, IPPROTO_IP, IP_MTU_DISCOVER, &val, sizeof(int));
+    }
+#else
+    UNREFERENCED_PARAMETER(af);
+    UNREFERENCED_PARAMETER(sd);
+#endif  /* #if defined __linux && ... */
```
huitema, line 259: Damn. This doesn't work on MacOS:
```
/Users/runner/work/picoquic/picoquic/picoquic/picosocks.c:255:5: error: implicit declaration of function 'UNREFERENCED_PARAMETER' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
    UNREFERENCED_PARAMETER(af);
    ^
1 error generated.
```

huitema (commented):
Proposing a fix for MacOS

huitema (approved):

huitema:
Thanks for working on that, and sorry for issues with multiple platforms.

Issue #1396, opened 2022-08-15
BUG: Receipt of NEW_TOKEN by a server is not handled correctly


bathooman:
### Description
- Type: Non-conformance Bug
- Priority: Medium

---------------------------------------------------------------
## Non-conformance Bug
Commit: 7f2fbdfb9f1f7431b31c04fb5d9b4d30b560540b

**Expected behavior**   
The QUIC 1 [RFC](https://www.rfc-editor.org/rfc/rfc9000.html) in section 19.7 states that:

> Clients MUST NOT send NEW_TOKEN frames. A server MUST treat receipt of a NEW_TOKEN frame as a connection error of type PROTOCOL_VIOLATION.

**Actual behavior** 
When a client and a server are exchanging packets, receipt of a NEW_TOKEN frame will not trigger any type of error by the server. In the following, the server is running on port 20220, and the client is running on port 60581. As one can see, when the server receives an NT, it does not generate any error.

![image](https://user-images.githubusercontent.com/3464382/184635947-5b3deb88-a261-418b-bd07-5c7d925d82dc.png)

I have attached the captured interaction between a QUIC client and a server containing the mentioned problem. The master secret log file is also attached so that one can decrypt the exchanged traffic.

[client-generated-new-token.zip](https://github.com/private-octopus/picoquic/files/9337180/client-generated-new-token.zip)

huitema:
Verified. The parsing code merely skips over these frames if received by a server, should trigger an error.

huitema:
Fixed in Pr #1398

Issue #1397, opened 2022-08-15
BUG: Negative computed packet number in ACK frames


bathooman:
### Description
- Type: Non-conformance Bug
- Priority: Medium

## Non-conformance Bug
Commit: 7f2fbdfb9f1f7431b31c04fb5d9b4d30b560540b

**Expected behavior **   
The QUIC 1 [RFC](https://www.rfc-editor.org/rfc/rfc9000.html) in section 19.3.1 states that:

> Each ACK Range acknowledges a contiguous range of packets by indicating the number of acknowledged packets that precede the largest packet number in that range. A value of 0 indicates that only the largest packet number is acknowledged. Larger ACK Range values indicate a larger range, with corresponding lower values for the smallest packet number in the range. Thus, given a largest packet number for the range, the smallest value is determined by the following formula: smallest = largest - ack_range.
If any computed packet number is negative, an endpoint MUST generate a connection error of type FRAME_ENCODING_ERROR.

**Actual behavior ** 
 During an interaction, upon receiving an ACK frame, if the computed packet number is negative, in some scenarios, you can complete the handshake without any error being generated.

![image](https://user-images.githubusercontent.com/3464382/184652294-5757fc0a-e5be-49fa-8570-0de70e786162.png)

I have attached the captured interaction between a QUIC client and a server containing the mentioned problem. The master secret log file is also attached so that one can decrypt the exchanged traffic.

[negative-acked-packet-number.zip](https://github.com/private-octopus/picoquic/files/9337739/negative-acked-packet-number.zip)

huitema:
Thanks for the report, @bathooman.

This is weird. There is already a unit test for checking the parsing of "bad" ACK frames. I added another test case with an exact copy of the frame above, and the parser does return the expected "frame encoding" error.

There may be something else going on. As the handshake progresses, the picoquic nodes progressively close the "initial" and then "handshake" packet ranges, and just ignore any further incoming packets in that range. Could it be that your test packet was just ignored?

This is a bit difficult to debug with the wireshark captures. Could you reproduce that while capturing "qlog" files on the picoquic node under test? The qlog files will include information about the local state of the program, which of course are not present in the PCAP.

bathooman:
Hello,
I have checked it again. In presence of bad ACK frames, we get back PROTOCOL_VIOLATION.

I attached a new trace where two of the acknowledgements sent to the server are invalid, along with the qlog files.

[...]

PR #1398, opened 2022-08-16
Check frame parsing and add tests

huitema:
Verify that parsing bad frames produces the expected errors, and in particular fixes the issues #1395 and #1396.
Tighten the "skip frames" and "parse frames" tests, to check that error codes on bad frames are as expected.

Close(#1395)
Close(#1396)


huitema:
@bathooman, you may want to verify that this fixes your issues.

bathooman:
The PR seems to solve the problems. (#1395 #1396)

huitema:
Issue #1397 is also fixed in this PR -- ACK parsing with bad ranges is tested in the "skip frames" and "parse frames" unit tests, and the qlog rendering of ACK frames has been updated to report negative numbers when those bad frames are encountered.

Close(#1397)

PR #1399, opened 2022-08-28
Ensure Dockerfile works with current picotls

pyfisch:
Upgrade Ubuntu to 20.04 and set timezone non-interactively.

Otherwise the following cc errors occur:
```
cc: error: unrecognized command line option '-mvaes'; did you mean '-maes'?
cc: error: unrecognized command line option '-mvpclmulqdq'; did you mean '-mpclmul'?
```

In addition the version of CMake in Ubuntu 18.04 is too old so another error would occur while building picotls.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,8 @@
-FROM ubuntu:18.04
+FROM ubuntu:20.04
 
-RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


huitema:
Thanks for fixing this!

PR #1416, opened 2022-12-05
Initial test of "warp" mode.

huitema:


huitema:
There was a small hurdle when processing this PR. On some actions, GitHub insisted for require the version3 of action/checkout. This apparently pulled in SSL v3, which surfaced a memory leak. Took a lot of trial and errors to understand where that came from. Finally traced it a call to `ERR_load_crypto_strings()`. That API is deprecated in OpenSSL version > 1.1; in version 3, it causes a memory allocation that is not fixed by calling `ERR_free_strings`. Code is now fixed, and the LeakTest reports no leaks when running OpenSSL v3.

The leak itself is not a huge issue: 518 byte(s) leaked in 9 allocation(s), once per process. It would be nice to have an execution of the LeakTest with OpenSSL 1.1, and maybe also with LibreSSL, but that may not be urgent.

huitema:
The ASAN-UBSAN test passes, but the Valgrind tests fixes an issue -- apparently, an occurrence of decryption of uninitialized memory. Need another series of trying test ranges to isolate the test case in which that occurs.

huitema:
Found an actual error in the code. When packets were received and the decryption key was not available, picoquic stores them in a temporary buffer. However, the code was wrongly copying the buffered bytes from the results of the decoding attempt, which was not initialized. Fixing that ensures that when packets are dequeued from buffers, they are properly decrypted.

huitema:
The number of valgrind failures appears correlated with the number of packets decrypted. Trying simpler tests to try isolate the code segment causing this.

huitema:
Checking this now, as fixing the valgrind issue takes precedence over further developing the warp tests.

PR #1421, opened 2022-12-12
Update QUICv2 final constants

IvanNardi:


huitema (approved):
Thanks for doing this!

PR #1425, opened 2023-01-09
Update protection vector V2

huitema:
Update the definition of the retry protection secret for QUIC V2, and check the corresponding test vector.


IvanNardi (approved):
LGTM!
Sorry for missing that in the previous PR

PR #1426, opened 2023-01-09
Add support for V2 draft version

huitema:
Support V2 draft version in addition to V2 IESG approved version. This is required to continue working with the interop-runner tests.


huitema:
@IvanNardi I reinstalled support for the old V2 draft version, in addition to the latest one, because that old version is still used in the interop tests. It probably would be nice to just get rid of all these intermediate versions, but keeping them is not too expensive.

IvanNardi (approved):

PR #1427, opened 2023-01-11
Implementation of Web Transport in Picoquic

huitema:
Work in progress.


huitema:
At this point, we have demonstrated that it is possible to develop a web transport application in picoquic. The "baton_app" exercises a lot of the web transport features: create and use unidirectional and bi directional streams in both directions. There are obvious next steps:

* validate through interop testing,
* derive a unit test of stream creation and removal based on the baton app and make it part of picohttp tests
* split the baton application into a web-transport API and the baton application
* possibly, extract an "h3zero_client" from the democlient, and unify with the data types used in "h3zero_server" and web transport.

huitema:
The first test verifies memory leaks and other issues. There are a couple of other useful tests, such as what happens if the server closes the session first, what if there is an unanticipated connection break, what about receiving resets or stop sending, what if either client or server silently departs.

huitema:
At this point, the baton sample for web transport looks reasonably good. This is a lot of code, so it might be better to just check it in, and create a new PR after interop testing.

PR #1428, opened 2023-01-17
Delay tolerant networking with Picoquic

huitema:
This branch is created to test Picoquic over "delay tolerant networks", with latencies expressed in minutes.

The goal is to test at least three scenarios:

1) Basic test, establish a connection and then request a small response,
2) Data test, i.e., same as basic but larger data size and some packet loss, so as to retransmission issues tied to high latencies. Consider using Cubic.
3) Same as 2, but using BBR.

We may want to do several versions of these tests, with latency set to 1 minute, 2 minutes, 10 minutes, etc.

This is largely work in progress.


huitema:
The first checking installs a very basic test: establish a connection and then request a small response, with a delay set to 1 minute. The only configuration needed is to set the "idle timer" to a large value -- in our case, 5 times the target RTT. The only changes so far in the picoquic stack are related to using this idle timer value instead of the "maximum handshake time".

This all works, but it is gross. The "packet lost" timers fire left and right, initial packets are repeated multiple times, which leads to repeating handshake packets multiple. The full handshake succeeds after 4 minutes, as expected, but at that point clients and server have each sent 240 packets -- a bit more than the usual 4 or 5. The data exchange succeeds after 6 minutes, as expected, but we see quite a lot of spurious messages. The connection closes cleanly after 8 minutes.

Next step is to understand how to avoid all that overhead!

huitema:
The second commit programs longer timers during the handshake when the "idle timeout" is set to a large value. During the handshake, the timer is set to min(initial_RTT << nb_repeats, idle_timeout/10). With that, the initial packet is repeated about 16 times, instead of more than 200 in the previous version. This is a compromise between several priorities:

* Do not assume that we know the actual latency at the beginning of the connection. If the latency is actually low, spacing the repeats too much is counterproductive.
* Have enough repetitions to survive high packet loss rates. We have interop tests with loss rate of 30%, and to get 5 nines in such conditions we have to allow for 10 repetitions during the delay allocated to the handshake
* Still limit the number of packets sent.

huitema:
There is a bit of a side effect that could be eliminated. The server sends a response to each repeated Initial packet, in which it repeats there the "server handshake" packet. That's expected. But we also see the client acknowledging each of these packets. This is a bit of a waste. at a minimum, these acknowledgements could be spaced out.

PR #1430, opened 2023-01-23
Fix minor compile warning

paulej:


huitema (approved):
Oops! Thanks for fixing that.

PR #1447, opened 2023-02-21
Try fix pkg config issue

huitema:


huitema:
@svidzinsky One of the CI checks in your PR was not passing. Nothing to do with your fix, the MacOS test was failing because pkg-config is not installed by default in the latest versions of MacOS. This fixes the issue.

Issue #1451, opened 2023-02-27
Amplification of connection migration attacks


huitema:
An [NDSS symposium paper by Yuri Gbur and Florian Tschorsch](https://www.ndss-symposium.org/wp-content/uploads/2023/02/ndss2023_s72_paper.pdf) shows that picoquic can amplify a connection migration attack. In this attack, an attacker establishes a regular QUIC connection with a server, and then uses secrets acquired when setting up that connection to send forged packets to the server. These packets could be of two forms:

* NAT variant: simulate a NAT traversal by sending regular 1-RTT packets using the current CID but replacing the source address by the address of a target;
* Challenge variant: simulate a connection migration by sending a packet containing a "source challenge" with a new CID from the spoofed source address of a target.

The focus of the NDSS paper was the "request forgery" attack in which a client tricks a server to send packets to a third party, running a different application. The connection migration attacks are not very powerful there, because only up to 20 bytes of the packet sent by the server can be predicted by the attacker -- and the server will normally refuse to send such packets to "suspicious addresses" per PR #1450. But the method can still be used to generate "volumetric" attacks against a selected target.

In the challenge variant, the current picoquic server will respond with a challenge of its own, in a packet padded to the minimum required 1200 bytes. As the attacker will not acknowledge this packet, picoquic will repeat it a couple of times. We have an amplification factor 3: the attacker sends a 1200 bytes challenge, picoquic repeats the challenge + response packet three times. In theory, this is not very interesting for the attacker, because it also had to send a couple of full size packets to set up the connection, and then a couple of short packets to keep it alive. However, nothing prevents the attacker from creating many challenges on a single connection.

In the NAT variant, the server that detects a packet that seems affected by NAT rebinding is supposed to send a full size challenge to the "natted" address, and also a parallel challenge to the original address. If the challenge to the original address succeeds, the NAT traversal detection is considered spurious and the server uses the old address. If neither challenge succeeds, the challenge will be repeated three times. The packet triggering all that could be very short, so that's a lot of amplification. The attacker would reply to the third repeat of the "original source" challenge to keep the connection alive, and could then repeat the process. This could turn in a reasonably powerful attack.

[...]

PR #1452, opened 2023-02-27
Mitigate connection migration amplification

huitema:
This PR streamlines the processing of Path Challenge and Path Response frames. It maintains a count of bytes received on a path, and use that count to limit the size of Path Challenge and Path Response packets, not applying padding if that would bypass amplification limits.


huitema:
With the limitation of repetition of challenge responses, this is now a full respone. We can close #1451

PR #1454, opened 2023-03-13
adding some basic CLI usage info to help bootstrap picoquic use

marcblanchet:
- took me some time to find out, with the help of Christian!
- not sure what picohttp_ct does since its -h is not useful.
- location of binaries are for linux. might need to edit for other platforms.

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -5,7 +5,7 @@ On Linux after building, the following executables are available in the project 
 * picoquic_ct: runs various QUIC tests
 	* with no arguments, runs all tests.
 * picoquicdemo: QUIC/HTTP demo client and server for HTTP/3, HTTP/0.9, QUIC performance tests and Siduck(simple ping-pong)
-* picoquic_sample: QUIC/HTTP demo client and server. See sample/README.md for detailed usage
+* picoquic_sample: QUIC/HTTP sample demonstrating how to write an application using the picoquic stack. It is not meant to be actually used. See sample/README.md for detailed usage
 
 ## HTTP Client and Server Usage
 * client: ./picoquicdemo servername serverportnumber HTTPpath
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -4,7 +4,7 @@ On Linux after building, the following executables are available in the project 
 * picohttp_ct: runs various HTTP tests
 * picoquic_ct: runs various QUIC tests
 	* with no arguments, runs all tests.
-* picoquicdemo: QUIC/HTTP demo client and server for HTTP/3, HTTP/0.9, QUIC performance tests and Siduck(simple ping-pong)
+* picoquicdemo: QUIC/HTTP demo client and server for HTTP/3, HTTP/0.9, QUIC performance tests and Siduck(simple test of Datagram support)
 * picoquic_sample: QUIC/HTTP sample demonstrating how to write an application using the picoquic stack. It is not meant to be actually used. See sample/README.md for detailed usage
 
 ## HTTP Client and Server Usage
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -1,0 +1,22 @@
+# Picoquic CLI Usage
+On Linux after building, the following executables are available in the project root directory. All support "-h" argument to get the detailed usage:
+
+* picohttp_ct: runs various HTTP tests
+* picoquic_ct: runs various QUIC tests
+	* with no arguments, runs all tests.
+* picoquicdemo: QUIC/HTTP demo client and server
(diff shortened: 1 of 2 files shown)


doc/usage.md
```diff
@@ -0,0 +1,22 @@
...
+
+## HTTP CLI
+* client: ./picoquicdemo servername serverportnumber HTTPpath
+  * downloaded files will be in current directory. use -o for another folder 
+  * exemple: ./picoquicdemo -o ../received 192.0.2.1 4433 index.html
+* server: ./picoquicdemo -p portnumber
+  * -p argument tells the program that it is acting as a server
+  * HTML root folder is current directory by default. use -w for another folder
+  * exemple: ./picoquicdemo -w ../htmlroot -p 4433
+
```
huitema, line 18: Technically, this is a "demo program", not just an "http cli". It supports the following application protocols:

* HTTP/3 -- the version of HTTP evolved to take advantage if QUIC, ALPN="h3"
* HTTP/0.9 -- the original version of HTTP, characterized by ALPN="hq-interop". this version is used in a number of introperability tests, e.g., [quic interop runner](https://interop.seemann.io/)
* QUIC Performance tests -- a simple protocol used to test performance of QUIC applications, ALPN="perf"
* Siduck -- a simple ping-pong protocol used to test support for datagrams, ALNP="siduck-00"

doc/usage.md
```diff
@@ -0,0 +1,22 @@
+# Picoquic CLI Usage
+On Linux after building, the following executables are available in the project root directory. All support "-h" argument to get the detailed usage:
+
+* picohttp_ct: runs various HTTP tests
+* picoquic_ct: runs various QUIC tests
```
huitema, line 5: This is the test program for the HTTP implementation that is part of the distribution. With no arguments, it runs all the HTTP tests.

huitema (commented):
Suggesting a couple of precisions

doc/usage.md
marcblanchet: I wanted something very short, straight to the point, for people to try it out. So I made some changes in the new commit, but did not go deep as you wrote. I would suggest that you put that very useful information in some other document describing the internals. But this is your repo, so you decide!

doc/usage.md
marcblanchet: ok. I was not sure what else to say. so did not do anything on the commit.

doc/usage.md
```diff
@@ -4,10 +4,10 @@ On Linux after building, the following executables are available in the project
 * picohttp_ct: runs various HTTP tests
 * picoquic_ct: runs various QUIC tests
 	* with no arguments, runs all tests.
-* picoquicdemo: QUIC/HTTP demo client and server
+* picoquicdemo: QUIC/HTTP demo client and server for HTTP/3, HTTP/0.9, QUIC performance tests and Siduck(simple ping-pong)
 * picoquic_sample: QUIC/HTTP demo client and server. See sample/README.md for detailed usage
```
huitema, line 9: The "sample" code is really just that -- a code sample demonstrating how to write an application using the picoquic stack. It is not meant to be actually used.

marcblanchet: revised description based on your comment

doc/usage.md
```diff
@@ -4,10 +4,10 @@ On Linux after building, the following executables are available in the project
...
 
-## HTTP CLI
+## HTTP Client and Server Usage
 * client: ./picoquicdemo servername serverportnumber HTTPpath
```
huitema, line 11: There are a bunch of optional arguments that can be used by both client and server. You probably don't want to detail them all, but for the server it is important to understand setting encryption keys (-k) and setting certificate (-c). In any case, a pointer to using "-h" to get the list of options would be helpful.

huitema (commented):
There are other sources of documentation already, maybe link to them?

* https://github.com/private-octopus/picoquic/wiki/Testing-without-DNS-names-or-Certificates
* https://github.com/private-octopus/picoquic/wiki/Import-key-and-cert-on-server-from-Let's-encrypt
* https://github.com/private-octopus/picoquic/wiki/Running-picoquicdemo-with-absolute-path

doc/usage.md
marcblanchet: yeah, I've been thinking about adding some arguments, but way too many. That is why at the top of the file, I wrote: "All support "-h" argument to get the detailed usage". But I guess -k and -c are useful to recall explicitly.

marcblanchet: added wiki links in latest commit

huitema (commented):
Looks good, checking in.

PR #1456, opened 2023-03-22
Fix sample server FIN handling

mattcaswell:
If the FIN does not arrive at the same time as the rest of the stream data then the stream_ctx gets forgotten. When creating a new stream_ctx we need to call picoquic_set_app_stream_ctx.


huitema (approved):
Thanks for the fix. I agree that it fixes the context creation issue for a new stream. I am a bit concerned about what happens if a stream was accidentally closed, and I am also concerned that calling "reset stream" is pretty much a no op if the call back return -1 -- the stack is going to give up and close the connection. For a code sample, just "return -1" on error should be enough.

But that's probably for another day. I opened issue #1457 to track that.

PR #1462, opened 2023-04-04
Separate wt from baton

huitema:


huitema:
The goal of this PR is to provide a simple way to setup web transport applications. This requires setting an H3 client context, or adding parameters to an H3 server context. We want to use as much common code as possible between client and server -- same callback, in particular. ideally, we would like the "demo" code to also use this common code -- certainly on the server side, but even better if this works on the client side. This requires a fair bit of code refactoring, which is proving a bit arduous.

PR #1480, opened 2023-05-04
Add new API for providing datagram buffers

huitema:
The new extended API:
~~~
uint8_t* picoquic_provide_datagram_buffer_ex(void* context, size_t length, int is_active);
~~~
adds an "is_active" parameter to the previous version. See comments in `picoquic.h` explaining the behavior of the API.


picoquic/picoquic.h
```diff
@@ -1056,18 +1056,38 @@ int picoquic_mark_datagram_ready(picoquic_cnx_t* cnx, int is_ready);
...
+ * 
+ * - if the application marked the context ready by mistake, it 
+ *   should set the "length" argument to 0, and the "is_active" argument
+ *   to 0. This is a way of saying "oops". The stack will stop
+ *   polling for datagrams, until there is a new call to
+ *   "picoquic_mark_datagram_ready"
+ * 
+ * - if the application does have data to send but the available
+ *   length indicated in the callback is too small, it should set the "length"
+ *   argument to 0, and the "is_active" argument to 0. The stack will try to
```
suhasHere, line 1071: shouldn't is_active be 1 here ?

picoquic/frames.c
```diff
@@ -4579,19 +4579,35 @@ uint8_t * picoquic_format_first_datagram_frame(picoquic_cnx_t* cnx, uint8_t* byt
...
+        /* We apply the state change at this point, rather than after the return of the
+        * callback, so as to minimize "developer surprise". If the  application calls
+        * "picoquic_mark_datagram_ready" after this call, the vlaue set by the
+        * application will stick.
+        */
+        data_ctx->cnx->is_datagram_ready = is_active;
+    }
+
+    if (length <= data_ctx->allowed_space && length > 0) {
         /* Compute the length of header and length field */
```
suhasHere, line 4611: curious why not ``` (length > 0 && length <= data_ctx->allowed_space) ```

suhasHere (commented):
Looks good to me . couple of comments for my better undestanding

picoquic/picoquic.h
huitema: No. Setting "is_active" to 1 would start a hot loop.

huitema: Good catch! It should be 1 indeed.

picoquic/frames.c
huitema: Yes. "length == 0" happens in normal operation, while "length > data_ctx->allowed_space" is a programming mistake that normally never happens, so your proposed order will save some cycles. Fixed now.

suhasHere (approved):

Issue #1459, opened 2023-03-23
How to evaluate quality of several multi-path-connections? 


dongsf:
There are several multi-path connections to different servers in one client, and I want to evaluate quality of them before stream transmitting.
Whether it is reasonable using RTT values (like smoothed_rtt in st_picoquic_path_t）to evaluate the quality ?

I have tested the RTT values , but it seems not been updated after packet received. It was caused by one_way_delay_sample always be negative in my test case. 
Is it a result of clock drift things?

huitema:
Currently, streams data is spread over the available paths, according to what the congestion control of each individual path allows, and avoiding the paths that have too many packet losses. It does not directly use a classic measurement of path quality, only an indirect one. We have internal data that could be exposed, namely:

* the pacing rate for the path, as measured by the congestion control algorithm;
* the RTT for the path -- but this needs work;
* the RTT variance for the path, which is an expression of delay jitter, another factor of quality;
* maybe some estimate of the loss rate for the path, which could be added as well.

Regarding RTT and RTT VAR, the one-way delay estimate is probably a bit too complex. To do it right, we would need to do multi-variable modelling, maybe approximated as a Kalman filter. We had lots of discussion of that as part of the standardization of MP QUIC (see for example this discussion on [how to send ACK MP and compute RTT](https://github.com/quicwg/multipath/issues/190)). This is a part of the code that could be simplified and made more robust, and I would like to do as soon as the multipath draft stabilizes -- probably in a couple of weeks. That would be a good time to add some API exposing path quality.

Just another question. You are asking to "evaluate quality of the paths before stream transmitting". Is this the same requirement as the "path affinity" API discussed in issue #1412 by @clarkzjw ?

clarkzjw:
> You are asking to "evaluate quality of the paths before stream transmitting". Is this the same requirement as the "path affinity" API discussed in issue https://github.com/private-octopus/picoquic/issues/1412 by @clarkzjw ?

If I understand correctly, they are similar. In "path affinity", essentially we also need to evaluate path qualities before associating paths and streams and possibly disassociating them when paths become unavailable https://github.com/private-octopus/picoquic/issues/1412#issuecomment-1416670798. It might be a good idea to expose those internal metrics as @huitema mentioned, e.g., pacing rate, RTT, loss rate, etc.

huitema:
[...]

Issue #1412, opened 2022-11-07
Stream and path affinity


clarkzjw:
Following https://github.com/private-octopus/picoquic/issues/1295#issuecomment-1288220135, I guess it's better to open a separate issue to discuss this.

So talking about the affinity of streams and paths, the motivation is to send specific streams over specific links in a multipath scenario. And maybe "links" or "interfaces" is a better alternative for "paths", since "path" in picoquic is a logical concept and new paths will be created if socket ports are different, even though the underlying physical links remain the same. 

My initial thoughts about implementing this are:
1. introduce a new `stream_ifindex` priority to `picoquic_stream_head_t` similar to `stream_priority`. And similar functions such as `int picoquic_set_stream_ifindex(picoquic_cnx_t* cnx, uint64_t stream_id, uint8_t ifindex)` is added to set the preferred interface index for streams.
2. In a multipath scenario, if the client calls `picoquic_set_stream_ifindex`, then the following packets belonging to this stream are sent out to the server via this interface, regardless of the path statistics, as long as there is any path available on this interface.
3. The server side remains the same, it just replies to the corresponding incoming IP and ports.

huitema:
I think that you should keep the current definition of paths. The interface index is a local variable, designating the network card. You can very easily have two different paths using the same interface index, if they use different IP addresses.

If I implemented this, I would add a "path_affinity" variable to the stream context, holding either NULL or a pointer to the selected path context. Something like:
```
int picoquic_set_stream_path(connection, stream, struct sockaddr * source address, struct sockaddr * remote address, interface)
```
With the code finding the path that matches the specified addresses.

Then, I would modify `picoquic_find_ready_stream` to add a path argument, specifying the path on which the current packet is being sent, and only selecting available streams if they have no affinity (NULL) or affinity set to the current path.

You will need some way to remove the path affinity if a path breaks, is demoted, etc.

huitema:
Of course, you have a separate problem of deciding paths on the server. I think it is best to decide that at the application level, rather than inside the picoquic code.

clarkzjw:
> I think it is best to decide that at the application level, rather than inside the picoquic code.

By this, are you referring to the decision on the server side or the whole stream affinity problem?

> You can very easily have two different paths using the same interface index, if they use different IP addresses.

True, I didn't think about that.

[...]

PR #1482, opened 2023-05-07
First cut at describing a new path API

huitema:
We want to develop a new set of path related API, per issues #1459 and #1412.

The first step here is to propose changes in "picoquic.h" describing the new API. The next stps will be to implement these changes.


clarkzjw (approved):
Looks good to me as describing new APIs.

alexrabi:
I'm going to go ahead and butt in to the conversation to give my two cents (since I am currently looking at similar problems in regards to picoquic).

In my use-case, I am primarily focusing on using the datagram API, and I am interested in doing per-datagram scheduling onto the different paths (i.e. using picoquic_callback_prepare_datagram to decide which datagram should be scheduled onto which path, if any at all). For this to work, the application would have to be aware of which path is considered for sending whenever picoquic_callback_prepare_datagram is called. I therefore suggest that the datagram callbacks also use the "stream_id" field to carry the "unique_path_id" similarly to the newly suggested callbacks, such that the application may make such decisions.

In terms of path quality information, I think that the proposed picoquic_path_quality_t structure is a good start. However, from my experience in scheduling for MPTCP and MP-DCCP (including an early MP-QUIC prototype), the information contained within this structure is currently far too course-grained for use-cases such as mine.  Ideally, as much information from the congestion control state as possible should be exposed (e.g. bytes_in_transmit, rtt_min, rtt_variant, and more).

huitema:
@alexrabi Good point about the datagram callback. Providing the path identifier in the callback is a very good idea.

huitema:
I am contemplating a different change. I would rather not have the "quality" callback change every time some information changes, because that could happen on every acknowledgement or maybe on every packet. To issue the callback less often, the stack needs to remember the values that it previously announced, and decide when there was a "significant" change. That means pretty much keeping a copy of the "quality" information in the path context. If there is such a copy, then it could be passed along the callback, say by repurposing the "bytes" and "length" parameters to provide a pointer to the up to date "quality" structure for that path, and its length. Would that be a good idea?

alexrabi:
@huitema 
> There is a related point concerning the scheduling of the call back. The API function picoquic_mark_datagram_ready states that the application is ready to send datagrams, presumably on any path. Should we add a path identifier there too?

A call to specify on which paths datagrams are ready to be sent sounds like a great idea; there could be many reasons why an application may prefer to wait for another path to become available rather than to send immediately. Currently there is no way to signal this from the application. This may need to be coupled with some change in `picoquic_select_next_path_mp` such that a path is "penalized" in the path selection when it is application limited, so that the path is not selected over and over again.

alexrabi:
@huitema I see your point regarding the stability of information in the public header. However, if it is expected for datagrams to be scheduled on paths at the application level, then non-stable, congestion control algorithm specific information can be very useful. For example, BBR can provide some useful information that can be used to keep the latency low, which other algorithms like New Reno and Cubic do not. Perhaps there could be some "advanced" API  function (non-callback related, keep that stable and generic) that the application may call in order to fetch such information?

huitema:
There is a little dilemma regarding how to handle streams that have no affinity specified. Suppose that we have two streams, S1 and S2, and two path, P1 and P2. The application sets affinity of S1 to any, and affinity of S2 to P2. I think that means:

* Part of S1 is sent on P1
* Part of S1 and S2 share path P2 according to scheduling defined by priorities -- S1 first if FIFO, S1 and S2 alternate if round robin?

In he round robin case, the stream S1 will be put back to the end of the queue whenever packets are sent on P1, so in practice S2 will get more of P2's capacity than S2.

I think that makes sense, but I would like a second opinion.

huitema:
Supporting stream affinity seems simpler than I thought. There is more testing to do, such as seeing what happens if there are two parallel streams with affinity to two paths, or if the affinity path breaks before the stream is done, but it looks simple enough:

* do not schedule streams on another path if the affinity is set
* select the affinity path as next path if the affinity stream is the next to be sent.

I need to complete these tests, and add something similar for datagrams. Bit of design work to do there, plus a fair amount of new test code.

huitema:
And with that, we now have path affinity support for datagrams as well!

picoquic/picoquic.h
```diff
@@ -708,23 +712,122 @@ void picoquic_delete_cnx(picoquic_cnx_t* cnx);
...
+int picoquic_probe_new_path_ex(picoquic_cnx_t* cnx, const struct sockaddr* addr_from,
+    const struct sockaddr* addr_to, int if_index, uint64_t current_time, int to_preferred_address);
+void picoquic_enable_path_callbacks(picoquic_cnx_t* cnx, int are_enabled);
+void picoquic_enable_path_callbacks_default(picoquic_quic_t* quic, int are_enabled);
+int picoquic_set_app_path_ctx(picoquic_cnx_t* cnx, uint64_t unique_path_id, void * app_path_ctx);
+int picoquic_abandon_path(picoquic_cnx_t* cnx, uint64_t unique_path_id, uint64_t reason, char const* phrase);
+int picoquic_get_path_quality(picoquic_cnx_t* cnx, uint64_t unique_path_id, picoquic_path_quality_t * quality);
+int picoquic_subscribe_to_quality_update(picoquic_cnx_t* cnx, uint64_t unique_path_id,
+    uint64_t pacing_rate_delta, uint64_t rtt_delta);
+void picoquic_default_quality_update(picoquic_quic_t* quic, uint64_t pacing_rate_delta, uint64_t rtt_delta);
```
huitema, line 819: Feedback from @suhasHere: monitoring quality is important in both single path and multipath applications. It would be nice if there was an API to "get the quality on the current default path", so we could have a unified behavior. This would allow us to deprecate the API "picoquic_subscribe_pacing_rate_updates" and the callback "picoquic_callback_pacing_changed".

huitema: This is addressed in the commit "create and test default path quality API", which:

* renames the previous `picoquic_subscribe_to_quality_update` to `picoquic_subscribe_to_quality_update_per_path`
* redefines `picoquic_subscribe_to_quality_update` to be for all the paths in the connection
* adds a new api `picoquic_get_default_path_quality`
* adds tests for the new API.

picoquic/frames.c
```diff
@@ -4601,10 +4622,16 @@ uint8_t* picoquic_provide_datagram_buffer_ex(void* context, size_t length, int i
...
         * application will stick.
+        * There are two active flag: global, if the application is ready to send datagrams
+        * on any stream, and per path, if the application wants to send datagrams
+        * again on that path.
         */
         data_ctx->cnx->is_datagram_ready = is_active;
+        if (data_ctx->path_x != NULL) {
+            data_ctx->path_x->is_datagram_ready = is_path_active;
+        }
     }
```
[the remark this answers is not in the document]
huitema: Path specific APi has been removed, the "is_active" argument of picoquic_provide_datagram_buffer_ex is now an enum.

PR #1492, opened 2023-06-02
ACK Frequency draft-04

huitema:
Update the code to support draft 04.

This implies: change of the Transport Parameter ID, change format of the ACK Frequency frame to include a reordering threshold, and addition of the "immediate ACK" frame.

The first commit does not include the immediate ACK frame support.


huitema:
Second commit implements the "receive" part of immediate ACK, and tests for it. The commit does not include any actual usage of immediate ACK by picoquic senders -- this option will have to come later, but will very likely disrupt a bunch of scenario tests, so it is better to leave it for another PR.

PR #1495, opened 2023-06-04
Fix IPv4 loopback address

alexrabi:
Seems like the IPv4 address used by the `picoquic_store_loopback_addr` helper function was incorrectly using 128.0.0.1 rather than 127.0.0.1. This will fix that.


huitema (approved):
Nasty typo! Thanks for fixing that!

PR #1496, opened 2023-06-04
Stress parameters in picohttp_t

huitema:
Rework the HTTP test `picohttp_t` so the parameters are easier to understand.
The http stress test was creating a server context with a fixed number of connections. Fixed that, so the number of expected connection matches the test.

### DIFF
--- a/picohttp_t/picohttp_t.c
+++ b/picohttp_t/picohttp_t.c
@@ -29,7 +29,8 @@
 #include <stdlib.h>
 
 extern size_t picohttp_nb_stress_clients;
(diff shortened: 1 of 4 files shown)


huitema:
There were several observed cases of the "http_corrupt" test failing, even when the number of clients was set to a low number. This was most probably caused by the arrival of packets with spoofed connection IDs, creating extra connection contexts that would then saturate the server. This is an issue with QUIC, not HTTP3, so the natural solution was to fix the test to not randomly corrupt the first bytes of long packet headers.

PR #1510, opened 2023-06-20
Ecn off

UCHIDAnobuhiro:


huitema:
@UCHIDAnobuhiro I see that you are trying to fix the use of ECN in QUIC. May I suggest that you open an issue? That would be a good way to have a discussion of problems and solutions.

Issue #1513, opened 2023-07-06
Having issues setting up Multipath with picoquicdemo


sureshpsu:
Hi,
   I am trying to setup a Realtime wireless systems with WiFi and WiGi as two paths. I am using ./picoquicdemo to stream data over both wifi and wigi but unfortunately it is not working (it just uses one path). Currently, I am using following commands:

on client: ./picoquicdemo -a perf -M 3 -A 192.168.10.245/5, 192.168.1.4/4 192.168.10.231 4433 "*1:0:-:397:5000000"
on Server: ./picoquicdemo -a perf -M 3 -A 192.168.10.231/3, 192.168.1.2/2

With the above setup, i am not able to stream QUIC data over both WiFi and WiGi in my setup. Please, advise me if i am doing some thing wrong?

Thanking you,
Warm regards,
Suresh

huitema:
Maybe ask @clarkzjw, who is using that option successfully?

I think you have two issues:

1) The parameters should be quoted, as in:
~~~
./picoquicdemo -a perf -M 3 -A "192.168.10.245/5, 192.168.1.4/4 192.168.10.231" 4433 "*1:0:-:397:5000000"
 ~~~
 2) The parameters are only applied on the client.

clarkzjw:
try quoting IPs and interface indexes on the client side
```
./picoquicdemo -a perf -M 3 -A "192.168.10.245/5,192.168.1.4/4" 192.168.10.231 4433 "*1:0:-:397:5000000"
```
and yes, the `-A` option only applies to the client.

sureshpsu:
Hi Jinwei Zhao,
                      I got it working, thanks for all the inputs. The issue was with the routing table, picoquic was not able to reach the server from the secondary link.....I fixed the routing issues and it started working.

I used tcpdump to debug and caught the issue.......

PS: Indeed, picoquic did send quic packets over the secondary link but it just could not route it to the server.

Thanking you,
Warm regards,
Suresh

PR #1517, opened 2023-07-07
Add quotes to description of -A parameter

huitema:
As seen in issue #1513, some users are confused by the syntax of the "-A" option. This PR tries to address the confusion by making the "usage" text in `picoquicdemo` more explicit:

* Use quotes around the example parameters, because many shells will apply special treatment to "," between addresses
* State explicitly that the "-A" parameter only applies to the client.

### DIFF
--- a/picoquicfirst/picoquicdemo.c
+++ b/picoquicfirst/picoquicdemo.c
@@ -1136,7 +1136,7 @@ void usage()
     fprintf(stderr, "  For the server mode, use -p to specify the port.\n");
     picoquic_config_usage();
     fprintf(stderr, "Picoquic demo options:\n");
-    fprintf(stderr, "  -A ip/ifindex[,ip/ifindex]  IP and interface index for multipath alternative path, e.g. 10.0.0.2/3,10.0.0.3/4\n");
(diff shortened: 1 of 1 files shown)


clarkzjw:
Do we want to support extra whitespaces in the multipath config option? Because currently `-A "10.0.2.2/2,  10.0.3.2/3"` won't work in https://github.com/private-octopus/picoquic/blob/master/picoquicfirst/picoquicdemo.c#L444

huitema:
@clarkzjw I think that we do not. The "usage" example do not include white spaces:
~~~
Picoquic demo options:
  -A "ip/ifindex[,ip/ifindex]"  IP and interface index for multipath alternative
                        path, e.g. "10.0.0.2/3,10.0.0.3/4". This option only
                        affects the behavior of the client.
~~~
Do you think that we should update the parsing code to support them?

clarkzjw:
Considering the demonstration purpose of `picoquicdemo`, I think we can leave it as is for now, unless we see the necessity in the future.

clarkzjw (approved):

Issue #1499, opened 2023-06-07
Did picoquic have any memory consumption exceptions during my test?


dongsf:
It seems that the client side has a lot of memory leaks. and stream readcb is not triggered properly.
The exception is occasional，and I can't find the root cause.
My picoquic is behind the current branch a lot. I will test it again with new one.

dongsf:
[Uploading ed83cb8a37546ed9.client.qlog.zip…]()

dongsf:
[ed83cb8a37546ed9.server.qlog.zip](https://github.com/private-octopus/picoquic/files/11672300/ed83cb8a37546ed9.server.qlog.zip)

dongsf:
[ed83cb8a37546ed9.client.qlog.zip](https://github.com/private-octopus/picoquic/files/11672939/ed83cb8a37546ed9.client.qlog.zip)

dongsf:
In the absence of a sanitizer on my client side, I have tested it by adding qlog. Through this, I have discovered that the memory consumption is a result of data loss in the stream queue. Additionally, I have noticed that the server side is experiencing a high number of retransmits, which unfortunately exacerbates the data loss problem.
![image](https://github.com/private-octopus/picoquic/assets/7805075/dd90987c-22d5-4d88-bf1e-7eec4f527008)

dongsf:
[10ecae863449daee.zip](https://github.com/private-octopus/picoquic/files/11752491/10ecae863449daee.zip)
[10ecae863449daee.server.qlog.zip](https://github.com/private-octopus/picoquic/files/11752492/10ecae863449daee.server.qlog.zip)

huitema:
Thanks for the detailed traces. I think I understand what is happening. This is not a memory leak per se, but for you the effect is probably the same. The amount of memory required grows a lot when lots of packets are lost. This may well causes adverse effects, such as slowing down the server, causing further losses. I think we should study ways to reduce or limit the memory consumption.

The normal path for packet processing is:

* "packet" structure allocated to hold packet.
* packet is composed, e.g., copying stream data, datagrams or other frames in content
* encrypted version of packet is sent over the network
* packet structure is kept in the queue of unacknowledged packets.
* acknowledgement is received
* packet is recycled.

In case of transmission error, this becomes:

*  "packet" structure allocated to hold packet.
* packet is composed, e.g., copying stream data, datagrams or other frames in content
* encrypted version of packet is sent over the network
[...]

PR #1520, opened 2023-07-12
Fix the perceived memory leak of nodes structures

huitema:
The stream context was not properly cleared after receiving a "reset stream" frame. This was demonstrated by adding a test of allocated node memory in the "stop sending" test, and then fixing the issue.

This should address issue #1499.


huitema:
Still working on this, because there is a secondary issue. The "node leak" issue can be best reproduced if there is a big packet loss rate. However, with the specific loss rate programmed for the repro, there is a high chance of experiencing spurious retransmissions. This leads somehow to cycles of repeating the "reset stream" packets, probably because of the very specific loss pattern programmed in the test.

The root cause of this repetition is that the only memory of the stream state is the stream context. The cycle could be stopped if the stream context recorded whether the "reset stream" was acked by the peer or not. It currently does not, because the stream context is closed pretty much as soon as the reset is sent. Instead we should:

* test the reset state before repeating the "reset stream" frame.
* only delete the stream context after the "reset stream" frame has been acknowledged.

Issue #1527, opened 2023-07-18
NULL pointer deference causing Picoquic server to crash


QUICTester:
Hi, when testing commit d2f01093, we discover Null pointer deference that causes a crash on the Picoquic server. 

Sending the following frame sequence as a client will crash the server: 
1) Ping frame
2) Crypto frame containing Client Hello message
3) ACK frame without padding that acknowledges the Initial packet (Server Hello message) from the server
3) ACK frame that acknowledges the Handshake packet (Certificate, Certificate Verify, Finished messages) from the server

Root cause: Picoquic does not correctly handles the Initial retransmission queue. Picoquic’s method of emptying the Initial retransmission queue in `picoquic_implicit_handshake_ack()` does not follow the order in which the packets were added. Some packets will remain in the retransmission queue after `picoquic_implicit_handshake_ack()` is called. This will trigger a retransmission callback even when the Initial key was discarded, causing NULL pointer deference when the server tries to encrypt the Initial packet for retransmission. 

In `picoquic/sender.c:picoquic_implicit_handshake_ack()`, `p_next` should be `p->next_packet` instead of `p->previous_packet`.

huitema:
I think something similar was fixed recently. Does the crash repeat on the latest version of picoquic?

huitema:
Clarification question about the exact test sequence. I understand it as:

1. Test Client sends an Initial packet containing a Ping frame, padded to required length;
2. Test Client sends an Initial packet with a Crypto frame containing the Client Hello message;
3. Test Client waits for Initial packet from server containing the Server Hello message
4. Test Client sends an Initial packet containing an ACK frame without padding that acknowledges the Initial packet (Server Hello message) from the server
5. Test Client sends an Handshake packet containing an ACK frame that acknowledges the Handshake packet (Certificate, Certificate Verify, Finished messages) from the server

Is this correct?

huitema:
Here is the qlog visualization of my trial at reproducing the bug:

![image](https://github.com/private-octopus/picoquic/assets/7464128/2ee24e5a-9d4c-4a35-97aa-5905365e510e)

Start with an initial ping, then the client sends the client hello and waits for the server hello. That server hello arrives at T=12.00. (The graph says 1.0, but there is a transmission delay in the simulation.) The client cannot send handshake packets immediately, but by the time they arrive the sender has finished sending a whole flight. The additional packets are processed without a glitch.

QUICTester:
[...]

PR #1529, opened 2023-07-19
Verify reported crash does not repro.

huitema:
There was a potential crash reported in issue #1527. It turns out that the test was conducted on a previous version of Picoquic, and that the issue was already fixed. The additional test verifies that.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,6 +149,7 @@ set(PICOQUIC_TEST_LIBRARY_FILES
     picoquictest/parseheadertest.c
     picoquictest/picoquic_lb_test.c
     picoquictest/pn2pn64test.c
+    picoquictest/quic_tester.c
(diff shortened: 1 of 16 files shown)


huitema:
Analysis of the issue pointed out the misleading names of the pointers for the queue of packets awaiting ack or retransmission. Applying a set of name changes to reverse that:

Old name | what | new name
---------- | ---- | ----------
`pkt_ctx->retransmit_newest` | the last packet inserted in the queue of packets waiting ACK or retransmit | `pkt_ctx->pending_last`
`pkt_ctx->retransmit_oldest` | the oldest packet in the queue | `pkt_ctx->pending_first`
`packet->next_packet` | pointer to the previous packet in the queue | `packet->packet_previous`
`packet->previous_packet` | pointer to the next packet in the queue | `packet->packet_next`

PR #1544, opened 2023-09-01
Allow debug output to be delivered via callbacks

paulej:
These changes allow both normal debug logging and `debug_dump()` output to be delivered to the application via callback functions.  This is useful so that debug output can be seamlessly integrated with other logging output.  The changes allow the user to specify either a logging stream or callback function, but not both.  (I did not see a use case for allowing delivery of messages to two places, but I have no objection to it.)

### DIFF
--- a/picoquic/picoquic_utils.h
+++ b/picoquic/picoquic_utils.h
@@ -65,6 +65,7 @@ extern "C" {
 
 
 void debug_set_stream(FILE *F);
+void debug_set_callback(void (*cb)(const char *msg, void *argp), void *argp);
 void debug_printf(const char* fmt, ...);
 void debug_printf_push_stream(FILE* f);
 void debug_printf_pop_stream(void);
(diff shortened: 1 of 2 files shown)


huitema (changes requested):
Looks good, but please fix the warning in appveyor:
~~~
       "C:\projects\picoquic\picoquic.sln" (default target) (1) ->
       "C:\projects\picoquic\picoquic\picoquic.vcxproj" (default target) (2) ->
       (ClCompile target) -> 
         c:\projects\picoquic\picoquic\util.c(168): warning C4267: '=': conversion from 'size_t' to 'int', possible loss of data [C:\projects\picoquic\picoquic\picoquic.vcxproj]
         c:\projects\picoquic\picoquic\util.c(172): warning C4267: '=': conversion from 'size_t' to 'int', possible loss of data 
~~~

paulej:
@huitema Thanks for pointing that out. I fixed that issue, as requested.

paulej:
Actually, I don't like where I defined `mlen`. I moved it.

Issue #1481, opened 2023-05-05
Delay spikes caused degraded conditions on Wi-Fi networks


huitema:
I knew that Wi-Fi was sometimes kind of bad, but I did not look closely until someone complained about the perfs of QUIC over WiFi. Just did a 2 minute sample of ping delays, and I get something like:

* On my iMAC: median 8.5ms, 90th percentile 13.2ms, 95th 19.1ms, max 150ms
* On my PC: median 2ms, 90th percentile 4ms, 95th 14.2ms, max 86ms

We can see that on this graph:

![image](https://user-images.githubusercontent.com/7464128/236545610-72ad7570-ef6a-4546-adee-fb159f06b8da.png)

That seems to me happening in many places, colleagues are reporting similar data.I did a series of ping at 0.1sec intervals, and I see the "spike event" spending 8 to 12 ping times, probably lasting about a second. 

![image](https://user-images.githubusercontent.com/7464128/236546637-42fe5698-e80a-420c-bae1-0f0d654e9f05.png)

We are getting reports that loss recovery algorithm gets very confused, congestion control kicks in, etc.

The test problem making the program obvious were sending QUIC datagrams at a high rate (600 datagrams per second). First step would be to try reproduce the problem in a simulation.

TimEvens:
https://www.privateoctopus.com/2023/05/18/the-weird-case-of-wifi-latency-spikes.html describes 

This problem is still there. 

We have tried bbr, newreno, and cubic in picoquic.  All of them have the same behavior with the latency spikes.  Seems a bit odd that they all have the same behavior where the callbacks are delayed per stream (including datagram) during the latency spike. Please note that latency spike results in ZERO packet loss, it's just a latency spike of `~100 - 150ms` for a few packets over a span of a few seconds.  Not all packets are delayed during the spike.

huitema:
When a spike occurs, Wi-Fi operation is suspended for 100-250ms. The Wi-Fi station sends a message to the Wi-Fi base station to "please queue data for 250ms, and ping me after that", and then it goes to sleep. Nothing is sent on the link, nothing is received. Technically, it is possible to send of the UDP socket -- the packets will be queued inside the kernel, and will only be sent when Wi-Fi operation resumes, after the end of the event.

[...]

PR #1545, opened 2023-09-01
Try wifi-shadow-rtt

huitema:
This is an attempt to mitigate issue #1481.

An annoying side effect of "wifi-spikes" (in fact, wi-fi suspensions) is that the congestion window soon becomes exhausted, causing the stack to stop the "prepare packets" callbacks. The proposal adds a "shadow rtt" API, that could be used to tell BBR to allocated CWIN credits for at least that duration.


suhasHere (approved):

huitema:
This is probably not ready for merging yet. Let's first get report of field trials with actual wi-fi networks.

huitema:
I am thinking about this some more. There are a few open issues. For example, BBR reevaluates the data rate at the end of each "epoch", understood as "one RTT", but this is not well defined during error conditions. It could be:

* time based: use the RTT measured at the beginning of the epoch to limit the epoch's duration.
* packet based: receipt of an ACK for any packet sent after the beginning of the epoch. This is arguably better than time-based, because it adapts to changing RTT, queues, etc. In our case, it would only trigger after the end of the outage.
* packet based, with loss tracking: end epoch if any packet sent after the beginning of the epoch being declared lost.

I tried changing this in the implementation before, but it is one of those subtle changes with cascading effects, with some test cases improving and other worsening. Maybe we need to reconsider that if the "shadow RTT" is set, so the epoch only ends if either a packet is acked, or if a timer that includes the shadow RTT elapses. But then, that probably also requires including the shadow RTT in the management of repeat timers.

Before we do any of that, it would be good to get some practical results.

Issue #1547, opened 2023-09-13
BBR low sending rate when application-limited 


alexrabi:
Seems like there is a problem in the BBR implementation if the sender is application-limited rather than network limited. If the sender only sporadically has data to send, it appears that the CWND becomes very small. This severely limits the sending rate once the application actually has data to send.

A similar problem has already been described in https://github.com/private-octopus/picoquic/issues/1499#issuecomment-1645750380

huitema:
This is a classic congestion control issue. BBR sets the pacing rate based on the highest value observed in the recent past. The definition of recent is debatable: if the highest rate was observed a long time ago, then the network conditions have probably changed and the observation is not valid anymore. If rates observed too recently are discarded, we get the situation that you describe.

Can you characterize the sending pattern of your applications? That would help setting a simulation, and understanding the issue more precisely.

alexrabi:
BBR should ignore rate samples if it is application-limited, looking at the ietf draft, though I'm not sure if the picoquic implementation does this correctly. 

The sending pattern is literally TCP traffic; the problem is observed when tunneling a TCP connection over QUIC datagrams. More specifically, the problem occurs when TCP is sending pure TCP ACKs, so this would be able to be simulated by sending one or more small datagrams roughly every round-trip-time (or slightly longer).

alexrabi:
After doing some more tests, it looks this problem only happens when using multipath, and there is too little data to send to use more than one path. If there is only a single path available, there is no issue (whether multipath was negotiated or not).

huitema:
Thanks for the precision. We test code includes test cases of limited senders, which are passing, but they obviously do not include your multipath configuration.

There is indeed a special case if there is no traffic at all on a path for some time. We probably need to add a test case for that.

huitema:
Trying to understand your scenario. Apparently:

* you are testing a VPN connection, with traffic carried as datagrams over multipath QUIC.
* for an extended period, the traffic from the server corresponds to TCP data packets, and the traffic from the client consists of TCP ACK packets. 
* somehow, the TCP ACK packets are sent on just one of the available paths -- maybe because one path is preferred by the scheduler.
* after some time, the client side of the connection wakes up and start sending TCP data packets.
* the expectation is that those packets will be spread over both paths, but at this point BBR parameters are such that very little traffic is sent on the second path.

Is that a correct assessment?

Any idea of the duration of the "low traffic" interval?

alexrabi:
[...]

PR #1548, opened 2023-09-19
Rewrite BBR handling of limited applications

huitema:
The main part of this PR is a rewrite of the function `BBRUpdateBtlBw` in `bbr.c`, to implement a proper handling of limited applications. The code includes a list of previous bandwidth measurements, and BBR selects the maximum available in that list. The previous version would forget the oldest measurement at the end of each epoch; the new version only does that if the new measurement is larger than the current data rate, or was not captured in limited conditions. This will help solving the issue observed by @alexrabi in #1547.

The PR also includes a new test, in which short datagrams are sent each in their own packet instead of being possibly grouped in larger packets. This is an attempt to reproduce issue #1547, but it is still work in progress. There may be a remaining issue caused by the function `picoquic_update_pacing_after_send` in `sender.c`. That function consumes a fixed amount of pacing credit per packet, which could be problematic if the traffic consists mostly of short packets. We need to address that before merging this PR.


huitema:
The last commit fixes the pacing of small packets by accounting for packet length in `picoquic_update_pacing_after_send` in `sender.c`. The effect of that fixes are mixed: it changes the timing of packets, which most of the time results in better performance, but sometimes causes event sequences to change, causing some tests to run slower. Overall it is better, because it does fix pacing issues when the traffic includes lots of small packets, and ensure that the data rate computed by the congestion control algorithm is consistent with the actual pacing rate.

huitema:
The updated "small packet" test verifies that 10,000 small packets can be sent in a short time using BBR, without triggering a congestion control issue.

PR #1549, opened 2023-09-20
Web transport doc

huitema:
Create a very rough first cut for a web transport documentation.


huitema:
These last commits fix the various constants to be consistent with draft 7 of web transport.

Close #1550

PR #1553, opened 2023-10-02
Only set H3 datagrams if set in QUIC

huitema:


huitema:
This should fix issue #1553

Issue #1561, opened 2023-10-11
Crash in picoquic_aead_get_checksum_length due to state mismatch or unprotected


fbussery:
- Problem on picoquic vesion: https://github.com/private-octopus/picoquic/commit/1e2979e8db0957c8ee798940091c4d0ef13bf8af
- Using it here as a server

Sometime, I have the following crash due to the call of apicoquic_protect_packet called with a null aead_context.
 
Below the kind of stack we have:
``` 
#4  picoquic_aead_get_checksum_length (aead_context=aead_context@entry=0x0)
    at picoquic/tls_api.c:2266
#5  0x00005603e3cfa38c in picoquic_protect_packet (cnx=cnx@entry=0x5603e4c3c800, ptype=ptype@entry=picoquic_packet_initial, 
    bytes=bytes@entry=0x5603e478fb8a "", sequence_number=124746, length=length@entry=39, header_length=30, 
    send_buffer=0x5603e4c9fe10 "...", send_buffer_max=1232, aead_context=0x0, 
    pn_enc=0x0, path_x=0x5603e4c55de0, current_time=1697029105359355)
    at picoquic/sender.c:774
#6  0x00005603e3cfb74b in picoquic_finalize_and_protect_packet (cnx=0x5603e4c3c800, packet=0x5603e478faf0, ret=<optimized out>, length=39, 
    header_length=<optimized out>, checksum_overhead=8, send_length=<optimized out>, send_buffer=<optimized out>, send_buffer_max=<optimized out>, 
    path_x=<optimized out>, current_time=<optimized out>)
    at picoquic/sender.c:1266
#7  0x00005603e3cfed41 in picoquic_prepare_packet_server_init (cnx=cnx@entry=0x5603e4c3c800, path_x=0x5603e4c55de0, packet=packet@entry=0x5603e478faf0, 
    current_time=current_time@entry=1697029105359355, send_buffer=<optimized out>, send_buffer_max=<optimized out>, send_length=<optimized out>, 
    next_wake_time=<optimized out>, is_initial_sent=<optimized out>)
    at picoquic/sender.c:3013
#8  0x00005603e3d037e8 in picoquic_prepare_segment (is_initial_sent=0x7ffc4c1943ac, next_wake_time=0x7ffc4c1943b0, send_length=0x7ffc4c1943b8, 
    send_buffer_max=<optimized out>, 
    send_buffer=0x5603e4c9fe10 "...", current_time=1697029105359355, 
[...]

PR #1563, opened 2023-10-12
Simple protection against use deleted crypto

huitema:
This should address issue #1561


fbussery (approved):
Seems good, so far. Tested on a mesh network with 60 nodes and I don't see any more problem now.

huitema:
Will check this in now. Reading the code again, it could clearly be refactored, because it was built step by step as the specification was developed. We could break the handshake states into specific subsets each with simpler processing. But this is a project in itself, not quite on the top of the priority pile right now.

Close #1561

PR #1566, opened 2023-10-13
Only check receive times for idle timer

### LINKED ISSUE #1559: Retransmissions should not reset the idle timer.
The current code resets the idle timer if a new packet is sent, for example for retransmission. This means the idle timer never fires if the local endpoint is sending packets, even if these packets are never acknowledged. The idle timer should only be reset if new data is received from the peer, or if acknowledgements from the peer show progress in the connection.

huitema:
Close #1559


TimEvens (approved):
Idle timeouts appear to be working correctly now. Looks good.

PR #1570, opened 2023-10-17
Expose RTT sample in Path Quality API

alexrabi:
The most recent RTT sample collected on a path may be useful when evaluating path quality. This PR adds the RTT sample to the Path Quality API.


huitema (approved):
I like it, sounds like a good idea. It works without any impact on the tests, but that shows a bit of an oversight in the test design. The current tests `quality_update_test` in `tls_api_test.c` only monitors "Time, Path_id, Sending_rate_CB, Pacing_rate, Receive_Rate, CWIN, RTT". We should probably add `rtt_sample` to that. But then, we should also add `rtt_variant`, `rtt_min` and `rtt_max`. So, that will have to wait for another PR.

PR #1571, opened 2023-10-24
Remove unused variable

paulej:
Remove an unused variable in the function h3zero_varint_from_stream().


huitema (approved):
Good catch. Probably a leftover of hesitating between two designs...

PR #1588, opened 2023-12-02
ACK and Stream are not path probing

### LINKED ISSUE #1587: A pure ACK_ENC packet should be non-probing
Currently, if a packet contains only an ACK_ENC frame, then picoquic_decode_frames treats it as a validating_packet.  This behavior is not in line with the RFC, which says:

> PATH_CHALLENGE, PATH_RESPONSE, NEW_CONNECTION_ID, and PADDING frames are "probing frames", and all other frames are "non-probing frames". A packet containing only probing frames is a "probing packet", and a packet containing any other frame is a "non-probing packet".

As a result, the server continues to send packets on an old path even if both parties have been validated a new path and the client only sends ACK_ENC frames.

I believe the following simple patch fixes the issue.

```patch
diff --git a/picoquic/frames.c b/picoquic/frames.c
index b9e5dd50..fffe8306 100644
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -5603,8 +5603,8 @@ int picoquic_decode_frames(picoquic_cnx_t* cnx, picoquic_path_t * path_x, const
                 break;
             }
             }
-            is_path_validating_packet &= is_path_validating_frame;
         }
+        is_path_validating_packet &= is_path_validating_frame;
     }
 
     if (bytes != NULL) {
```

Thank you.

huitema:
Close #1587

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -5263,7 +5263,7 @@ int picoquic_decode_frames(picoquic_cnx_t* cnx, picoquic_path_t * path_x, const 
 {
     const uint8_t *bytes_max = bytes + bytes_maxsize;
     int ack_needed = 0;
-    int is_path_validating_packet = 1; /* Will be set to zero if non validating frame received */
+    int is_path_probing_packet = 1; /* Will be set to zero if non validating frame received */
     picoquic_packet_context_enum pc = picoquic_context_from_epoch(epoch);
     picoquic_packet_data_t packet_data;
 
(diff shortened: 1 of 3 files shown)


picoquic/frames.c
```diff
@@ -5263,15 +5263,15 @@ int picoquic_decode_frames(picoquic_cnx_t* cnx, picoquic_path_t * path_x, const
 {
     const uint8_t *bytes_max = bytes + bytes_maxsize;
     int ack_needed = 0;
-    int is_path_validating_packet = 1; /* Will be set to zero if non validating frame received */
+    int is_path_probing_packet = 1; /* Will be set to zero if non validating frame received */
```
nemethf, line 5266: Maybe the comment should be updated as well.

(Thank you for your quick response.)

PR #1591, opened 2023-12-06
Sock loop wakeup

huitema:
Trial only -- work in progress.


huitema:
Merging now. The purpose of the exercise was to explore options for supporting threads across Windows and Unix. The "thread tester" test code does that, showing how to set up threads, how to wake-up threads using signals on Windows, or pipes on Unix, and how to properly wind down threads. There are multiple follow-on:

- Use a threaded test program to do "end to end" testing of the stack and the socket loop using local sockets.
- Maybe do the same for an end-to-end test of the picoquicdemo app.
- Upgrade the socket loop to allow wakeup. Maybe requires unifying the windows and unix version of the socket loop.
- Write a sample of a web-socket style application using the improved socket loop.

Those will have to be pursued in separate PRs.

PR #1596, opened 2023-12-13
Verify the idle timeout negotiation

huitema:
Add test of idle timeout negotiation, trying 10 combinations of client and server proposed values, including the default value, and negotiating the "infinity" value. Verify that the connection stays up if the timer is not expired, and is closed soon after the timer expires.

Only change in the stack code is to remove a "default_timeout" parameter in the QUIC context, as that parameter was not used.

### DIFF
--- a/picoquic/picoquic.h
+++ b/picoquic/picoquic.h
@@ -40,7 +40,7 @@
 extern "C" {
 #endif
 
-#define PICOQUIC_VERSION "1.1.16.0"
+#define PICOQUIC_VERSION "1.1.16.1"
 #define PICOQUIC_ERROR_CLASS 0x400
 #define PICOQUIC_ERROR_DUPLICATE (PICOQUIC_ERROR_CLASS + 1)
 #define PICOQUIC_ERROR_AEAD_CHECK (PICOQUIC_ERROR_CLASS + 3)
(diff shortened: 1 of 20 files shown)


picoquic/picoquic.h
```diff
@@ -287,7 +287,7 @@ typedef struct st_picoquic_tp_t {
     uint64_t initial_max_data;
     uint64_t initial_max_stream_id_bidir;
     uint64_t initial_max_stream_id_unidir;
-    uint64_t idle_timeout;
+    uint64_t max_idle_timeout;
```
TimEvens, line 290: Thanks for the update in the name.

picoquic/sender.c
```diff
@@ -3860,8 +3860,8 @@ static int picoquic_check_idle_timer(picoquic_cnx_t* cnx, uint64_t* next_wake_ti
             idle_timer = UINT64_MAX;
         }
     }
-    else if (cnx->local_parameters.idle_timeout > (PICOQUIC_MICROSEC_HANDSHAKE_MAX / 1000)) {
-        idle_timer = cnx->start_time + cnx->local_parameters.idle_timeout*1000ull;
+    else if (cnx->local_parameters.max_idle_timeout > (PICOQUIC_MICROSEC_HANDSHAKE_MAX / 1000)) {
+        idle_timer = cnx->start_time + cnx->local_parameters.max_idle_timeout*1000ull;
     }
     else {
         idle_timer = cnx->start_time + PICOQUIC_MICROSEC_HANDSHAKE_MAX;
```
TimEvens, line 3867: This becomes the **connection timeout** before negotiation happens when `max_idle_timeout` is less than `PICOQUIC_MICROSEC_HANDSHAKE_MAX`.  It's currently hard coded to a minimum of 30 seconds. It would be nice if we had a configuration for connection timeout.

TimEvens (commented):
It looks good with the name change to `max_idle_timeout` and to test negotiation, but there is still no test for connection timeout, which idle timeout is being used for.  

I suppose since this PR is titled "... negotiation" it doesn't apply to connection timeout tests or configuration....?

Currently connection timeout is using a hardcoded value of 30 seconds for the **minimum** idle timeout when not connected/ready.  This is resulting in delays of 30 seconds for clients that had a typo in a port number or IP.  The error indicate is also a bit odd saying idle timeout when it wasn't even connected in the first place. IMHO, "idle" means it was connected and now has become idle.  A connection that was never connected should have had an error  to indicate connection establishment timeout.

TimEvens (approved):
LGTM. The changes directly address the issue we were seeing.

PR #1600, opened 2023-12-20
Tests for coverage of config.c

huitema:


huitema:
This PR was motivated by a first run of test coverage measurements using `gcov`. The focus is on the libraries `picoquic-core`, `picohttp-core`, and `picoquic-log`. The first measurements show generally good coverage, with exceptions. One of those was `picoquic\config.c`. The goal of this first PR was to assess how much time it would take to fix coverage for a moderately large file. Answer: about a day.

Test coverage, `config.c` | Lines | Branches 
----------------|------|-----------
Before this PR | 59.7% | 51.0%
After this PR   | 92.8% | 76.0%

As shown in the change log, this was done with a combination of writing new tests, adding negative test cases to existing tests, and removing unused code. The exercise did uncover a number of issues in some rarely used faulty input scenarios, proving that there is merit in measuring code coverage.

Issue #1601, opened 2023-12-20
[RTT] Miscalculation of rtt on server side


fbussery:
**Picoquic version:** 2023-09-20  8:58
**Test environement:** Simulated adsl network

The simulated adsl: 
    - with ping time 60ms
    - upload speed of 2Mb/s
    - download speed of 12Mb/s
    - (ping between 2 peers is then around 120ms)

**The observed problem:**

Sometime we notice that the rtt min is very low. (around 500 micro seconds). This is definitively not expected. The impact is that the bbr is mistaking the bandwidth.

I did attach the log
[qlog-05944aef04573529.server.log](https://github.com/private-octopus/picoquic/files/13725313/qlog-05944aef04573529.server.log)

On line 208, we have:
[12904429, "recovery", "metrics_updated", {"cwnd": 5760,"bytes_in_flight": 0,"smoothed_rtt": 105856,"min_rtt": 552,"latest_rtt": 552}],

**Expected behaviour**

We expect the rtt to be all the time more the 120000

I did not reproduced the issue using picoquicdemo  perf that always return a correct rtt_min (120ms)

fbussery:
Some further investigation drives that the problem is certainly linked to the poor accurancy of gettimeofday which is used in picoquic_current_time(). 
May be we should use clock_gettime instead.

Something like that:
```
uint64_t picoquic_current_time()
...
#else
    struct timespec currentTime;
    clock_gettime(CLOCK_MONOTONIC, &currentTime);
    now = (currentTime.tv_sec * 1000000ull) + currentTime.tv_nsec / 1000ull;
#endif
    return now;

```
Seems better now...

fbussery:
Further tests shows that problem may be the picoquic_current_time accurency in my system. Did test also with CLOCK_MONOTONIC. But with similar behaviour.

fbussery:
seems that there are no bug in picoquic concerning the rtt. 
I have correct sequence in picoquic. My simulation setup may be mis-functionning.
I will close the ticket tomorrow

huitema:
On Linux and macOS, the code indeed uses "gettimeofday", which does provide a time in microseconds but does not exactly specify the resolution. I think it does work well, and has the advantage of being portable. There is a drawback, per the history section of the man page: "The time returned by gettimeofday() is affected by discontinuous jumps in the system time (e.g., if the system administrator manually changes the system time)."

Given that, yes, using `clock_gettime()` might be safer. If you have a PR, I will review it.

[...]

PR #1605, opened 2023-12-28
Fix time issues

huitema:
Issue #1601 points out that using `gettimeofday()` for managing timers can cause errors if the local admin resets that time. It is preferable on Linux to use `clock_gettime(CLOCK_MONOTONIC)` for that purpose. This requires separating the TLS related usage of time, which has to track "wall time", from the timer-related usage prevalent in QUIC protocol handling.

The solution here is:

* make sure that ticket and token related calls use `picoquic_get_tls_time(quic)`, and rewrite the token and ticket API so as to *not* take a time parameter;
* incorporate the patch proposed by @fbussery and use `CLOCK_MONOTONIC` for other protocol actions.


huitema:
Last point: clock_gettime may in some cases return an error, but merely reading the clock should only cause an issue if the calling arguments are malformed. I would rather ignore the return code than create an error branch that is never exercised.

picoquictest/tls_api_test.c
```diff
@@ -5117,7 +5116,7 @@ int virtual_time_test()
         {
             /* Check that the simulated time follows the simulation */
             for (int i = 0; ret == 0 && i < 5; i++) {
-                simulated_time += 12345678;
+                simulated_time += 12345000;
```
fbussery, line 5119: To have same behaviour as before, may be we should have simulated_time += 12345678000 ?

picoquictest/ticket_store_test.c
```diff
@@ -118,20 +118,31 @@ int ticket_store_test()
...
+    }
+
+    if (ret == 0) {
+        /* Writing an empty file */
+        ret = picoquic_save_tickets(p_first_ticket, current_time, test_ticket_file_name);
+    }
 
     /* Load the empty file again */
     if (ret == 0) {
-        ret = picoquic_load_tickets(&p_first_ticket_empty, retrieve_time, test_ticket_file_name);
```
fbussery, line 127: p_first_ticket_empty still defined in line 111. Not needed any more

picoquictest/tls_api_test.c
```diff
@@ -5116,7 +5116,7 @@ int virtual_time_test()
         {
             /* Check that the simulated time follows the simulation */
             for (int i = 0; ret == 0 && i < 5; i++) {
-                simulated_time += 12345000;
+                simulated_time += 12345678000;
```
fbussery, line 5119: I have a doubt.  In my mind, simulated_time was of same unity as picoquic_get_quic_time which did not change

huitema:
I also have a doubt. The "gettimeofday" approach has the nice property of inheriting clock synchronization, NTP, etc. This is not strictly required for the transport, but it does make debugging easier, e.g., when comparing time stamps at the client and the server. I wonder whether the use of monotonic clock should be controlled by a compile option.

PR #1607, opened 2023-12-30
Add API to set BBR quantum ratio

huitema:
This PR adds an experimental API to change the "quantum ratio" of BBR from the default 0.001 (1/1000th of the pacing rate in bytes per second) to a larger value. The API is:
~~~
void picoquic_set_default_bbr_quantum_ratio(picoquic_quic_t* quic, double quantum_ratio);
~~~
It should be called before a connection is created. Calling the API does not affect existing connections.

### DIFF
--- a/picoquic/bbr.c
+++ b/picoquic/bbr.c
@@ -279,6 +279,7 @@ typedef struct st_picoquic_bbr_state_t {
     uint64_t cwin_before_suspension; /* So it can be restored if suspension stops. */
 
     uint64_t wifi_shadow_rtt; /* Shadow RTT used for wifi connections. */
+    double quantum_ratio; /* Fraction of pacing rate used for Quantum, or zero if not set*/
     unsigned int filled_pipe : 1;
     unsigned int round_start : 1;
     unsigned int rt_prop_expired : 1;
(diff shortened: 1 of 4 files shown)


huitema:
I made an additional change to BBR after analyzing a misbehaving implementation of the socket loop. On a PC build, the socket loop gets performance from four features: asynchronous receive, batched send, the Windows equivalent of GSO, batched receive (equivalent to GRO) and asynchronous send. The work in progress only had implementation for asynchronous receive, batched send and batched receive, and that turned out to be very slow - 6 to 10 Mbps instead of the expected 1Gbps. Of course, this is most probably due to the missing feature, but investigation showed that the slowness was made worse by a weird interaction with BBR:

1) When using batched send, the RTT increases. If the connection remained with a high RTT for over 10 seconds, the "renew RTT min evaluation" feature of RTT min kicks in, and after that the bandwidth is stuck at 6-7 Mbps

2) When the batches are sent slowly, they last a long time, way more than the low RTT. This causes the datarate evaluation module to misfunction, which creates a negative feedback loop inside BBR, driving the pacing rate down to a floor value of about 5 Mbps.

The two fixes are to make the RTT min evaluation much more lenient if the RTT is really small, and to add a parallel evaluation of the datarate as a sanity check. This mitigates the issue somewhat, with observed data rates in the 20-30 Mbps range. That's still bad, but not quite as bad.

huitema:
Of course, this mitigation is just a temporary band-aid. When performance is low, the proper solution is to understand what causes the slow-down.

huitema:
This branch is obsolete. It is superceded by the branch "bbr3-quantum-ratio", which provides support for the quantum ratio API over BBRv3 (default CC) and also BBRv1 (option).

PR #1611, opened 2024-01-05
Document building picoquic

### LINKED ISSUE #1604: [Documentation] Building (and Rebuilding) single targets
To build a single target defined in `CMakeLists.txt` the command `cmake --build . --target <target_name>` can be used this makes (re)building targets easier. This will rebuild the target and it's dependencies (if anything has changed). Since this works for all executables defined in `CMakeLists.txt` I'm unsure how to add this to the documentation. Nonetheless this would be useful to mention somewhere, seeing as independent building is mentioned in `sample/README.md`.

For example to (re)build only the sample in `sample/` the command `cmake --build . --target picoquic_sample` can be used. The command was run from the top most directory (the one`CMakeLists.txt` is in).

notBroman:
resolve #1604 

I would appreciate pointers on documentation for how picoquic is supposed to be added/integrated into other projects. I think it would be useful to explain how this project is expected to be used in other projects.

Additionally pointers on what is missing would be appreciated.

### DIFF
--- a/doc/building_picoquic.md
+++ b/doc/building_picoquic.md
@@ -4,11 +4,16 @@ See the 'Building Picoquic' section [README](../README.md) for building picoquic
 
 ## Dependencies
 
-- [Picotls implementation of TLS 1.3](https://github.com/h2o/picotls) 
+- [Picotls implementation of TLS 1.3](https://github.com/h2o/picotls)
+
+## How to provide your own crypto library
+
+
 
 ## Picoquic as a Dependency
 
-TODO
+TODO:
+_describe how to include picoquic in other projects?_
 
 ## Targets
 
@@ -28,8 +33,8 @@ All of these targets are built when the `make .` or `cmake --build .` commands a
 `picoquicdemo` ([found in picoquicfirst/picoquicdemo.c](../picoquicfirst/picoquicdemo.c)) and `picoquic_sample` ([sample documentation](../sample/README.md)) are a good starting point for developing an application on top of this QUIC implementation.
 
 
-## Building a Single Target
+## (Re)Building a Single Target
 
-Targets defined in the `CMakeLists.txt` can be build individually using cmake's `--target` option. The command `cmake --build . --target <target_name>` can be used to build a specific target and all it's dependcies. For example `cmake --build . --target picoquic_sample` will build the sample file transfer application in [sample/](../sample/README.md) as defined in the `CMakeLists.txt`. All the required components will also be built. This works if the command is issued in the top-level directory of the repository. This is useful for recompiling test programs and programs like `picoquic_sample` and `picoquicdemo` if they have been modified. Therefore reducing build times when making small changes.
+Targets defined in the `CMakeLists.txt`, listed above for convenience, can be build individually using cmake's `--target` option. The command `cmake --build . --target <target_name>` can be used to build a specific target and all it's dependcies. For example `cmake --build . --target picoquic_sample` will build the sample file transfer application in [sample/](../sample/README.md) as defined in the `CMakeLists.txt`. All the required components will also be built. This works if the command is issued in the top-level directory of the repository. This is useful for recompiling test programs and programs like `picoquic_sample` and `picoquicdemo` if they have been modified. Therefore reducing build times when making small changes.
 
 This also works if `picoquic` has not been built as a whole before.
--- a/doc/building_picoquic.md
+++ b/doc/building_picoquic.md
@@ -37,7 +37,18 @@ set(PICOQUIC_FETCH_PTLS ON)
 FetchContent_MakeAvailable(picoquic)
 ```
 
-FetchContent can be used to fetch a specific release or tag.
+FetchContent can be used to fetch a specific release or tag, if a certain version or release is desired.
+This can be used as fine grain as choosing a single commit, be specifying the commit hash.
+
+In the main `CMakeLists.txt` the subdirectory can be added using the `add_subdirectory(<directory_name>)` directive.
+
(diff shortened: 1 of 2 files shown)


doc/building_picoquic.md
```diff
@@ -0,0 +1,37 @@
+# Building PicoQUIC
+
+See the 'Building Picoquic' section [README](../README.md) for building picoquic for the first time and installing the dependencies.
+
+## Dependencies
+
+- [Picotls implementation of TLS 1.3](https://github.com/h2o/picotls)
+
```
huitema, line 8: The other dependencies are inherited from picotls. Currently, this means openssl, or a compatible library like libressl. There is work in progress to allow use of mbedtls instead of openssl.

There is a cmake option to build the dependencies as part of building Picoquic. You can see examples of that in the github actions.

doc/building_picoquic.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Picoquic as a Dependency
+
+TODO:
+_describe how to include picoquic in other projects?_
+
```
huitema, line 14: You can find an example of that in https://github.com/Quicr/quicrq, a project prototyping media over QUIC.
This project uses picoquic as a dependency, and build picoquic if needed.

huitema (commented):
Added comments on dependencies

doc/building_picoquic.md
notBroman: I have started to describe `picoquic as a dependency`, but it feels like there is still something missing in order for it to be useful. I can't put my finger on it though, what's your opinion on this?

doc/building_picoquic.md
```diff
@@ -0,0 +1,84 @@
...
+In the main `CMakeLists.txt` the subdirectory can be added using the `add_subdirectory(<directory_name>)` directive.
+
+```cmake
+# Add the subdirectory dependencies, where dependencies/CMakeLists.txt looks as above
+add_subdirectory(dependencies)
+
+# check if the package was loaded/is available and stop the build if it was not
+find_package(picoquic REQUIRED)
+```
+
```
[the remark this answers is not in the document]
notBroman: I've been testing this in [quicdoq](https://github.com/private-octopus/quicdoq) and got it to work when the fetching snipped was in the top-level directory, but it breaks if it is put in a subdirectory (e.g. dependencies) and added via `add_subdirectory(dependencies)`. I have looked at [quicrq](https://github.com/Quicr/quicrq) since it is fetched in a subdirectory there, but I still can get it to find the fetched content.

It works if I fetchContent in the subdirectory and then configure the path with `ccmake` though.

notBroman:
I'm ready to merge this branch to main, the file has been linked to the README.md. Of course this will have to be update as development progresses, but I think this is a good start.

doc/building_picoquic.md
```diff
@@ -0,0 +1,102 @@
...
+- [Picotls implementation of TLS 1.3](https://github.com/h2o/picotls)
+
+Fetching and installing this package can be done through an option in the CMake.
+
+```shell
+ cmake -DPICOQUIC_FETCH_PTLS=Y .
+```
+
+This can be set in the cmake file as well by setting the `PICOQUIC_FETCH_PTLS` cmake property to `ON`.
+
```
huitema, line 17: The command on line 13 sets `PICOQUIC_FETCH_PTLS` to `Y`. Line 16 suggests setting to `ON`. The GiHub action code uses:
```shell
cmake -S . -B build -DPICOQUIC_FETCH_PTLS:BOOL=ON
```
My suggestion would be to say something like:

Fetching and installing this package can be done by launching CMake with a fetch option:

```shell
 cmake -DPICOQUIC_FETCH_PTLS=Y .
```

And not adding the extra line 16, which seems to encourage people to edit the CMake file manually.

doc/building_picoquic.md
```diff
@@ -0,0 +1,102 @@
...
+- building and fetching PTLS `-DPICOQUIC_FETCH_PTLS=Y`
+- building with exported debug information `-DCMAKE_BUILD_TYPE=Debug`
+
+These two options can be set when configuring cmake.
+
+- adress sanitation `ASAN`
+- undefined behaviour sanitation `UBSAN`
+
+`ASAN` and `UBSAN` need to be set in the `CMakeLists.txt` using the `ENABLE_ASAN` and `ENABLE_UBSAN` options.
+
```
huitema, line 60: There is no need to edit the CMake file, and I would rather not encourage people to do that. The correct text should be something like:

`ASAN` and `UBSAN` are C and C++ compiler options, and can be instantiated by setting compiler flags when launching `cmake`, as in:
```shell
cmake "-DCMAKE_C_FLAGS=-fsanitize=address,undefined -DCMAKE_CXX_FLAGS=-fsanitize=address,undefined"
```

doc/building_picoquic.md
```diff
@@ -0,0 +1,102 @@
...
+set(PICOQUIC_FETCH_PTLS ON)
+end_if()
+
+# Make dependencies available
+FetchContent_MakeAvailable(picoquic)
+```
+
+FetchContent can be used to fetch a specific release or tag, if a certain version or release is desired.
+This can be used as fine grain as choosing a single commit, be specifying the commit hash.
+
```
huitema, line 89: Typo: "by" specifying the commit hash.

huitema (approved):
Thank you for doing this! It is much better to see that as separate file than in the README, and you answer a number of practical questions. I just noted a couple of minor issues, that should be fixed before merging.

huitema (approved):

PR #1624, opened 2024-01-24
Update BBR code to BBRv3

huitema:
This unfortunately required updating the API between transport and congestion control modules, detecting lost packets as part of ACK processing, and changing the way error statistics are computed. These changes required modifications in a number of modules.

<Not ready yet!>


huitema:
It is ready now. All tests are passing, but i had to loosen perf requirements for some of them. There will be further work to do for WiFi and for Multipath, plus adding ECN support and tuning the response to packet drops, but we will do that later.

PR #1633, opened 2024-02-13
Add wake up function to packet loop

huitema:
This PR adds new API to create a background thread handling the packet loop, to wake up that thread and for example create a new data stream, and to delete the thread.

The first commit simply adds the API, and tests that the old way of running a network loop in line with the app still works.


notBroman:
Would it be appreciated if I wrote some documentation around this and the non-threaded socket loop (in a separate branch that can be merged into whatever branch makes sense)?

huitema:
@notBroman yes, documenting the use of the background thread would be very nice! There is minimal documentation in the header file, but an actual doc would help!

huitema:
Tests are passing. Merging now, so we get feedback from developers.

PR #1673, opened 2024-04-14
Allow OpenSSL verifier to be called with NULL SNI

### LINKED ISSUE #1184: Connecting with NULL SNI disables verifier for all connections
The current code has this hack in `picoquic_create_cnx`
```
            if (!quic->is_cert_store_not_empty || sni == NULL) {
                /* This is a hack. The open SSL certifier crashes if no name is specified,
                 * and always fails if no certificate is stored, so we just use a NULL verifier */
                picoquic_log_app_message(cnx, "%s -- certificate will not be verified.\n",
                    (sni == NULL) ? "No server name specified" : "No root crt list specified");

                picoquic_set_null_verifier(quic);
            }
```
This was expedient, but wrong. On one hand, it seems fair to disable the default verifier if no root certificate is verified, but this should be done in the QUIC context, not per connection. On the other hand, we do want to avoid crashes when the SNI is NULL, and it is fine to bypass verification when no name is specified, but this needs to be done somehow by connection, not globally. The next connection may well specify a server name.

If the openssl verifier crashes when no name is specified, the proper solution is to wrap it and test for NULL SNI before calling the inner code.

davidk-ad8:
Removed NULL SNI check in picoquic_create_cnx

Following this recent Picotls change https://github.com/h2o/picotls/pull/520 the verifier now skips the client side SNI check if it is NULL.

This change resolves #1184

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -209,7 +209,7 @@ if(PICOQUIC_FETCH_PTLS)
     include(FetchContent)
     FetchContent_Declare(picotls
         GIT_REPOSITORY      https://github.com/h2o/picotls.git
-        GIT_TAG             af66fc4aa8853b0725fcb2c18a702e8f1c656cf1)
+        GIT_TAG             5a4461d8a3948d9d26bf861e7d90cb80d8093515)
     FetchContent_MakeAvailable(picotls)
 endif()
 
--- a/ci/build_picotls.ps1
+++ b/ci/build_picotls.ps1
@@ -1,5 +1,5 @@
 # Build at a known-good commit
-$COMMIT_ID=" af66fc4aa8853b0725fcb2c18a702e8f1c656cf1"
+$COMMIT_ID=" 5a4461d8a3948d9d26bf861e7d90cb80d8093515"
 
 # Match expectations of picotlsvs project.
 mkdir $dir\include\
--- a/ci/build_picotls.sh
+++ b/ci/build_picotls.sh
@@ -2,7 +2,7 @@
 #build last picotls master (for Travis)
 
 # Build at a known-good commit
-COMMIT_ID= af66fc4aa8853b0725fcb2c18a702e8f1c656cf1
+COMMIT_ID= 5a4461d8a3948d9d26bf861e7d90cb80d8093515
 
 cd ..
 # git clone --branch master --single-branch --shallow-submodules --recurse-submodules --no-tags https://github.com/h2o/picotls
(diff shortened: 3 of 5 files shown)


huitema (commented):
This is very cool, thank you. May I ask just one more thing? There is a "null_sni" test in the test suite today. You can run it as `picoquic_ct null_sni`. It is implemented in `picoquic_test/tls_api_test.c`, lines 9827-9834:

```
/*
 * Test whether a connection succeed when SNI is not specified.
 */

int null_sni_test()
{
    return tls_api_test_with_loss(NULL, PICOQUIC_INTERNAL_TEST_VERSION_1, NULL, PICOQUIC_TEST_ALPN);
}
```
That test succeeds today because of the hack that you are removing, and also because of a corresponding hack in the function `tls_api_test_with_loss()`, also in `picoquic_test/tls_api_test.c`, lines 2167-2180:

```
static int tls_api_test_with_loss(uint64_t* loss_mask, uint32_t proposed_version,
    char const* sni, char const* alpn)
{
    uint64_t simulated_time = 0;
    picoquic_test_tls_api_ctx_t* test_ctx = NULL;
    int ret = tls_api_init_ctx(&test_ctx, proposed_version, sni, alpn, &simulated_time, NULL, NULL, 0, 0, 0);

    if (ret != 0)
    {
        DBG_PRINTF("Could not create the QUIC test contexts for V=%x\n", proposed_version);
    }
    else if (sni == NULL) {
        picoquic_set_null_verifier(test_ctx->qclient);
    }
    ...
```

Could you remove the hack `else if (sni == NULL) { ...`? Then that test will check that your fixes are working, and detect any regression in the future...

huitema:
We also need to understand why the "CiTest" run is failing, with the diagnostic:
```
Starting test number 335, cert_verify_null_sni
picoquic_ct: /home/runner/work/picoquic/picoquic/build/_deps/picotls-src/lib/openssl.c:1560: verify_cert_chain: Assertion `server_name != NULL && "ptls_set_server_name MUST be called"' failed.
<end of output>
Test time =  30.87 sec
```

davidk-ad8:
I'll update the test suite.
Picotls only merged the changes a few days ago so the the version pulled in by Picoquic will need to be updated too.

davidk-ad8:
I've updated `tls_api_test_with_loss` and `Picotls` has been updated to the latest commit needed to support these changes.

I've run `picoquic_ct` locally and all tests pass. `mediatest_video2_back` failed the first time with the message 

> Max latency expected: 600000, got 615833 

but it succeeded on the next 2 runs.

huitema (commented):
Thanks for the PR!

PR #1680, opened 2024-04-25
Notification of loss of feedback

huitema:
With the early notification of loss of feedback, congestion control algorithms can react to loss of connectivity or temporary suspension sooner, avoiding the build up of queues of low priority packets.

### DIFF
--- a/picoquic/picoquic.h
+++ b/picoquic/picoquic.h
@@ -261,7 +261,7 @@ typedef enum {
     picoquic_callback_path_available, /* A new path is available, or a suspended path is available again */
     picoquic_callback_path_suspended, /* An available path is suspended */
     picoquic_callback_path_deleted, /* An existing path has been deleted */
-    picoquic_callback_path_quality_changed /* Some path quality parameters have changed */
+    picoquic_callback_path_quality_changed, /* Some path quality parameters have changed */
 } picoquic_call_back_event_t;
 
 typedef struct st_picoquic_tp_prefered_address_t {
@@ -1406,7 +1406,8 @@ typedef enum {
     picoquic_congestion_notification_ecn_ec,
     picoquic_congestion_notification_cwin_blocked,
     picoquic_congestion_notification_seed_cwin,
-    picoquic_congestion_notification_reset
+    picoquic_congestion_notification_reset,
+    picoquic_congestion_notification_lost_feedback /* notification of lost feedback */
 } picoquic_congestion_notification_t;
 
 typedef struct st_picoquic_per_ack_state_t {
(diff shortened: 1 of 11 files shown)


picoquic/sender.c
```diff
@@ -4433,6 +4433,39 @@ static int picoquic_select_next_path(picoquic_cnx_t * cnx, uint64_t current_time
...
+                picoquic_packet_context_t* pkt_ctx = (cnx->is_multipath_enabled)?
+                    &path_x->pkt_ctx:&cnx->pkt_ctx[picoquic_packet_context_application];
+                if (pkt_ctx->pending_first != NULL) {
+                    uint64_t delta_sent = (pkt_ctx->pending_first->send_time <= path_x->last_time_acked_data_frame_sent) ? 0 :
+                        (pkt_ctx->pending_first->send_time - path_x->last_time_acked_data_frame_sent);
+                    uint64_t lost_feedback_time = pkt_ctx->highest_acknowledged_time + delta_sent + 2 * cnx->ack_frequency_delay_local;
+                            
+                    if (lost_feedback_time <= current_time) {
+                        path_x->is_lost_feedback_notified = 1;
+                        cnx->congestion_alg->alg_notify(cnx, path_x,
```
suhasHere, line 4454: may be I am missing it , where and how is is_lost_feedback_notified used ?

picoquic/picoquic.h
```diff
@@ -1485,6 +1486,24 @@ void picoquic_set_default_wifi_shadow_rtt(picoquic_quic_t* quic, uint64_t wifi_s
...
+* and thus avoid the queues of less urgent data to delay
+* arrival of urgent real time frames when connectivity is restored.
+* On the other hand, this feature may lower the performance of
+* applications sending lots of data, and thus should only be
+* used when applications require it.
+* 
+* The `should_notify` should be set 1 to enable the feature, or to 0
+* to stop notifications. It is set by default to zero when a connection
+* is created.
+*/
```
suhasHere, line 1504: i wonder of we should  also add the interval of the feedback lost events. From the code, it is 2 * ack frequency + some delta .. Is that right understanding?

huitema: Yes, it is set to twice the maximum delay between receiving a packet and sending an ack. This is a bit arbitrary. The factor 2 is meant to allow for the loss of an ACK, but that not ideal. We could do something complicated, e.g., measuring the actual interval between ACKs and comparing it to the sending interval, maybe doing something like average plus couple of deviations. But that's harder to describe, and probably error prone.

suhasHere (approved):

picoquic/picoquic.h
huitema: I updated the comment to describe the different with RTO.

PR #1681, opened 2024-04-26
Flow control low priority streams

huitema:


huitema:
This PR was motivated by a research idea, that if we allow fewer packets in transit for low priority flows then there will be fewer "priority inversions" during events like Wifi Suspension. The results do not really confirm that hypothesis. The change appears to create two competing effects:

* having fewer packets in transit does limit the priority inversions, and some tests to show a (very small) improvement
* but when the low priority flows are rate limited, their load gets spread overtime, instead of being dealt with immediately when plenty of bandwidth is available. This means more packets lingering over time and thus more priority inversions during network events.

The measurements also show that any random change in scheduling does create random changes in performance, which means that observing small effects can be very misleading.

Given this lack of compelling results, it is probably best to not merge that PR, and avoid increasing the code complexity.

PR #1688, opened 2024-05-02
Fixes issue #1687

### LINKED ISSUE #1687: Getting PICOQUIC_ERROR_IDLE_TIMEOUT even though remote side is sending PINGs
While integrating https://github.com/quic-go/quic-go based client with picoquic, we ran into an issue where picoquic would close the connection due to PICOQUIC_ERROR_IDLE_TIMEOUT.  Both sides are set correctly for keep alvie and idle timeouts.  Both sides are sending PINGs as expected during idle periods.  Picoquic does receive and process the PING, but does not consider the ping as received data, resulting in in idle timeout kicking in.

TimEvens:
Fixes #1687 by updating last received time based on receiving PING.

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -5809,6 +5809,7 @@ int picoquic_decode_frames(picoquic_cnx_t* cnx, picoquic_path_t * path_x, const 
                 ack_needed = 1;
                 break;
             case picoquic_frame_type_ping:
+                cnx->latest_receive_time = current_time;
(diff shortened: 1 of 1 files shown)


huitema (commented):
I would like to understand a bit more what is happening here. The "latest_receive_time" is normally set after each successful packet reception in the function `picoquic_incoming_segment()` in packet.c. The "ping" packets received from QUIC GO should have triggered that update. Do you understand why they did not?

TimEvens:
Close for now, need to investigate it more.

PR #1696, opened 2024-05-21
Fix build on Android

pol51:
Android's NDK doesn't support `pthread_cancel`: use `pthread_kill` instead (only when building for Android)


huitema (approved):
Thanks for the fix. Good to know that picoquic otherwise works on Android!

Issue #1697, opened 2024-05-31
Long period with low CWND on WiFi


huitema:
On stationary Wi-Fi, when testing high defintion real time video, we sometimes observe successions of Wi-Fi suspension, followed by a long period of low bandwidth transmission.

![image](https://github.com/private-octopus/picoquic/assets/7464128/beb6bb7b-be42-4b75-898e-6e188d3c580b)

The series of yellow vertical bars in the RTT graph correspond to "end of suspension" behavior. All the packets that were queued during a Wi-Fi suspension are delivered at once, and acknowledged at very short intervals. We see 19 bars like that between t=11,000 and T=14,150, average duration of interval 166ms. After T=12,000, the RTT is always about 140ms.

When, the RTT drops to 37ms, packets are acked rapidly. There is an exponential process going on, and the bandwidth is restored at T=15535. Interestingly, it takes 37 RTT to increase the bandwidth by a factor 11. That seems to correspond more or less to 3-4 RTT for a 25% increase, which could be a cycle of probe BW Down, Cruise (half the cycles), Refill and Up. Much slower than startup, which would have restored the BW in just 3 or 4 RTT.

The slow "probeBW_UP" behavior does not really correspond to the probeBW_UP spec. In theory, BBR should remain in "probeBW_UP" state as long as the measured bandwidth is growing, and should exit the state if the bandwidth has not been growing for 3 successive RTT. But this is an area were the slide deck presenting BBR differs from the draft. We probably need to revise that part of the code.

The drop in bandwidth after about 10 suspensions is probably spurious. The issue would not have appeared if the 10th and 11th suspension had been processed like the previous ones. This needs to be investigated, understood and fixed.

huitema:
Additional tests point to a couple of issues with the BBR "Probe BW UP" behavior:

* in some conditions such as "application limited", BBR remains in probe UP indefinitely. This is not too good, because the CWND is larger than it should be, and in case of path issue there will be more data loss than necessary.
* in some tests (e.g., "l4s_updown") BBR does not get back to the nominal bandwidth even after the bandwidth goes back to normal.
* in the trace above, BBR uses a very slow process to get back to nominal bandwidth.

huitema:
This is mostly fixed with recent changes in BBR.

PR #1698, opened 2024-06-01
Rework scheduling in socket loop

huitema:
Issue #1697 describes a failure mode of real time communication, with series of apparently "suspended" intervals. These effects may have multiple cause. One of these events happens when both ends are attempting to send high definition video. It seems that the code can become locked into either sending a large number of packets without spending time receiving, or receiving a large number of packets without spending time sending. This PR adds limits to how many packets can be received without sending something, or sent without attempting to receive the latest packet.

The PR also includes updates to the "mediatest" test code, to better simulate the effects of Wi-Fi suspension. That code was instrumental in pointing out that something else than WI-Fi might be creating the problem.


picoquic/sockloop.c
```diff
@@ -764,6 +770,9 @@ void* picoquic_packet_loop_v3(void* v_ctx)
                 }
             }
         }
+        else {
+            nb_loop_immediate++;
+        }
         loop_immediate = 0;
```
suhasHere, line 776: I wonder aren't we resetting loop_immediate to 0 always here ?

huitema: Yes, it will be set to 1 later if there is a reason.

huitema:
@suhasHere I rewrote the comments in the code for clarity.

suhasHere (approved):

PR #1699, opened 2024-06-05
Reenter startup after big BW increase

huitema:
In the "probe BW" states, probing for bandwidth is cautious and slow. This is generally a good idea, but it results in long delays of low bandwidth operation when recovering from a big slow down in capacity. To speed this up, we trigger a reentry of the "startup" state if the measured bandwidth increased by more than 50% during "probeBW".


suhasHere:
Looks good to me

huitema:
Merging as the AppVeyor issue appears spurious.

PR #1700, opened 2024-06-05
Add fake traffic during probe up states

huitema:
This is meant to help app limited connections assess the available bandwidth quickly. It is not clear that it actually helps, we may have to not do this.


huitema (commented):
Merging this, because the function is not enabled by default.

Issue #1702, opened 2024-06-19
[HTTP3] Implementation of range request in picohttp


fbussery:
Right now picohttp doesn't provide the range request.
I did attached to this issue a quick implementation of the range request.
According to [https://datatracker.ietf.org/doc/html/rfc9204#appendix-A](https://datatracker.ietf.org/doc/html/rfc9204#appendix-A)
We just need to implement the qpack code 55

In the attached patch, I just did modify the api of h3zero.h to pass the range

Previous api was:
```
uint8_t * h3zero_create_post_header_frame_ex(uint8_t * bytes, uint8_t * bytes_max, uint8_t const * path, size_t path_length, char const * host, h3zero_content_type_enum content_type, char const* ua_string)

```
new
is
```
uint8_t * h3zero_create_post_header_frame_ex(uint8_t * bytes, uint8_t * bytes_max, uint8_t const * path, size_t path_length, 
uint8_t const * range, size_t range_length, char const * host, h3zero_content_type_enum content_type, char const* ua_string)
```

So, we can pass range this way for instance:
range = "bytes=100-1000"  range_length = 14

I see no regressions with tests: picohttp_ct & picoquicdemo

Note that there is nothing right now in the code that check that range is accepted.

patch: [0001-Implementation-of-byte-range-according-to-rfc9204.patch.txt](https://github.com/user-attachments/files/15900968/0001-Implementation-of-byte-range-according-to-rfc9204.patch.txt)

huitema:
See PR #1703 which is based on the patch. But we need to extend that PR to add actual support of ranges in the demo app, both on client and on server.

huitema:
Fixed in PR #1703

PR #1703, opened 2024-06-23
Implementation of range header

huitema:
Adding an option to encode a range in GET and POST requests.

This is motivated by issue #1702, but the implementation differs somewhat: the new "range" and "range_length" arguments are only added to the "extended" methods `h3zero_create_request_header_frame_ex` and `h3zero_create_post_header_frame_ex`, so as to minimize the churn in existing projects. No support in the "connect" method, because range does not make sense there. Or does it?

This implementation is insufficient. At a minimum, we need unit tests for the frame creation APIs. But it would be better to have a way to exercise the option. Maybe allow "document" arguments in the "picoquicdemo" app to include a range, as in:
~~~
picoquicdemo test.privateoctopus.com 4433 /something#1000-50000
~~~
Then, split the argument into path and range components.

But really, for a complete implementation, we should make sure that the demo server supports the range request, and only returns the requested range. At list for requests, maybe for "post" too, but we need to understand what "range" means for a post.

### DIFF
--- a/picohttp/democlient.c
+++ b/picohttp/democlient.c
@@ -871,6 +871,34 @@ char const * demo_client_parse_post_size(char const * text, uint64_t * post_size
     return text;
 }
 
+char const * demo_client_parse_range(char const * text, char ** range)
+{
+    if (text[0]  != '#') {
+        *range = NULL;
+    }
+    else {
+        char const* range_start = ++text;
+        size_t l_range = 0;
+        
+        while (*text != ':' && *text != ';' && *text != 0) {
+            text++;
+        }
+        l_range = text - range_start;
+        *range = malloc(l_range + 1);
+        if (*range == NULL) {
+            text = NULL;
+        } else {
+            memcpy(*range, range_start, l_range);
+            (*range)[l_range] = 0;
+
+            if (*text == ':') {
+                text++;
+            }
+        }
+    }
+
+    return text;
+}
 
 char const * demo_client_parse_stream_desc(char const * text, uint64_t default_stream, uint64_t default_previous,
     picoquic_demo_stream_desc_t * desc)
@@ -895,6 +923,10 @@ char const * demo_client_parse_stream_desc(char const * text, uint64_t default_s
         text = demo_client_parse_post_size(demo_client_parse_stream_spaces(text), &desc->post_size);
     }
 
+    if (text != NULL) {
+        text = demo_client_parse_range(demo_client_parse_stream_spaces(text), (char **)&desc->range);
+    }
+
     /* Skip the final ';' */
     if (text != NULL && *text == ';') {
         text++;
@@ -913,6 +945,10 @@ void demo_client_delete_scenario_desc(size_t nb_streams, picoquic_demo_stream_de
         if (desc[i].doc_name != NULL) {
             free((char*)desc[i].doc_name);
             *(char**)(&desc[i].doc_name) = NULL;
+        }
+        if (desc[i].range != NULL) {
+            free((char*)desc[i].range);
+            *(char**)(&desc[i].range) = NULL;
         }
     }
     free(desc);
--- a/picohttp/democlient.c
+++ b/picohttp/democlient.c
@@ -226,7 +226,7 @@ int h09_demo_client_prepare_stream_open_command(
 
 static int picoquic_demo_client_open_stream(picoquic_cnx_t* cnx,
     picoquic_demo_callback_ctx_t* ctx,
-    uint64_t stream_id, char const* doc_name, char const* fname, uint64_t post_size, uint64_t nb_repeat)
+    uint64_t stream_id, char const* doc_name, char const* fname, char const* range, uint64_t post_size, uint64_t nb_repeat)
 {
     int ret = 0;
     uint8_t buffer[1024];
@@ -315,8 +315,10 @@ static int picoquic_demo_client_open_stream(picoquic_cnx_t* cnx,
 
         switch (ctx->alpn) {
         case picoquic_alpn_http_3:
-            ret = h3zero_client_create_stream_request(
-                buffer, sizeof(buffer), path, path_len, post_size, cnx->sni, &request_length);
+            ret = h3zero_client_create_stream_request_ex(
+                buffer, sizeof(buffer), path, path_len, 
+                range, (range == NULL)?0:strlen(range), post_size,
+                cnx->sni, &request_length);
             break;
         case picoquic_alpn_http_0_9:
         default:
@@ -397,6 +399,7 @@ int picoquic_demo_client_start_streams(picoquic_cnx_t* cnx,
                 ret = picoquic_demo_client_open_stream(cnx, ctx, ctx->demo_stream[i].stream_id,
                     ctx->demo_stream[i].doc_name,
                     ctx->demo_stream[i].f_name,
+                    ctx->demo_stream[i].range,
                     (size_t)ctx->demo_stream[i].post_size,
                     repeat_nb);
                 repeat_nb++;
--- a/picohttp/democlient.c
+++ b/picohttp/democlient.c
@@ -906,6 +906,8 @@ char const * demo_client_parse_range(char const * text, char ** range)
 char const * demo_client_parse_stream_desc(char const * text, uint64_t default_stream, uint64_t default_previous,
     picoquic_demo_stream_desc_t * desc)
 {
+    memset(desc, 0, sizeof(picoquic_demo_stream_desc_t));
+
     text = demo_client_parse_stream_repeat(text, &desc->repeat_count);
 
     if (text != NULL) {
--- a/picohttp/democlient.h
+++ b/picohttp/democlient.h
@@ -45,6 +45,7 @@ typedef struct st_picoquic_demo_stream_desc_t {
     char const* doc_name;
     char const* f_name;
     uint64_t post_size;
+    char const* range;
 } picoquic_demo_stream_desc_t;
 
 #define PICOQUIC_DEMO_STREAM_LIST_MAX 16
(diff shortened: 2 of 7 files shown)


fbussery (approved):
Wondering if picoquicdemo, instead of 
```
picoquicdemo test.privateoctopus.com 4433 /something#1000-50000
```
could align with curl and use such command:

```
picoquicdemo -H "Range: bytes=0-499" -H "Accept-Encoding: gzip" ... test.privateoctopus.com 4433 /something
```

huitema:
The picoquic demo CLI allows for downloading multiple files. Options like "-H" would apply the same range to all files. Isn't that a problem?

huitema:
The last two commit allow specifying ranges as part of the "scenario" syntax used by picoquicdemo. The h3zero tests verify that this is correctly parsed, the test value being:
~~~
"test.html:#bytes=80-800;/cgi-sink:10000:#bytes=100-1000;"
~~~

huitema:
@fbussery I have not been able to try the range requests in picoquicdemo -- I would need to implement the server side first, and that may take some time. Can you test this PR against a server that supports ranges?

fbussery:
I did test it quicky. And It is not working. I will test your branch in our env today to check if I have  the same status. I run the test in your env this way: picoquicdemo -l log.txt -G -v dash.akamaized.net 443  /akamai/bbb_30fps/bbb_30fps_3840x2160_12000k/bbb_30fps_3840x2160_12000k_12.m4v:#bytes=100-20000

huitema:
Yes, I tried this locally and got an error. This is visible in qlog:

image

Error code 268 (0x10c) is "H3_REQUEST_CANCELLED (0x010c): The request or its response (including pushed response) is cancelled".

huitema:
@fbussery: I added parsing code for the range header, and then used that in a unit test, which works. I tried repeating the test against the akamai server, and it still fails.

I think that we are missing a few steps. I read RFC 7233, and there are a few interesting points:

1. There is a special response code: 206 Partial Content, if the response is a range. If the response is code 200, the whole doc is returned.
2. There is multipart component (one per range) with an indication of which range is server. 
3. Servers may use the "accept range" header to indicate that they accept range requests. However, the RFC is clear that range requests can be sent even if that header has not been received yet.

Do we know who is developing the Akamai server? They may be able to help.

fbussery:
Hello,
I'm very sorry of this bad commit. It is definitively not working and I don't understand how I managed to test it. Still a mistery.
I'm very busy on an other topic now, so I can't work on it these next days.
For your info, I just tested the server akamai with curl http3 and it is working. Here is the following headers:
```
/usr/local/bin/curl -v --http3 -H "Range: bytes=0-103" https://akamaibroadcasteruseast.akamaized.net/cmaf/live/657078/akasource/1719536401/chunk-stream_1-11926.m4s -o outputfile
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0* Host akamaibroadcasteruseast.akamaized.net:443 was resolved.
* IPv6: 2a02:26f0:2b00:22::687d:3ce, 2a02:26f0:2b00:22::687d:3d3
* IPv4: 96.16.122.56, 96.16.122.83
* WARNING: no socket in pollset, transfer may stall!
*   Trying 96.16.122.56:443...
*  CAfile: /etc/ssl/certs/ca-certificates.crt
*  CApath: /etc/ssl/certs
* Server certificate:
*  subject: C=US; ST=Massachusetts; L=Cambridge; O=Akamai Technologies, Inc.; CN=a248.e.akamai.net
*  start date: Apr 18 00:00:00 2024 GMT
*  expire date: Apr 19 23:59:59 2025 GMT
*  subjectAltName: host "akamaibroadcasteruseast.akamaized.net" matched cert's "*.akamaized.net"
*  issuer: C=US; O=DigiCert Inc; CN=DigiCert TLS RSA SHA256 2020 CA1
*  SSL certificate verify ok.
*   Certificate level 0: Public key type EC/prime256v1 (256/128 Bits/secBits), signed using sha256WithRSAEncryption
*   Certificate level 1: Public key type RSA (2048/112 Bits/secBits), signed using sha256WithRSAEncryption
*   Certificate level 2: Public key type RSA (2048/112 Bits/secBits), signed using sha1WithRSAEncryption
* Connected to akamaibroadcasteruseast.akamaized.net (96.16.122.56) port 443
* using HTTP/3
* [HTTP/3] [0] OPENED stream for https://akamaibroadcasteruseast.akamaized.net/cmaf/live/657078/akasource/1719536401/chunk-stream_1-11926.m4s
* [HTTP/3] [0] [:method: GET]
* [HTTP/3] [0] [:scheme: https]
* [HTTP/3] [0] [:authority: akamaibroadcasteruseast.akamaized.net]
* [HTTP/3] [0] [:path: /cmaf/live/657078/akasource/1719536401/chunk-stream_1-11926.m4s]
* [HTTP/3] [0] [user-agent: curl/8.9.0-DEV]
* [HTTP/3] [0] [accept: */*]
* [HTTP/3] [0] [range: bytes=0-103]
> GET /cmaf/live/657078/akasource/1719536401/chunk-stream_1-11926.m4s HTTP/3
> Host: akamaibroadcasteruseast.akamaized.net
> User-Agent: curl/8.9.0-DEV
> Accept: */*
> Range: bytes=0-103
> 
* Request completely sent off
* old SSL session ID is stale, removing
< HTTP/3 206 
< akamai-path-timestamp: i=1719560276.072;xi=1719560276.084;xo=1719560276.981;s=1719560276.990;
< last-modified: Thu, 01 Jan 1970 00:00:00 GMT
< x-akamai-live-origin-qos: d=2000;t=0.000
< x-akamai-server: Akamai-SMT
< pragma: no-cache
< akamai-mon-iucid-ing: 657078
< cache-control: max-age=31532657
< expires: Sat, 28 Jun 2025 07:37:58 GMT
< date: Fri, 28 Jun 2024 08:33:41 GMT
< content-range: bytes 0-103/37838
< content-length: 104
< akamai-mon-iucid-del: 621137
< quic-version: 0x00000001
< alt-svc: h3-Q050=":443"; ma=93600,quic=":443"; ma=93600; v="46,43"
< content-type: video/mp4
< access-control-max-age: 86400
< access-control-allow-credentials: true
< access-control-expose-headers: Server,range,hdntl,hdnts
< access-control-allow-headers: origin,range,hdntl,hdnts
< access-control-allow-methods: GET,POST,OPTIONS
< access-control-allow-origin: *
< 
{ [104 bytes data]
100   104  100   104    0     0   1861      0 --:--:-- --:--:-- --:--:--  1890
* Connection #0 to host akamaibroadcasteruseast.akamaized.net left intact

```

I will investigate further next week I hope.

huitema:
I tried repeating exactly the command that you used, and it still fails. If I remove the range specification, I do get the content of the whole chunk. I suspect that we are missing some negotiation step. It would be good to compare PCAPs of the successful session with curl and the failing session with picoquic, and look for potential differences.

fbussery:
In fact, I found out how to make it work.
Seems that the order of the qpack code matters.
range must be sent after the user agent. (found experimentally)

so, if we apply
```diff --git a/picohttp/h3zero.c b/picohttp/h3zero.c
index f9b28a94..5d6e055f 100644
--- a/picohttp/h3zero.c
+++ b/picohttp/h3zero.c
@@ -903,10 +903,6 @@ uint8_t * h3zero_create_request_header_frame_ex(uint8_t * bytes, uint8_t * bytes
     bytes = h3zero_qpack_code_encode(bytes, bytes_max, 0xC0, 0x3F, H3ZERO_QPACK_SCHEME_HTTPS);
     /* Path: doc_name. Use literal plus reference format */
     bytes = h3zero_qpack_literal_plus_ref_encode(bytes, bytes_max, H3ZERO_QPACK_CODE_PATH, path, path_length);
-    /*Optional: range. Use literal plus reference format */
-    if (range_length > 0) {
-        bytes = h3zero_qpack_literal_plus_ref_encode(bytes, bytes_max, H3ZERO_QPACK_RANGE, (uint8_t const *)range, range_length);
-    }
     /*Authority: host. Use literal plus reference format */
     if (host != NULL) {
         bytes = h3zero_qpack_literal_plus_ref_encode(bytes, bytes_max, H3ZERO_QPACK_AUTHORITY, (uint8_t const *)host, strlen(host));
@@ -915,6 +911,10 @@ uint8_t * h3zero_create_request_header_frame_ex(uint8_t * bytes, uint8_t * bytes
     if (ua_string != NULL) {
         bytes = h3zero_qpack_literal_plus_ref_encode(bytes, bytes_max, H3ZERO_QPACK_USER_AGENT, (uint8_t const*)ua_string, strlen(ua_string));
     }
+    /*Optional: range. Use literal plus reference format */
+    if (range_length > 0) {
+        bytes = h3zero_qpack_literal_plus_ref_encode(bytes, bytes_max, H3ZERO_QPACK_RANGE, (uint8_t const *)range, range_length);
+    }
     return bytes;
 }
```

This works
```
../picoquicdemo -l log.txt -G -v dash.akamaized.net 443 /akamai/bbb_30fps/bbb_30fps_3840x2160_12000k/bbb_30fps_3840x2160_12000k_12.m4v:#bytes=100-2000

...
-rw-rw-r-- 1 fbussery fbussery  1901 juin  28 21:44 _akamai_bbb_30fps_bbb_30fps_3840x2160_12000k_bbb_30fps_3840x2160_12000k_12.m4v
-rw-rw-r-- 1 fbussery fbussery   450 juin  28 21:44 demo_ticket_store.bin
-rw-rw-r-- 1 fbussery fbussery     0 juin  28 21:44 demo_token_store.bin
-rw-rw-r-- 1 fbussery fbussery 10787 juin  28 21:44 log.txt
```

huitema:
Good find, but wow! I can't find that any part of RFC 7233 or 7232...

huitema:
I moved the code around a bit. The "range" header does not have to come after the UA header, but it must be sent after the "host" header. That makes more sense: the document is specified by the combination of path and host, thus both may be needed before a subset of the document can be specified.

PR #1711, opened 2024-07-24
replace removed pthread_setname_np api with prctl on Linux

victorstewart:
pthread_setname_np was removed from glibc in version 2.34 to more strictly align with POSIX standards. this Linux fix works universally. leaving Apple and BSD fixes to someone else if and when this breaks in the future so they can test those solutions.


huitema (approved):
Thanks for the fix, @victorstewart !

PR #1712, opened 2024-07-25
Cmake changes

victorstewart:
1) i made the build fully modular, as i don't personally use any of the other executables or libraries, but left the default build the same

now we can do this:
`-DBUILD_DEMO=OFF -DBUILD_HTTP=OFF -DBUILD_LOGREADER=OFF -DBUILD_LOGLIB=OFF`

2) i added some parameterization to the install functionality

now we can do this:
`-DCMAKE_INSTALL_PREFIX=${LIBRARIES_DIR} -DCMAKE_INSTALL_INCLUDEDIR=${LIBRARIES_DIR}/include/picoquic`

3) i experimented with trying a versioning by commit hash. this doesn't work by default with find_package, and can't/won't unless cmake builds some new functionality... but we can hack our way into this working like so below...

`find_or_build_and_link_package(picoquic 0.0.0-1d680f84 YES)`

```
function(find_or_build_and_link_package PACKAGE_NAME VERSION)

    #...

    if(ARGC LESS 3)
        set(HAS_COMMIT_HASH_TAIL NO)
    else()
        set(HAS_COMMIT_HASH_TAIL YES)
    endif()

    #....

    if (HAS_COMMIT_HASH_TAIL)
        string(REGEX REPLACE "-.*$" "" MINIMUM_VERSION ${VERSION})
    else()
        set(MINIMUM_VERSION ${VERSION})
    endif()

    find_package(${PACKAGE_NAME} ${MINIMUM_VERSION} NAMES ${PACKAGE_NAME} PATHS ${LIBRARIES_DIR} NO_DEFAULT_PATH)

    if (${PACKAGE_NAME}_FOUND)
        if (HAS_COMMIT_HASH_TAIL)
            if (VERSION STREQUAL ${PACKAGE_NAME}_CONSIDERED_VERSIONS)
                message(STATUS "${VERSION} == ${${PACKAGE_NAME}_CONSIDERED_VERSIONS}")
            else()
                message(STATUS "${VERSION} != ${${PACKAGE_NAME}_CONSIDERED_VERSIONS}")
                set(${PACKAGE_NAME}_FOUND NO)
            endif()
        endif()
    endif()

   #...
endfunction()
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,10 @@ find_package(Threads REQUIRED)
 option(DISABLE_DEBUG_PRINTF "Disable Picoquic debug output" OFF)
 option(ENABLE_ASAN "Enable AddressSanitizer (ASAN) for debugging" OFF)
 option(ENABLE_UBSAN "Enable UndefinedBehaviorSanitizer (UBSan) for debugging" OFF)
+option(BUILD_DEMO "Build picoquicdemo" ON)
+option(BUILD_HTTP "Build picohttp" ON)
+option(BUILD_LOGLIB "Build picoquic-log" ON)
+option(BUILD_LOGREADER "Build picolog_t the log reader" ON)
 
 message(STATUS "Initial CMAKE_C_FLAGS=${CMAKE_C_FLAGS}")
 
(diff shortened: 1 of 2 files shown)


huitema:
We already have a github action to test the "picoquic + picotls" variant. How would I write an action that verifies your modifications?

victorstewart:
can you point me to the instructions for that action so i can see what you're referring to?

i didn't change anything in regards to the variants used of either library. i had mentioned to you privately i debated building one fat archive of picoquic + picotls, but i abandoned that idea.

all these changes do is basically 1) enable the option to not build any binary and any library other than picoquic-core and 2) enable installing the picotls static archives if the user elected to build picotls within this build.

huitema:
The github actions use one of two ways to build picoquic. For example, the basic "ci-test" uses:
~~~
      - name: Building picoquic
        run: |
          sudo apt-get install -y libssl-dev
          cmake -S . -B build -DPICOQUIC_FETCH_PTLS:BOOL=ON
          cmake --build build
~~~
That way, the test verifies that the "fetch ptls" option works. 

Then, the test  "CITestsClang" uses the old way to build picotls and picoquic separately:
~~~
      - name: Building picoquic
        run: |
          sudo apt-get install clangd
          sudo apt-get install -y libssl-dev
          echo $CC
          echo $CXX
          $CC --version
          ./ci/build_picotls.sh
          cmake -S . -B build
          cmake --build build
~~~

I was wondering if we could change the "building picoquic" step in another action to test your code, especially the part building a common library that includes picoquic and picotls. Maybe replace the "building picoquic" step in the action "CITestsNoFusion".

But if that's too complicated, we can check your changes as is, and add the github action later.

huitema (approved):
Thanks for this PR. Making the build more modular is indeed helpful!

PR #1713, opened 2024-07-29
Revise stress and fix one fuzzing bug

huitema:
After trying the "stress" option of `picoquic_ct` during IETF 120, we observed that the stress code was behaving strangely when we tried to increase the number of clients. Further analysis pointed a couple of issues:

* there was an initialization issue that caused the stress code to wait 100 seconds (virtual) before starting any connection.
* after fixing that issue, we observed that the elapsed time increased more than linearly with the number of clients.

For simplicity, the old code included a loop through all client contexts to find out what to simulate next, but this was too simple. That code was started each time a packet could be received or sent. The number of packets sent or received increases linearly with the number of clients, and so does the size of the loop. That means we had an O(N2) component. This PR replaces that by organizing the client contexts in a splay. This leaves us with an O(N.log(N)) component, which is acceptable because the constant factor is small. The performance observed on my laptop (Intel core I7) are:

| Nb clients | Wall time | Number of connections |
| ----|----|----|
| 1 | 0.041726 | 53 |
| 2 | 0.16519 | 172 |
| 4 | 0.448717 | 487 |
| 8 | 2.051663 | 2676 |
| 16 | 4.858492 | 6304 |
| 32 | 8.615786 | 12696 |
| 64 | 15.963318 | 23790 |
| 128 | 41.625725 | 60102 |
| 192 | 63.040836 | 87536 |
| 256 | 83.057018 | 108479 |

The "wall time" indicates how long it took for simulating the specified number of clients cycling through connections to the server for 1 minute. Each client will try to set a connection, start up to 4 streams, and then close the connection. Once that connection is done, the client will immediately start a new one. The last column in this table provides the total number of connections that were simulated during the specified "virtual" duration.

The PR also adds new option, "-C ccc", to specified the number of clients (ccc) in stress tests when running picoquic_ct.

The same code is used by the fuzz test, which suddenly became more efficient and found an issue. Test was crashing when a fuzzed packet could not be properly parsed by the "packet repeat" function -- that is, fuzzing voluntarily caused a "local error" in the fuzzing client itself. This has been fixed.

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -2149,6 +2149,7 @@ uint8_t* picoquic_copy_stream_frames_for_retransmit(picoquic_cnx_t* cnx,
 {
     int more_retransmit = 0;
     int packet_dequeued = 0;
+    uint8_t* bytes_first = bytes_next;
     picoquic_packet_t* packet = NULL;
     do {
         packet_dequeued = 0;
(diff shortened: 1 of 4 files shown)


huitema:
Got the all clear from @suhasHere, so merging now.

Issue #1674, opened 2024-04-15
How to get the multipath functionality working on windows?


SupradhaBhat:
The multipath functionality is working in the Linux environment as shown in the screenshot attached, following are the commands used to run the client and server - 

server:
 ./picoquicdemo -c ./ca-cert.pem -k ./server-key.pem -p 4433 -w ./server_files -M 1

client:
./picoquicdemo -M 1 -A 192.168.43.1/3,192.168.43.2/4 -o .\client_files localhost 4433 server_files/index.htm

The above commands set up alternate paths for file transfer in the case of Linux
but similar results are not being shown on Windows. 

My goal is to understand why similar results are not observed while running these commands on Windows.

Linux output:
![Linux output](https://github.com/private-octopus/picoquic/assets/97398229/6370d5fe-2c18-46c0-963e-72c95be4a8fb)

Windows 11 output:
![windows_11 output](https://github.com/private-octopus/picoquic/assets/97398229/a37c79be-b409-45ea-9cdd-2b2e0c2f0c88)

huitema:
Have you tried with a larger file? You may have a timing issue. If the handshake and the file transmission conclude, the connection will be closed before the time comes to establish alternate paths.

SupradhaBhat:
I tried with a file of size about 7000KB but only one path is being used.

huitema:
I just tested the running of picoquicdemo.exe on Windows. To launch the client, I used the command:
~~~
.\picoquicdemo.exe -q cclog -M 1 -A 0.0.0.0/0 -n test ::1 4433 /1000000
~~~
I do see two paths being established, but the traffic flows mostly on path 0. This is a bug in the scheduling algorithm, discussed in issue #1707. It affects both Linux and Windows versions. It should be fixed soon!

huitema:
The option for testing multipath without an alternative address was poorly documented. It turns out that with the previous fixes, a command using the IPv6 NULL address like:
~~~
.\picoquicdemo.exe -D -q cclog -M -A ::0/0 test.privateoctopus.com 4433 /1000000
~~~
... would get the expected behavior, with load split more or less equally between the two paths. But a variation of that command using the IPv4 NULL address like:
~~~
.\picoquicdemo.exe -D -q cclog -M -A 0.0.0.0/0 test.privateoctopus.com 4433 /1000000
~~~
... would not.

I fixed the code so that the two forms of NULL address produce the same result.

Your initial trial with:
~~~
[...]

PR #1720, opened 2024-08-02
Fix multipath simulation in picoquic demo

huitema:
Found the root cause of issue #1674. The option for testing multipath without an alternative address was poorly documented. It turns out that with the previous fixes, a command like:
~~~
.\picoquicdemo.exe -D -q cclog -M -A ::0/0 test.privateoctopus.com 4433 /1000000
~~~
... would work as expected, but if the IPv4 address is used instead, in would not work. No second path would be created with:
~~~
.\picoquicdemo.exe -D -q cclog -M -A 0.0.0.0/0 test.privateoctopus.com 4433 /1000000
~~~

This PR fixes the issue, making sure that both the IPv4 and IPv6 forms are recognized. It also adds a bit of text in the "usage" feedback (See `picoquicdemo -h`).

### DIFF
--- a/picoquicfirst/picoquicdemo.c
+++ b/picoquicfirst/picoquicdemo.c
@@ -577,11 +577,14 @@ int client_loop_cb(picoquic_quic_t* quic, picoquic_packet_loop_cb_enum cb_mode,
                      */
                     int remote_cid_ready = (picoquic_obtain_stashed_cnxid(cb_ctx->cnx_client, 1) != NULL);
                     if (remote_cid_ready) {
-                        struct sockaddr_in6 addr_zero = { 0 };
-                        addr_zero.sin6_family = AF_INET6;
+                        struct sockaddr_in6 addr_zero6 = { 0 };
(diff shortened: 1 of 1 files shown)


arpan21020:
Hello,  

I am trying to run the `picoquicdemo` client using the following command:  

```sh
./picoquicdemo -D -q cclog -M -A ::0/0 test.privateoctopus.com 4433
```  

This works when connecting to `test.privateoctopus.com`. However, instead of using this external server, I want to run the client with the `quicserver` provided within `picoquicdemo`. When attempting to connect to the locally running server, the server crashes.  

I would appreciate any guidance on:  
1. The correct command to start the `quicserver` and connect the client to it.  
2. Potential reasons for the server crash and how to debug the issue.  

Attached is a screenshot of the crash.  
image

Thank you in advance for your help!

huitema:
You are commenting an unrelated PR. Please open a separate issue, so we can document it properly.

PR #1729, opened 2024-08-07
h3zero get GET content types by path

hfstco:
Since we are currently playing with Chrome and the picoquicdemo as a server.

Currently, unless the `h3zero_server_default_page` is requested (`stream_ctx->echo_size`/`length == 0`), the content type `h3zero_content_type_text_plain` is set in the response header for each GET request. Wouldn't it be possible to set the content type via extension?

Screenshot 2024-08-06 at 23 29 29

If I am correct, `h3zero` does not currently support query strings. That's why I haven't added support for it at the moment. If desired, I will include this in another PULL request later.


huitema (approved):
That's nice! It does not increase the size of the code all that much, it makes the server definitely more useful, and there are tests!

huitema:
I entered https://github.com/private-octopus/picoquic/issues/1730 to capture the "query string" discussion.

PR #1731, opened 2024-08-11
Fix RTT computation on first sample

huitema:


huitema:
File issue #1732 for the UBSAN/ASAN error on the satellite_cubic_loss test. It is tied to the management of incoming data buffers, and should be addressed separately from the RTT computation issues.

huitema:
The UBSAN-ASAN issue was easier to fix than I expected. The combination of frames in packets that caused it only happen rarely in "natural" traffic, but could possibly be manufactured by an attacker. The fix to the RTT computation somehow caused that combination to appear.

Close #1732

hfstco (commented):
Looks good. No loss reported for CUBIC.

Verified for netem, Konnect and SkyDSL.

[rtt-fix-results.zip](https://github.com/user-attachments/files/16716582/rtt-fix-results.zip)

PR #1735, opened 2024-08-21
compiler warnings

hfstco:
https://github.com/private-octopus/picoquic/actions/runs/10436424505/job/28901475179

1) fix

I think a few parantheses were forgotten here when solving the RTT problem. Ends in a compiler warning. Please check again whether I have set the brackets correctly.

2) fix

And another warning, **which I haven't fixed yet**, where I unfortunately don't know whether it is intended or not. 
The `bytes` pointer is modified
https://github.com/private-octopus/picoquic/blob/e78af222dab8eb1816d1b505da20dffe557aa3e1/picoquic/frames.c#L1299
and used afterwards,
https://github.com/private-octopus/picoquic/blob/e78af222dab8eb1816d1b505da20dffe557aa3e1/picoquic/frames.c#L1300
which leads to unsequenced modification. Maybe we can move the `elseif` part into the `else` part and introduce a new variable for `bytes + consumed`. But I want to make sure that I have understood this code correctly.

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -4562,10 +4562,10 @@ const uint8_t* picoquic_decode_path_challenge_frame(picoquic_cnx_t* cnx, const u
         bytes += challenge_length;
         if (path_x != NULL && (cnx->is_multipath_enabled ||
             (addr_from == NULL || 
-                (picoquic_compare_addr(addr_from, (struct sockaddr *)&path_x->peer_addr) == 0 &&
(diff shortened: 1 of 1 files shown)


huitema (commented):
I need to review this a bit more. Writing a single condition with multiple parentheses was a bad idea, and the solution is probably to rewrite the entire test for clarity. I will also fix the second issue that you mention.

hfstco:
> I need to review this a bit more. Writing a single condition with multiple parentheses was a bad idea, and the solution is probably to rewrite the entire test for clarity. I will also fix the second issue that you mention.

Probably the best solution.

PR #1736, opened 2024-08-21
Rewrite path challenge test for clarity

huitema:
@hfstco this PR fixes the same issue that you addressed in PR #1735. Instead of taking your fix, it rewrites the code for clarity. Relying on a long IF statement with a mix of AND and OR delimited by parenthesis is too hard to maintain!


hfstco (approved):
Much easier to read.

PR #1737, opened 2024-08-21
Fix code warning in decode stream frame

huitema:
The old code had a test written as: 
~~~
        if (picoquic_stream_network_input(cnx, stream_id, offset,
            fin, (bytes += consumed), data_length, received_data,
            picoquic_is_last_stream_frame(bytes + data_length, bytes_max),
            current_time) != 0) {
            bytes = NULL;
        }
        else {
            bytes += data_length;
        }
~~~
That code worked, but only because compilers were making the right decision to update the bytes variable in the embedded statement `(bytes += consumed)`, and use the updated value in the call to `picoquic_is_last_stream_frame`. But the C language does not really prescribe which of these statement should be executed first -- compilers could for example perform parallel evaluation of all function arguments. Rewriting the code for clarity, and also to suppress the warning.


hfstco (approved):
Looks good. 

Now it is clear that `picoquic_is_last_stream_frame` and the `else` afterwards are also affected by `bytes += consumed`.

PR #1746, opened 2024-09-06
Enforce limit on offset in crypto packets

### LINKED ISSUE #1745: Security vulnerability due to unbounded storage of TLS stream
## Description
picoquic treats TLS streams the same as normal data streams, and there is no capacity limit for them. As a result, their size may grow infinitely since there is no flow control for crypto data in QUIC.

An remote attacker may send an unlimited number of `CRYPTO` frames after handshaking, resulting in unbounded memory usage and leading to a denial-of-service attack.

## Tested versions
Latest git commit (deaf162)

## Suggestion
Limit the capacity of TLS streams and close the connection with a `CRYPTO_BUFFER_EXCEEDED` error code when the limit is exceeded as described in [RFC 9000 - 7.5. Cryptographic Message Buffering](https://www.rfc-editor.org/rfc/rfc9000.html#section-7.5).

huitema:
Close #1745 

The root cause of #1745 was that picoquic did not limit the size of TLS messages, and did not implement the recommendations of [Section 7.5 of RFC 9000](https://www.rfc-editor.org/rfc/rfc9000.html#name-cryptographic-message-buffe), which states: "Implementations MUST support buffering at least 4096 bytes of data received in out-of-order CRYPTO frames. Endpoints MAY choose to allow more data to be buffered during the handshake."

It turn out that the limit to 4096 bytes is problematic. It broke the test "eca1_amplification_loss_test", which checks that the application will limit the amplification factor to 3 even if the server flight is large and we observe packet loss. In our test, the server flight is about 8K, which hopefully matches the large keys of post quantum algorithms. This requires a train of 7 or 8 packets. If the first packet of that train is lost, there will be way more than "4096 bytes of data received in out-of-order CRYPTO frames".

This was fixed by setting the limit to 16KB (16384 bytes).


AsakuraMizu:
Thank you for addressing the issue. I've tested this PR and the fix works as expected. The vulnerability is no longer reproducible.

PR #1749, opened 2024-09-10
Flags for testing BBR extensions

huitema:
@suhasHere this is a first cut at an API for testing the BBR extensions. Please tell me if this is something that you can use.


huitema:
I have renamed the test "do_enter_probeBW_after_new_stream" to  "do_enter_probeBW_after_limited". This matches the code. We may need to rework that in the draft.

PR #1753, opened 2024-09-16
Probe up if app exits app-limited

huitema:
We suspect that if an application remains a long time in "app limited" mode, BBR will "drift down" and the pacing rate will become less and less. The PR forces a transition to "probe BW up"  if the application exist app limited status.


huitema:
Checkin that in, so I can refer to the code in Pr #1749

PR #1754, opened 2024-09-16
cubic seed bug

hfstco:
If we seed a cubic connection we have to calculate `W_max`. Otherwise `K` is calculated with `W_max=UINT64_MAX` and `cwin` is set to a value that is much too high after.

Before:
before-fix

After:
after-fix

Things to discuss:
- added a new test case, on my system it takes round about 4.5 seconds, set max_completion_time to 5 seconds (quictest/tls_api_test.c:3001 [tls_api_one_scenario_body_verify]: Scenario completes in 4663622 microsec, more than 1000000, 2 losses, 0 spurious)

[cubic-seed-bug-qlogs.zip](https://github.com/user-attachments/files/17016587/cubic-seed-bug-qlogs.zip)


huitema (approved):
Nice work!

PR #1757, opened 2024-09-26
Datagram capsule and other picowt work

huitema:
Revising the web transport code (pico webtransport):

1) Verify that datagrams are proposed at H3 level
2) Add code to receive datagram capsules
3) Move "unregister" code from wt_baton to common code in webtransport.c


suhasHere (approved):
Looks good to me. I see few places where a error needs to be returned , but that can come in next.

PR #1761, opened 2024-10-11
Remove delay tests from BBR if min delay lower than 128 microseconds.

huitema:
Loopback tests on Windows showed BBR much slower than Cubic: Cubic achieved transfers at 2 to 3.5 Gbps, while BBR only achieved rates of 1 to 2.4 Gbps.

After multiple experiments and analyses, the root cause seems to be the dependency of BBR on delay measurements. In a loopback test on Windows, the min RTT is about 20 microseconds, but the RTT measurements vary widely. The delays are largely caused by system calls and process or thread scheduling. This introduce random slowdowns in the process.

This PR partially fixes the issue for loopback tests by disabling BBR extensions that exit the startup or probe bandwidth UP phases upon noticing a delay increase. These tests are disabled if the min RTT is less than PICOQUIC_MINRTT_MARGIN -- 128 microseconds. This results in a 50% performance improvement over the base line. The data rates with BB become almost on par with Cubic.

Partial fix and almost on par, because BBR performance also depends on correct estimates of the received bandwidth. These measurements are affected by the randomness of the delays. Fixing that is outside the scope of this PR.


picoquic/picoquic_internal.h
```diff
@@ -74,6 +74,8 @@ extern "C" {
 #define PICOQUIC_MAX_BANDWIDTH_TIME_INTERVAL_MIN 1000
 #define PICOQUIC_MAX_BANDWIDTH_TIME_INTERVAL_MAX 15000
 
+#define PICOQUIC_MINRTT_MARGIN 128
```
suhasHere, line 77: wonder if we should call it as PICOQUIC_MINRTT_MARGIN_MICROSECONDS or add a comment says its unit is in microseconds

huitema: We don't say that for the other variables -- microsec is the default.

suhasHere (approved):

huitema:
@suhasHere Before merging, I added the PICOQUIC_MINRTT_THRESHOLD variable, so we have:

* PICOQUIC_MINRTT_MARGIN: RTT uncertainty derived from application level scheduling
* PICOQUIC_MINRTT_THRESHOLD: Min RTT value for taking RTT into account in congestion control

The 128 microsecond value is probably an OK setting for the first one. For the second one, I kept the 128 microsecond value for now, but we will probably change it over time. The idea is "a value large enough so that measurement errors do not affect the CC algorithms". We will need a series of trials, but we may end up with something between 1 and 4 ms.

PR #1762, opened 2024-10-14
Perf protocol enhancements for media streams

huitema:
We are enhancing the "QUIC Perf" protocol to enable performance testing of multimedia transports. The goal is to test the processing of a realistic multimedia load by the transport stack.

First step is to write down the specification, as part of `doc/quicperf.md`


suhasHere (approved):
LGTM.

I wonder if we can add an example showing how to use the sample scenario with the grammar

huitema:
I am implementing the syntax parser, and I think I need to make some changes.

The old syntax carried two stream-id: the ID of the stream on which the data is send, and the id of the stream after which the data is sent. This allowed description of scenarios such as "send this on stream 4, and when it is all done send that on stream 8". It was convenient for initial tests, but it forced scenarios developers to understand stream numbering, deal with obscure error, and get a narrow view of scenarios.

I naturally added an identifier to the streams, mostly because that's needed for tying scenarios and reports. But if we have alphanumeric stream ids, the concept above can be expressed as "start stream AAA, and then start stream BBB after AAA completes." That's probably easier to understand. The actual stream numbers can be assigned by the stack. In fact they will need to, because if we do "quic stream by subgroup" there will be multiple quic stream for a media stream.

PR #1806, opened 2024-12-11
Update quicperf.md

Karthikdasari0423:
Just a small change in doc


huitema (commented):
Good catch. I copied the definitions from the test code, tried to adjust the extra quotes and semicolons, but obviously did not try hard enough!

Karthikdasari0423:
Thanks @huitema

PR #1810, opened 2024-12-16
Make FindPTLS cmake aware of MBEDTLS flag

sav21153:
When building picoquic with mbedtls, we only need "picotls-core" and "picotls-minicrypto" by default.
Make the FindPTLS cmake file aware of the WITH_MBEDTLS flag and include only the required libraries.


huitema (commented):
I like these changes -- they should result in greater modularity. But i will wait the results of the test run before approving.

huitema (commented):
Thanks for the PR. Merging, since all tests are green.

PR #1815, opened 2025-01-06
Fix function name in documentation

a-andre:


huitema (approved):
Thanks for the fix, @a-andre !

PR #1817, opened 2025-01-09
Fix datagram documentation

a-andre:


huitema (approved):
Good catch. Thank you for the fix!

PR #1824, opened 2025-01-25
Congestion compete test

huitema:
This is the first version of a congestion compete test. The basic design is to simulate the competition between multiple connections, all sharing the same link. The test creates a set of connection, all running the Quic "Perf" protocol. The main connection downloads some megabytes from the server, perf "main" Quic Perf scenario specification. The "background" connection[s] download more data, so they can run longer. The success test is that the main scenario download its data in the expected time. The main connection and the background connection[s] run different algorithms.

This first version has limitations: the link bandwidth and latency are constant, all competing connections share the exact same path and thus the exact same RTT, all background connections run the same cryptographic algorithm. The test setup is not able to simulate "link suspension" or other link events. We might want to look at more data than just a completion time.

As set, we have one test in the test suite per competition scenario, e.g. cubic vs cubic, cubic vs BBR, cubic vs Prague, BBR vs Cubic. We could have a great number of variants, e.g., start the background connection before or after the main connection, etc. I am concerned that adding too many such variants will bloat the test suite. We probably want enough variants to "test the test" and to improve the test coverage of algorithm implementations, but it would probably be nice to export this test tool into a program by itself, that researchers could then use to explore a variety of scenarios.

I expect that we will improve this code before merging it, but it would be nice to have initial feedback.


clarkzjw:
> We probably want enough variants to "test the test" and to improve the test coverage of algorithm implementations, but **it would probably be nice to export this test tool into a program by itself, that researchers could then use to explore a variety of scenarios**.

This indeed would be very helpful in addition to "testing". Some of my thoughts:
1. Maybe we can export this tool into a program such that users can define parameters at least like a) CC algorithms for the "main" and "background" connections, b) number of background connections.
2. I agree we might not want to overcomplicate this tool for very specific scenarios. Hopefully others would fork it for their specific needs, including but not limited to a) different network topology, b) dynamic network link conditions, etc.

huitema:
@clarkzjw any preference between adding a build target for e.g. "pico_cc_test" in the picoquic repo, or creating a separate repo that imports picoquic, etc. ?

clarkzjw:
> @clarkzjw any preference between adding a build target for e.g. "pico_cc_test" in the picoquic repo, or creating a separate repo that imports picoquic, etc. ?

Personally I would prefer the latter, probbaly it is easier for others to understand how it works.
But I can be convinced otherwise, either should be fine.

huitema:
On the TODO list:

1) verify that tests can use the ECN simulation, for example by testing Prague
2) add parameters for speed and latency of link, use in at least one test
3) add an option to get a qlog of the main connection, so we can get nice graphics of at least that connection. May require having a way to specify the ICID, so the name of the log file becomes predictable
4) add a console app that takes a text version of the test specifications as a parameter and produce a nice output of the results.

Next priority:
4) add an option to  request qlogs of all connections. But currently all qlogs start at relative time = 0. We need to have a way to manage time alignment, so we can easily do graphs with multiple connections.

Next priority: 
5) add an option for simulating links with variable profile: wifi, LEO, cellular. It may be sufficient to add an array of event such as "at time t, the bandwidth go to 13 Mbps and the latency goes to 23ms". This is a bit out of the initial "compete" goal, but...

huitema:
I think this PR is ready as is. The code in `cc_compete_test.c` has been updated to simply be a client of `picoquic_ns.c`. A independent test application would use the same API. I will wait until I get at least one review before merging it.

picoquictest/cc_compete_test.c
```diff
@@ -0,0 +1,124 @@
...
+* These tests measure what happens when multiple connections fight for the same
+* resource. The typical scenario involves connections with lots of data to
+* send, with a duration sufficient to test congestion control mechanisms.
+* An example would be testing BBR against itself, against Cubic, and against Reno.
+* We will consider a "connection under test", and treat the other connections
+* as "background".
+*
+* We are not aiming for an exact sharing between the competing connections, but we
+* would like to verify that "nobody starves". The "main" connection should not
+* get a reasonable share of the bandwidth, say at least 25%, and it should also
```
suhasHere, line 38: s/should not/should ?

huitema: yes, should. will fix.

suhasHere (approved):
Looks Good .. Thanks

Issue #1830, opened 2025-02-03
Cubic needs multiple stabilization steps after Hystart


huitema:
Cubic connections start with an initial 'slow start' phase running Hystart. In theory, Hystart stops after sensing delay increases, and then the connection should continue from a safe point. In practice, we see something else, as shown in this snapshot of the initial connection phase in the qlog traces of a cubic run:

![Image](https://github.com/user-attachments/assets/8ff7cdd9-d814-4c9a-a7f7-2bf1112549d5)

We see three steps down, and we also see a series of packet losses continuing well after the first two steps. In this example, the Hystart process exits after the notification of delay increase, and immediately sets "ssthresh", "W_max" and "W_last_max" to that value. This proves over optimistic, and we see an immediate need to further reduce the transmission window. 

The delay increase is not caused by the current value of the CWIN, but rather by the value of CWIN when the last acknowledged packet was sent. The CWIN will increase as acknowledgements are received after this point, but those increase are still "in the pipe line". We should set the safe value to one of:

* the number of bytes in flight after the delay-causing acknowledgement packet was sent,
* or, the total number of bytes acknowledged since that packet was sent,
* or, if we cannot keep track of that, half the current value of the CWIN.

Instead, the apply the coefficient "beta cubic", which may be adequate during congestion avoidance but is not enough to stabilize the connection. This is probably wrong.

There is of course a risk that this is "too drastic", which means we should consider allowing a correction with HyStart++, see issue #1694 and comments from @hfstco.

huitema:
Tried implementing the divide by 2 simplification. Almost all tests pass, but the satellite cubic test completes in 8 seconds instead of 6.1 -- not quite an acceptable performance loss.

huitema:
The first attempt was too simple: we already had code managing the start of long delay connections, already lowering CWIN to a manageable value. We should not divide the window by 2 if that code is already applied. With that fix, all the tests pass. We do not observe performance regression. In fact, avoiding the "rebounds" also avoids packet losses, which improves performance overall. After the changes in PR #1831, we get the following:

![Image](https://github.com/user-attachments/assets/7cf2675d-7f3e-448f-9695-81a1fc914aae)

The changes have two effects:

1) use CWIN/2 as the target upong exit with HyStart
2) enter recovery upon exit, to avoid reacting to packet losses caused by excess CWIN.

hfstco:
Ok, just want to share my thoughts.

In general congestion feedback usually arrives 1 RTT later.

[...]

PR #1831, opened 2025-02-03
Fix cubic start bounce

huitema:
Fix issue #1830


huitema:
@hfstco there are 4 fixes in this PR:

1) Tighten the bound of the `satellite_cubic` and `satellite_cubic_loss` case. The maximum duration parameters were set at 10 seconds, while the actual value was much better. Reflecting the actual value helped detect a potential regression noticed in the discussion of issue #1830.

2) When exiting startup with HyStart, use `path->cwin/2` instead of the full value. This is in line with the discussion in issue #1830, and avoids the succession of packet losses and window decrease observed on the traces.

3) After reducing the CWIN, enter recovery. This is logical, because we expect to experience packet losses. The window was 'just right' when the packet acknowledgment provided an RTT measurement, but it doubled between the moment this packet was sent and the moment the ack was received. We have to wait until after recovery to see signals that match the new value of CWIN.

4) Update the "ack frequency" test. It expect to observe an ACK interval of at least 1500us (in fact 1800), but we now observe an interval just above 1000us. This is 100 times the 10us RTT, probably triggered by the 1ms floor for the ACK frequency. The previous value was probably inflated due to excess delays after HyStart.

picoquic/cubic.c
```diff
@@ -344,13 +344,21 @@ static void cubic_notify(
...
 
                         cubic_state->ssthresh = path_x->cwin;
                         cubic_state->W_max = (double)path_x->cwin / (double)path_x->send_mtu;
                         cubic_state->W_last_max = cubic_state->W_max;
                         cubic_state->W_reno = ((double)path_x->cwin);
                         path_x->is_ssthresh_initialized = 1;
-                        cubic_enter_avoidance(cubic_state, current_time);
+                        /* enter recovery to ignore the losses expected if the window grew
+                        * too large after the acknowleded packet was sent. */
+                        cubic_enter_recovery(cnx, path_x, notification, cubic_state, current_time);
```
hfstco, line 361: Just for clarification. We are entering recovery when HyStart trigger?

hfstco (approved):

hfstco:
> Tighten the bound of the satellite_cubic and satellite_cubic_loss case. The maximum duration parameters were set at 10 seconds, while the actual value was much better. Reflecting the actual value helped detect a potential regression noticed in the discussion of issue https://github.com/private-octopus/picoquic/issues/1830.

Sounds good.

> When exiting startup with HyStart, use path->cwin/2 instead of the full value. This is in line with the discussion in issue https://github.com/private-octopus/picoquic/issues/1830, and avoids the succession of packet losses and window decrease observed on the traces.

> After reducing the CWIN, enter recovery. This is logical, because we expect to experience packet losses. The window was 'just right' when the packet acknowledgment provided an RTT measurement, but it doubled between the moment this packet was sent and the moment the ack was received. We have to wait until after recovery to see signals that match the new value of CWIN.

Shared my thoughts in https://github.com/private-octopus/picoquic/issues/1830. 
But in general I like these changes. Let's see if HyStart++ does better.

huitema:
You are correct about Hystart++. I did not want to mix two issues in one PR, but there is definitely a risk of exiting the slow start too soon due to some transient condition. And then Cubic takes a long time to ramp up and recover.

PR #1837, opened 2025-02-15
Simulate link variation in picoquic_ns

huitema:
Add support for variation of the simulated link bandwidth over time in `picoquic_ns`. This enables us to test specific scenarios such as:

* asymmetry between uplink and down-link
* transient blackhole for 2 seconds.
* link going down to 50% capacity after 1.5 seconds, then going back up after another 2 seconds
* link starting at 50% capacity then going to 100% after 2.5 second
* wi-fi style link going to 10% capacity for 2 seconds, then back up
* wi-fi suspension, 200ms every 2 seconds.


picoquictest/picoquic_ns.c
```diff
@@ -100,12 +100,27 @@ typedef struct st_picoquic_ns_client_t {
...
+} picoquic_ns_link_spec_t;
+
+
 typedef struct st_picoquic_ns_ctx_t {
     picoquic_quic_t* q_ctx[PICOQUIC_NS_NB_NODES];
     struct sockaddr_in addr[PICOQUIC_NS_NB_NODES];
     picoquictest_sim_link_t* link[PICOQUIC_NS_NB_LINKS];
     uint64_t simulated_time;
     int nb_connections;
+    picoquic_ns_link_spec_t* vary_link_spec;
```
clarkzjw, line 120: so if someone wants to create a testing scenario where the network condition is "dynamic", then it is required to pre-define a `vary_link_spec` with the length of `vary_link_nb` and assign corresponding `duration`, `data_rate_in_gbps_down` and `data_rate_in_gbps_up` right?

on the other hand, if someone wants to use picoquic_ns with trace-driven simulations, what will be the best practice to define `picoquic_ns_spec_t`, `picoquic_ns_ctx_t`, etc? by allocating a large `vary_link_spec` with `vary_link_nb` being the number of latency/throughput trace samples?

huitema: The current specification is deliberately kept simple, with an enum list of predetermined link variation patterns. I could think of a few interesting patterns, for example one simulating the periodic overflight of LEO systems. (We know that current algorithms struggle with that.) But that approach requires defining another enum value and adding the pattern generating code in `picoquic_ns.c`. We cannot do too many of that. 

The alternative is obviously to make that more complex, somehow loading a pattern vector and making sure that the conversion understands it. Maybe explore that in a follow-up PR?

clarkzjw: yes, I agree, we can keep picoquic simple as long as we leave APIs to extend it.

huitema: There is a bit of a dilemma. If the extension API was already available, we would not need to bother with an enumeration of variations. We could just have a set of predefined scenarios. The API, could just as well accept a pointer to a predefined scenario or a pointer to a user-specified extension, just like it can accept either a pointer to a predefined CC algorithm or a pointer to a user defined algorithm, as long as they use the same API. The dilemma is thus: develop now a simple hard to extend solution, which will become dead code later, or work a bit more and develop the extensible solution now.

If you look in picoquic_ns.c, the enum is used to generate the series of "varylink" items. For example:
~~~
    case link_scenario_wifi_fade:
        if ((ret = picoquic_ns_create_default_link_spec(cc_ctx, spec, 3)) == 0) {
            cc_ctx->vary_link_spec[0].duration = 1000000;
            cc_ctx->vary_link_spec[1].duration = 2000000;
            cc_ctx->vary_link_spec[1].data_rate_in_gbps_down *= 0.1;
            cc_ctx->vary_link_spec[1].data_rate_in_gbps_up *= 0.1;
            cc_ctx->vary_link_spec[2].duration = UINT64_MAX;
        }
        break;
~~~
There are a couple of design choices here:

* use the global specification of up and down data rate, latency, jitter, etc., to create default link specifications.
* specify duration and modifiers for each step, such as "10% of nominal rate". This allows running the same scenario at multiple data rates.
* only specify modifiers for the variables that do change: data rates in the example, but could be latency as well. For example, we would need to modify both delay and latency steps to emulate the LEO conditions.
* special case of duration as UINT64_MAX if we want to force no more transition after the last step, by opposition to repeating the scenario in a cycle.

If you think that's an OK design, maybe we should come up with a simple way to specify the scenario as text, much like we do for the application scenario.

clarkzjw: > only specify modifiers for the variables that do change: data rates in the example, but could be latency as well. For example, we would need to modify both delay and latency steps to emulate the LEO conditions.

I think probably this will be the better choice, so that latency and data rate could be specified at different time `t` (by specifing the `duration` of the current latency and data rate).

> The API, could just as well accept a pointer to a predefined scenario or a pointer to a user-specified extension

so in the case when users need to specify a more complex and dynamic link scenario other than the existing `picoquic_ns_link_scenario_enum`, instead of the the switch case in `picoquic_ns_create_link_spec`, they can pass a pointer to a custom `vary_link_spec`, I think that would be reasonable.

huitema: My alternative plan would be to get completely rid of the `picoquic_ns_link_scenario_enum`, and have a text specification. The "wifi fade" example could be specified as:

char const * wifi_fade_scenario = "1000000;2000000:D0.1:U0.1;-;"

A LEO scenario could be:
~~~
char const * leo_scenario "1000000:D0.25:U:025:L3;" \
                                         "2000000:D0.5:L1.5;" \
                                         "4000000:D0.75:U0.75:L1.25;"\
                                         "4000000;" \
                                         "4000000:D0.75:U0.75:L1.25;"\
                                         "2000000:D0.5:L1.5;" \
                                         "1000000:D0.25:U:025:L3;"
~~~

Or maybe broken in more steps if we are concerned about discontinuities.

huitema: I need to think a bit more. One of the main scenarios for the "link variation" is to get "replay a trace". For example, in the LEO case, capture a trace showing how latency and data rate vary over time. In such examples, a text based syntax does not help -- in fact, it goes in the way, forcing the trace to be re-encoded as a long character string. It seems much better to assume that the trace is processed by a program that directly produces a sequence of `picoquic_ns_link_spec_t` items. Or, if we want a two step process, a program that processes the trace produces a simple encoding of `picoquic_ns_link_spec_t` items, maybe in CSV format, then the simulation process  loads that file and runs the simulation.

So, maybe "pass a pointer to a custom vary_link_spec" is indeed the most reasonable answer.

huitema:
@clarkzjw Thanks for the review and the comments. I just implemented the suggestion to let users pass a varylink spec, and added a test to verify that it works correctly. I think that's good to merge now.

clarkzjw:
Thanks! This would be really useful.

PR #1841, opened 2025-02-18
Increase success margin of cc_ns_asym test

huitema:
The maximum test completion for the "cc_ns_asym" was set to 7 seconds, based on observations of the test completing in 6,998,120 microseconds. But further observations show that the test duration is a bit random, sometimes completing in just over 7 seconds -- I could observe 7,000,731. The small variations may be due to use of random numbers in the BBR congestion algorithm. Leaving less than 2ms of execution margin is too tight, causes random observations of failure. Increasing to 7.5 second.

### DIFF
--- a/picoquictest/cc_compete_test.c
+++ b/picoquictest/cc_compete_test.c
@@ -139,7 +139,7 @@ int cc_ns_asym_test()
      spec.data_rate_in_gbps = 0.01;
      spec.data_rate_up_in_gbps = 0.001;
      spec.latency = 300000;
-     spec.main_target_time = 7000000;
+     spec.main_target_time = 7500000;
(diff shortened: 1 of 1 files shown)


hfstco:
Another `cc_ns_` test failed for me today too. Think it was `cc_ns_wifi_fade`. After rebuilding the project `cc_ns_asym ` failed. After another rebuild round everything worked fine. First I thought it was my fault. But this PR explains everything.

huitema:
I cannot get `cc_ns_wifi_fade` to fail here -- there is a big execution time margin. But there is indeed a potential issue, because BBR3 only starts ramping up when it it reaches the "probe bandwidth" stage. Next time you have such a failure, please collect a trace!

PR #1844, opened 2025-02-20
Up completion time of cc_compte_dcubic to 10 sec.

huitema:
The current completion time of the "dcubic competes against cubic" test is set to 8.500 sec, but we see examples of runs with simulated time of 8.505 sec. The execution time is a bit random, probably due to the use of floating point. There is no point failing these tests for such issues -- the purpose of the test is to verify the behavior of the simulator, rather than the design of the dcubic algorithm. This PR loosens the time limit to 10 sec.


hfstco (approved):
👍

PR #1845, opened 2025-02-26
New add block API and test

huitema:
Add a new API to "add a block to a stream", which bypasses the malloc and memcpy used by the current "add to stream" API. Add a test for that API.


huitema:
@suhasHere please check the design of the "block sent" callback with your team. Do we have the right parameters? Should we add for example the application context and the per stream context defined in the application callback? Maybe the stream ID?

suhasHere:
> @suhasHere please check the design of the "block sent" callback with your team. Do we have the right parameters? Should we add for example the application context and the per stream context defined in the application callback? Maybe the stream ID?

thanks Christian .. We will review it and get back to you.

huitema:
Feedback from developers so far is that this is not needed. Using this proposed API is perceived as a downgrade from the "just in time" API. The proposed API would commit an application to sending a whole block of data, without meaningful feedback until that whole block is sent, while the just in time API allows for more granular control.

PR #1846, opened 2025-03-04
Fix bit order of port in observed address frame

huitema:


picoquic/frames.c
```diff
@@ -6126,14 +6126,14 @@ const uint8_t* picoquic_decode_observed_address_frame(picoquic_cnx_t* cnx, const
             memset(o_addr, 0, sizeof(struct sockaddr_in));
             o_addr->sin_family = AF_INET;
             memcpy(&o_addr->sin_addr, addr, 4);
-            o_addr->sin_port = port;
+            o_addr->sin_port = htons(port);
```
hawkinsw, line 6129: I believe that I have confirmed in my own mind that, as a result of using shift operations that treat the underlying bits as though they are pure values but store the result into memory according to the endianness of the host, `PICOPARSE_16` will give this value in host order. If that is analysis is correct, then I think that this change is great.

hawkinsw (commented):
Thank you! Assuming that my analysis is correct, below, I think this is exactly what was needed!

picoquic/frames.c
huitema: `PICOPARSE_16` parses a 16 bit integer encode in big-endian order, e.g.., parsing `{ 0x02, 0x01 }` will return 0x201.

hawkinsw: Yes!! Thank you, again, for fixing this little issue so quickly!

PR #1848, opened 2025-03-13
Add padding to packets too small for PN encryption

EndPositive:
When the user disables the padding policy (i.e. minimum padding 0), 1 byte packets may be sent. In this case, there are not enough bytes available for the IV when encrypting the packet number.

See https://datatracker.ietf.org/doc/html/rfc9001#section-5.4.2

### DIFF
--- a/picoquic/sender.c
+++ b/picoquic/sender.c
@@ -843,6 +843,9 @@ size_t picoquic_protect_packet(picoquic_cnx_t* cnx,
     size_t pn_offset = 0;
     size_t pn_length = 0;
     size_t aead_checksum_length = picoquic_aead_get_checksum_length(aead_context);
+    size_t pn_iv_size = picoquic_pn_iv_size(pn_enc);
+    size_t pn_sample_start;
+    size_t pn_sample_end;
     uint8_t first_mask = 0x0F;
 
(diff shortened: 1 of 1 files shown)


huitema (commented):
Thanks @EndPositive for this proposed update. I am not merging it yet, because I am not sure about our reading of https://datatracker.ietf.org/doc/html/rfc9001#section-5.4.2. The text does explain that "packets are padded so that the combined lengths of the encoded packet number and protected payload is at least 4 bytes longer than the sample required for header protection", but I am under the impression that the "protected payload" length includes the length of the AEAD checksum, and that when the checksum length is more than 4 bytes no padding is ever necessary. I am starting a thread on the QUIC mailing list to get feedback on that.

If it is necessary, then we should add an "unpadded" test in the test suite to ensure that the proposed fix is not broken in further updates. But let's wait for the discussion first.

huitema (approved):
Thanks for this fix. Weirdly, we did not encounter the issue before. We should add an "unpadded" test, but that can come in a separate PR.

huitema:
@EndPositive can you take a look at PR #1849 ? It is adding tests of different padding policies, including no padding. No change in code required, all the tests are passing, but this should prevent further regressions.

PR #1849, opened 2025-03-14
Add test of variations of padding policy

huitema:
This is a complement to PR #1848.

We verify that padding works as expected in three scenario: padding policy with min=64, increment=128, with min=0, increment=0, and with min=0, increment=128. In each of these tests, we create a series of packets of different lengths, and verify that the packets sent on the network are padded consistently with the policy.


EndPositive:
Great, thanks for adding that for me 🚀

PR #1853, opened 2025-03-15
Hybla congestion control

vvaale:
I implemented a Picoquic version of the TCP Hybla congestion control algorithm as part of my Master's final project. I thought you might be interested in including it among the other CC algorithms. 

Hybla is a variant of NewReno designed for high-RTT paths. The NewReno cwin normally increases slower the higher the RTT, so Hybla is designed to compensate for that through specific multiplicative factors.

Some resources:

- Original TCP Hybla proposal paper: https://onlinelibrary.wiley.com/doi/10.1002/sat.799
- Implementation in the Linux kernel: https://github.com/torvalds/linux/blob/master/net/ipv4/tcp_hybla.c


huitema:
Nice work. It is great to see implementations of a variety of congestion control algorithms in Picoquic. You should of course make sure that all the tests pass, but there is more. Before merging this work, I would like to have a framework so that we can test a large number of congestion control algorithms, while not inflating the size of executables built with the library.

huitema:
Please take a look at PR #1856 -- managing CC algorithms in a modular way.

picoquic/config.c
```diff
@@ -291,6 +292,16 @@ static int config_set_option(option_table_line_t* option_desc, option_param_t* p
...
+    case picoquic_option_HYBLA_RTT0: {
+        int v = config_atoi(params, nb_params, 0, &ret);
+        if (ret != 0) {
+            fprintf(stderr, "Invalid rtt0 value: %d\n", v);
+        }
+        else {
+            picoquic_hybla_set_rtt0_param(v);
+        }
+        break;
+    }
```
huitema, line 304: I understand that you need a parameter, but I would rather try to find a generic way to pass parameters to CC algorithms. Maybe something like "-g hybla:360".

I would also need some clarity on whether this parameter is required at both clients and servers, or just on one side of the connection. Is it possible to set a default value, and see it use by all connections served by a given server?

picoquic/hybla.c
```diff
@@ -0,0 +1,389 @@
...
+* WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+* DISCLAIMED. IN NO EVENT SHALL Private Octopus, Inc. BE LIABLE FOR ANY
+* DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+* (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+* LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+* ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+* (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+* SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
```
huitema, line 21: No objection to listing as the author, but how does this copyright notice affect the distribution to picoquic users?

picoquic/picoquic.h
```diff
@@ -1572,6 +1572,11 @@ extern picoquic_congestion_algorithm_t* picoquic_fastcc_algorithm;
 extern picoquic_congestion_algorithm_t* picoquic_bbr_algorithm;
 extern picoquic_congestion_algorithm_t* picoquic_prague_algorithm;
 extern picoquic_congestion_algorithm_t* picoquic_bbr1_algorithm;
+extern picoquic_congestion_algorithm_t* picoquic_hybla_algorithm;
+
+// Functions to set Hybla-specific parameters
+void picoquic_hybla_set_rtt0_param(int rtt0);
+void picoquic_hybla_set_initial_ssthresh_param(uint64_t initial_ssthresh);
```
huitema, line 1580: Please check PR #1856. I need to reduce the number of functions and externals listed in picoquic.h.

huitema (commented):
Thanks for contributing an implementation of Hybla to picoquic. My main comment is related to the need for an extensible congestion control framework, and the effort in PR #1856. Reading your code, I see that this effort needs to be extended to provide a generic support for per algorithm parameters. Let's finish the "generic" part first, it will make your code simpler.

By the way, I don't see a test file "picoquictest/hybla_test.c" or any test case. How do we verify that your code works as expected?

vvaale:
Sorry for the delayed response, I'll soon be working on ensuring all tests pass, and I'll implement a hybla_test.c file as well. I'll be looking into PR #1856 as well.

CMakeLists.txt
```diff
@@ -337,6 +338,7 @@ target_link_libraries(picoquic-core
     PRIVATE
         ${OPENSSL_LIBRARIES}
         ${MBEDTLS_LIBRARIES}
+        m
     PUBLIC
```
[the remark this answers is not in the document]
vvaale: It's to link the math library (libm.so, in Linux), since Hybla requires the usage of the pow function. From my understanding, this shouldn't be necessary in Windows, since it's linked automatically.

picoquic/hybla.c
vvaale: All source files have the same comment at the beginning of the file, so I figured I'd do the same in hybla.c, but I didn't know if it was appropriate to add the "Private Octopus, Inc." part. I don't intend to put any license restrictions on my work, so I will include that part.

picoquic/config.c
vvaale: I can see how providing a Hybla-specific parameter as a Picoquic option can be a bit intrusive. As for your question, RTT0 is meant to be a one-sided parameter that influences how the local node calculates the congestion window.

huitema: My plan is to add a "string" (text encoded parameter) after the algorithm ID, using maybe a colon as a separator. Then each algorithm can encode whatever they want in that string, or nothing. The parameter would be passed in the "init" call when a new congestion control context starts.

I now people already want something like that for experimental changes to BBR.

PR #1856, opened 2025-03-16
Make addition of CC algorithms modular

huitema:
We like to use picoquic to investigate congestion control and test multiple algorithms. However, we do not want to load the code of multiple algorithms in every executable program that uses picoquic. For example, some developers are looking at Arduino class devices, and want to make the code as small as possible.

The general approach is to use dynamic linking. If an object file does not include any function required by the program, it will not be linked into the executable. But to achieve that, the program shall not have any reference to the function, including no calls in a rarely used branch, and no addition to the function pointer in a list of "external" objects. We apply this logic to congestion control algorithms as follow:

* remove the list of congestion algorithm pointers in `picoquic.h`
* declare a separate header file for each congestion control algorithm, to be included in the programs using that algorithm
* add a dynamic table of congestion control algorithm, either initialized by a program to the small set that it uses or  initialized by a "register all algorithms" function for programs like `picoquicdemo` that want to surface all available options
* modify the "config" functions to use the table of algorithms registered by the program.

### DIFF
--- a/picoquic/quicctx.c
+++ b/picoquic/quicctx.c
@@ -32,7 +32,7 @@
 #include <time.h>
 #include <errno.h>
 #endif
-
+#include "picoquic_newreno.h"
 
 /*
  * Supported versions. Specific versions may mandate different processing of different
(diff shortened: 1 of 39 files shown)


huitema:
Review of PR #1853 show that we need one more effort to get a fully generic solution: some generic way to pass algorithm specific parameters from the UI/CLI/App to the CC code. Let's work on that before merging this PR.

picoquictest/getter_test.c
```diff
@@ -229,10 +235,13 @@ int getter_test()
...
     }
-
+    /* set the algorithm list to the completevalue before the alogorithm set/get test
+    * Hopefully, nobody is going to call their algorithm "wuovipfwds".
+     */
+    picoquic_register_all_congestion_control_algorithms();
     if (ret == 0) {
         char const* alg_name[] = {
-            "reno", "cubic", "dcubic", "fast", "bbr", "prague", "bbr1", "abracadabra", NULL
+            "reno", "cubic", "dcubic", "fast", "bbr", "prague", "bbr1", "wuovipfwds", NULL
```
hfstco, line 244: I am more a fan of "abracadabra", but the chance that someone calls his CC algorithm "abracadabra" is slightly higher than "wuovipfwds". :) No changes required.

hfstco (approved):

picoquictest/getter_test.c
```diff
@@ -229,10 +235,13 @@ int getter_test()
             ret = -1;
         }
     }
-
+    /* set the algorithm list to the completevalue before the alogorithm set/get test
```
hfstco, line 238: just a typo
`set the algorithm list to the complete value before the algorithm set/get test`

picoquictest/config_test.c
```diff
@@ -285,6 +288,23 @@ static config_error_test_t config_errors[] = {
 
 static size_t nb_config_errors = sizeof(config_errors) / sizeof(config_error_test_t);
 
+
+/* Register a small and stable list of congestion control algorithms,
+* sufficient to test the cc algorithm configurationfunctions.
```
hfstco, line 293: another typo,
`* sufficient to test the cc algorithm configuration functions.`

hfstco:
Just a note from my side:

Because there is a header file for each CC now, we can also consider to define the functions of the CC there. Currently there is no way to do some unit tests with the CC algos. And it should not increase the file size of the executable too.

But I think we could discuss and do that in a separate PR.

picoquic/quicctx.c
```diff
@@ -4909,42 +4909,41 @@ picoquic_cnx_t* picoquic_cnx_by_secret(picoquic_quic_t* quic, const uint8_t* res
...
+    if (alg_name != NULL && picoquic_congestion_control_algorithms != NULL) {
+        for (size_t i = 0; i < picoquic_nb_congestion_control_algorithms; i++) {
+            if (strcmp(alg_name, picoquic_congestion_control_algorithms[i]->congestion_algorithm_id) == 0) {
+                alg = picoquic_congestion_control_algorithms[i];
+                break;
+            }
         }
-        else {
-            alg = NULL;
+        if (alg == NULL && strcmp(alg_name, "reno") == 0) {
```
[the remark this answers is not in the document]
huitema: Yes. AN alternative would be to set the ID to "reno", but I was wondering whether that would break something.

picoquictest/config_test.c
huitema: Fixing

huitema:
@hfstco Thanks for the review. Yes, adding unit tests for CC algos would be a great idea, and documenting the functions to be tested in the header files would be very neat. I am also adding in this PR a generic way to pass an option string to the CC Algo. This is tested in one of the BBR/Wi-Fi tests.

huitema:
Next step after this is to update the `picoquic_ns` project to automatically pick up the definition of the algorithms that we add to the picoquic depot. With that, we should be able to quickly assess new protocols like Hybla...

PR #1860, opened 2025-03-24
Prevent crash if path callbacks are not disabled

alexrabi:
The `picoquic_free` call  may crash if the application has set `cnx->callback_fn` to NULL, and the flag `cnx->are_path_callbacks_enabled` has not been disabled. This PR adds an additional check in `picoquic_delete_path` to prevent crashing.


huitema:
@alexrabi Good catch. I did a quick search and found two other places where we could have the same failure, so I patched that too.

huitema (commented):
Thanks for the fix!
After fixing the other references, it is good.

PR #1861, opened 2025-03-26
Prevent Prague from crashing

alexrabi:
Another quick fix for an unchecked NULL pointer, this time for the Prague congestion control algorithm.

### DIFF
--- a/picoquic/prague.c
+++ b/picoquic/prague.c
@@ -300,7 +300,7 @@ void picoquic_prague_notify(
             /* Regardless of the alg state, update alpha */
             picoquic_prague_update_alpha(cnx, path_x, pr_state, ack_state->nb_bytes_acknowledged, current_time);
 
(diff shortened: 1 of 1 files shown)


huitema:
Thanks, @alexrabi. Any idea how we could improve the test suite and detect such issues?

huitema (approved):
Damn. That's a silly bug!

alexrabi:
> Thanks, @alexrabi. Any idea how we could improve the test suite and detect such issues?

This bug, as well as the issue fixed in PR #1860 were discovered when running a picoquic based application in the Mininet network emulator (https://github.com/mininet/mininet) on a Linux machine. Mininet creates a relatively "realistic" environment for testing since it will actually use the system network stack, but I am not sure how feasible it would be to integrate into the current test suite. It might be worth checking out, however.

Issue #1863, opened 2025-04-03
Build error: `picosocks.c:451:71: error: 'IP_RECVTOS' undeclared (first use in this function)`, `error: 'IP_SENDSRCADDR' undeclared (first use in this function)`


barracuda156:
These are not guaranteed to be available in the OS:

```
[ 21%] Building C object CMakeFiles/picoquic-core.dir/picoquic/picoquic_ptls_openssl.c.o
/opt/local/bin/gcc-mp-14 -DPTLS_WITHOUT_FUSION -I/opt/local/include -I/opt/local/libexec/openssl3/include -I/opt/local/var/macports/build/_opt_local_ppcports_devel_picoquic/picoquic/work/picoquic-14b222e36ecc292eca5645b95a2faf484416afc0/picoquic -I/opt/local/var/macports/build/_opt_local_ppcports_devel_picoquic/picoquic/work/picoquic-14b222e36ecc292eca5645b95a2faf484416afc0/picoquic_mbedtls -pipe -I/opt/local/libexec/openssl3/include -Os -DNDEBUG -I/opt/local/libexec/openssl3/include -isystem/opt/local/include -std=gnu11 -arch ppc -mmacosx-version-min=10.6 -O3 -Wall -fno-exceptions -fno-signed-zeros -frename-registers -fno-trapping-math -MD -MT CMakeFiles/picoquic-core.dir/picoquic/picoquic_ptls_openssl.c.o -MF CMakeFiles/picoquic-core.dir/picoquic/picoquic_ptls_openssl.c.o.d -o CMakeFiles/picoquic-core.dir/picoquic/picoquic_ptls_openssl.c.o -c /opt/local/var/macports/build/_opt_local_ppcports_devel_picoquic/picoquic/work/picoquic-14b222e36ecc292eca5645b95a2faf484416afc0/picoquic/picoquic_ptls_openssl.c
[ 22%] Building C object CMakeFiles/picoquic-core.dir/picoquic/picoquic_mbedtls.c.o
[...]

PR #1864, opened 2025-04-03
picosocks.c: improve fallback code

barracuda156:
Addresses https://github.com/private-octopus/picoquic/issues/1863

### DIFF
--- a/picoquic/picosocks.c
+++ b/picoquic/picosocks.c
@@ -448,7 +448,11 @@ void picoquic_socks_cmsg_parse(
                 }
             }
 #endif
-            else if ((cmsg->cmsg_type == IP_TOS || cmsg->cmsg_type == IP_RECVTOS) && cmsg->cmsg_len > 0) {
+            else if ((cmsg->cmsg_type == IP_TOS
+#ifdef IP_RECVTOS
+                || cmsg->cmsg_type == IP_RECVTOS
+#endif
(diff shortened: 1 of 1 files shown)


picoquic/picosocks.c
```diff
@@ -641,8 +645,15 @@ void picoquic_socks_cmsg_format(
...
             /* The IP_PKTINFO structure is not defined on BSD */
+            /* Some versions of freeBSD do not define IP_SENDSRCADDR, use IP_RECVDSTADDR instead. */
             struct in_addr* pktinfo = (struct in_addr*)cmsg_format_header_return_data_ptr(msg, &last_cmsg,
-                &control_length, IPPROTO_IP, IP_SENDSRCADDR, sizeof(struct in_addr));
+                &control_length, IPPROTO_IP,
+#ifdef IP_SENDSRCADDR
+                IP_SENDSRCADDR
+#else
+                IP_RECVDSTADDR
+#endif
```
huitema, line 655: I added a comment so we understand the motivation for the #ifdef, might be useful for maintenance.

huitema (approved):
Thanks for the fix.

picoquic/picosocks.c
barracuda156: Thank you! (This is also true for some versions of macOS.)

PR #1870, opened 2025-04-15
Add diagnostics to picoquic_ns

huitema:
When the picoquic_ns function runs inside test programs, we want it not verbose. But when it runs into user oriented programs like the "picoquic_ns" solution, we want diagnostics about errors. The solution is to pass a parameter "err_fd" to the function, NULL if no diagnostic is desired, pointing to a FILE descriptor otherwise. For example, pointing to "stderr".

### DIFF
--- a/picoquictest/cc_compete_test.c
+++ b/picoquictest/cc_compete_test.c
@@ -30,6 +30,7 @@
 #include "picoquic_bbr1.h"
 #include "picoquic_fastcc.h"
 #include "picoquic_prague.h"
+#include "picoquic_utils.h"
 
 /* Congestion compete test.
 * These tests measure what happens when multiple connections fight for the same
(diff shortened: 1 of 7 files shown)


huitema:
@hfstco this PR adds diagnostics to picoquic_ns. Mostly useful when running complex scenarios involving multiple media simulations. Please take a look. I am going to merge it if the tests pass, because I need it for other projects, and for a parallel PR in the `picoquic_ns` project, but I will correct issues that you might find in a further PR.

picoquictest/cc_compete_test.c
```diff
@@ -160,11 +161,14 @@ int cc_ns_asym_test()
 char const* cc_compete_media_scenario = "=a1:d50:p2:S:n250:80; \
      = vlow: s30 :p4:S:n150 : 3750 : G30 : I37500; \
      = vmid: s30 :p6:S:n150 : 6250 : G30 : I62500 : D250000;";
+#define MEDIA_TEST_LOG "ns_mediatest_log.csv"
```
hfstco, line 164: Change file extension to .txt, if we only want to write line separated error messages.

hfstco (commented):
Looks good.

picoquictest/cc_compete_test.c
huitema: OK, will do. It is only used in this test, but let's do it right!

PR #1871, opened 2025-04-16
Set TXT suffix for mediatest log file

huitema:


hfstco (approved):
Looks fine for me. As you said, it will not have any effects, but maybe someone will be less confused in the future.

PR #1872, opened 2025-04-20
redundant null check, compare array equal null

hfstco:
redundant null check, compare array equal null

Happy easter!


huitema (approved):
Nice catch.

PR #1881, opened 2025-05-30
Add limited support of ECH to picoquic

huitema:
This first PR adds limited support of ECH to picoquic. As is, the major limitations are:

- Only works with the OpenSSL crypto provider
- Only protects SNI and ALPN
- Not implemented yet in sample apps

The big additions are:

- Add ECH support APIs
- Add end-to-end unit test


huitema:
Todo: simplify the management of tables -- these should be using constant struct declared for each crypto provider. This will enable simple support of minicrypto and mbedtls.

PR #1903, opened 2025-07-22
Application flow control

### LINKED ISSUE #1625: Ability to create back pressure on a stream from the receiver side
Hi, thanks for the great QUIC implementation and please close this issue if there is something in the API I haven't seen. I'm wondering if there's a way for the receiving side of a stream to manually create backpressure on that stream, for example, by _not_ receiving data for some amount of time. It is useful for applications that receive data into fixed size buffers, and want the data to naturally back up and block the sending side of the stream if they can't process it fast enough.

From what I can tell, if I am using `picoquic_packet_loop()`, the library will give me a callback with `picoquic_callback_stream_data` every time it has received data to deliver on a stream, and after this callback is done it will unconditionally consider that data to be consumed by the application. Ideally there would be some way for the application to say "I wasn't able to process all this data" and it would remain unprocessed in the QUIC receive queue.

For an example of what I mean, MsQuic also uses a callback-oriented model to deliver received data, but it [provides a way for the callback to signal back that it did not consume all of the data](https://github.com/microsoft/msquic/blob/main/docs/Streams.md#partial-data-acceptance).

### LINKED ISSUE #1901: Postpone increasing the flow control
Dear Christian,

Upload (fast) ⇒ Picoquic (fast) ⇒ Backend (slow)
The `don't seat on a callback` clause may result in a huge buffer between `picoquic` and the  backend.

Question: Is there a way to postpone the increase of flow-control?
For example, could we instruct Picoquic not to increase the flow control if the buffer to the backend is full?
Once the buffer is back to normal: "Picoquic: please increase the flow control now"

Best wishes,
Marcus

huitema:
Add an API:
~~~
int picoquic_set_app_flow_control(picoquic_cnx_t* cnx, uint64_t stream_id, int use_app_flow_control);
~~~
The API sets a flag `use_app_flow_control` in the stream context. If this flag is set, automatic increases of the per stream flow control window are disabled. The application will need to call `picoquic_open_flow_control` to increment the flow control limits for that stream.

Add the `flow_control_test` to verify that the API is effective. It simulates a download by a receiver that processes the file slowly. Without the API, the application would have to buffer the whole file. With the API, the amount of application buffers required will in general be lower than the value of the relevant `initial_max_stream_data` transport parameter for the stream, plus of course the largest increase specified in `picoquic_open_flow_control`.

Close #1625
Close #1901

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -170,6 +170,7 @@ set(PICOQUIC_TEST_LIBRARY_FILES
     picoquictest/delay_tolerant_test.c
     picoquictest/ech_test.c
     picoquictest/edge_cases.c
+    picoquictest/flow_control_test.c
     picoquictest/getter_test.c
     picoquictest/hashtest.c
     picoquictest/high_latency_test.c


doublex:
Would it be possible to send the peer the accumulated flow increase alongside this function call?
````c
picoquic_set_app_flow_control( use_app_flow_control=0 )
````

huitema:
@doublex QUIC does not carry a flow control window. The Max Stream Data carries a "maximum stream data" parameter that represent the maximum number of bytes that the peer can send on the stream:
~~~
MAX_STREAM_DATA Frame {
  Type (i) = 0x11,
  Stream ID (i),
  Maximum Stream Data (i),
}
~~~
The key notions is that these are absolute numbers (not windows), and that they can only go up -- a frame with a lower value is just ignore. The stack on the receiver side maintains the largest value sent to the peer; the stack on the sender maintains the largest value sent to the peer.

The `picoquic_open_flow_control` has an incremental syntax: take the current value and increase by this much. It is a nice way to enforce the "never regress" constraint, and also to use the "delta" value to increase the "global" flow control (sum of all streams).
We could improve the API to return the new maximum. If `use_app_flow_control=1`, it will only change after next call to `picoquic_open_flow_control`. 

I expect the `picoquic_set_app_flow_control` to be used just once, typically right after opening the flow. It is not really an on-off switch. But yes, it could be enhanced to return the current value of the "max" parameter,

doublex:
Thanks for your explanation - I'm getting old.  :-)
Doesn't the application then have to keep a counter for received data?

Would it be more complicated to design the API like this? 

````c
picoquic_retain_increasing_flow_control( stream_id );
picoquic_now_increase_the_retained_flow_control( stream_id );
````

Greetings from Vienna
Marcus

huitema:
To make your proposal work, the client should have an idea of what the stack will do. Picoquic use exponential growth. Others use fixed flow control window and increase that with acks. Others do an heuristic evaluation of the window, much like flow control, and go with that. These mechanisms are all subject to change over time.

That's why I like the incremental design: call picoquic_open_flow_control to "allow 16K more bytes" (or 32K, or whatever), and repeat when the app has consumed the last 16K. Or implement a complicated algorithm. There is no ambiguity on what the stack will do.

doublex:
If absolute control is required, `picoquic` might not be the right product.
My use case: I don't care whether it's 16k or 1MB.
However, it should not be possible for a malicious sender to generate a 30 GB queue of unprocessed data.

huitema:
@plurex
If you don't care, then the algorithm implemented in the test would suffice:

- call picoquic_set_app_flow_control with use_app_flow_control=1 just after creating the stream,
- Keep a tally of the number of bytes processed by the application,
- When that number is larger than some preset chunk size, call picoquic_open_flow_control to increment the window by that chunk size, and substract the chunk size from the tally number,
- repeat until all bytes in he stream are processed.

I don't think we should try to add parameters to the proposed API -- they work as is. We could add another API to "get the max data parameter" later, if you want a precise handling.

I would rather merge the PR now, unless you object.

doublex:
Thank you for all your hard work.
Best regards!

PR #1914, opened 2025-07-30
Expose per-path byte transfer stats

akh7177:
In the current implementation of PicoQUIC, per-path byte statistics (bytes sent and bytes received) are tracked internally in the st_picoquic_path_t structure but are not exposed via the public API.

This PR intends on extending the st_picoquic_path_quality_t structure in [picoquic.h](https://github.com/private-octopus/picoquic/blob/1a537c8ef1d2c93374ba69f1f41bef85cd4c112e/picoquic/picoquic.h) and updating the picoquic_get_path_quality_from_context() function in [quicctx.c](https://github.com/private-octopus/picoquic/blob/1a537c8ef1d2c93374ba69f1f41bef85cd4c112e/picoquic/quicctx.c) to expose these statistics to applications.

Since this data is already maintained internally, the change is minimal.


huitema (approved):
Thanks for doing this. It is a nice addition to the path quality API.

Issue #1918, opened 2025-08-01
Error during interface re-addition


akh7177:
Hello @huitema,

Based on my understanding of how Picoquic handles multipath, and the discussion [here](https://github.com/private-octopus/picoquic/discussions/1889),  suppose we have two interfaces: A (used to create the initial path0) and B (used to create path1). When interface A goes down, path0 is abandoned and traffic continues over path1 as expected.

However, when interface A comes back up, calling picoquic_probe_new_path_ex() to re-add a path over it fails with PICOQUIC_ERROR_PATH_CID_BLOCKED.

Interestingly, if interface B goes down and comes back, it is successfully re-added as path2 which makes me believe this may not be a simple case of exhausted connection IDs.

Is there a specific reason why re-adding a new path using the interface that was linked to path0 would fail with PATH_CID_BLOCKED? 

Any insight or pointers on resolving this issue would be greatly appreciated.

Thank you!

huitema:
You can never reuse a path-id that was used before and then abandoned.

You can create a new path, with a new path id, as in the example "if interface B (path 1) goes down and comes back, it is successfully re-added as path2".

To continue that example,  "if interface A (path 0) goes down and comes back", it should be re-added with a new path ID -- that is, the next available number. Probably path 3 if you already used path 2 for the re-added path B.

akh7177:
Hello @huitema , thanks for the reply!

I do get that path-id once used cannot be reused. Continuing the above example, interface A should ideally get re-added, utilising a new path but currently it's returning PICOQUIC_ERROR_PATH_CID_BLOCKED when path re-addition using interface A is tried, even when suffecient CIDs are available and this is something unprecedented. However, interface B can be re added without an issue.

Thank you!

huitema:
OK. We need to add a test of this exact scenario in multipath_test, see whether we can reproduce the issue, and understand why a legitimate path id fails.

huitema:
@akh7177 Can you explain exactly what you see? 

We do have in the test suite 2 test cases in which path 0 fails and then comes back. The tests work as follow:

1. set up path 0 and path 1.
2. after a short delay, simulate that the link over which path 0 run is broken (all packets on that path are dropped)
3. verify that after a short delay the traffic is now sent over path 1
4. after a short interval (1 second in one case, 5 second in the other), make the link used for path 0 functional again.
5. verify that after a short delay the traffic is now sent over both path 0 and path 1

[...]

PR #1919, opened 2025-08-01
Fix definition of available path

huitema:
The check for available paths was too simplistic, and only tested for availability of CID for path_id=0. This caused the code to return an error 1099 if the application tried to open a new path after path 0 was abandoned.

Added two test cases in the multipath test suite:

* test `multipath_break3` simulates a disconnected interface, and thus an `address unreachable` signal for the path. Verify that path 0 is abandoned after that.
* test `multipath_back0` also simulates the disconnection of the link under path0, but also simulate that interface going back up and the application requesting the creation of a new path.

The fix required checking the test `multipath_ab1`, which simulates abandoning a series of path_id before the corresponding path was activated. We verify that the test for path number availability checks for this kind of deletion and skips path ids for which the CID were defined and then deleted.


huitema:
Close issue #1918

akh7177 (approved):
This indeed fixes the issue ✨Thanks a lot @huitema

PR #1920, opened 2025-08-03
Multicast / `MC_JOIN` and `MC_STATE` frame

j0nem:
- Implement `MC_JOIN` frame
- Implement `MC_STATE` frame (for multicast layer, not for application layer yet)
- Send `MC_STATE(Joined)` or `MC_STATE(Declined Join)` after receiving `MC_JOIN` frame from server


huitema:
@j0nem I understand you are experimenting with Multicast. That's a great idea, but I am worried about the number of code files that you end up modifying. It would be great if you reformatted this so most of the multicast code would be in its own files, and accessed through a set of function pointers + dedicated multicast context. Also, a compile flag. If you want to pursue that, we can discuss that in an discussion item.

j0nem:
Hi @huitema, thanks for your comment! Actually, I created this PR by mistake and didn't even intend to merge this one back into your main repository (my fork is also quite outdated). You can delete this PR if possible. Sorry, my mouse clicked too fast :grimacing: 

Yes, I am currently experiencing with multicast. Thanks for your suggestions! It's true that a lot of files are changed and it would be very hard to merge them back. Until now, I really was only experimenting and didn't think of merging it at all, but I will consider moving the code in separate files.

Issue #1922, opened 2025-08-04
Infinite loop


doublex:
Dear Christian,

There is an infinite loop if `stream->reset_sent == 1`:

````c
uint8_t * picoquic_format_stream_frame(picoquic_cnx_t* cnx, picoquic_stream_head_t* stream,
    uint8_t* bytes, uint8_t* bytes_max, int * more_data, int * is_pure_ack, int* is_still_active, int * ret)
{
    ...
    if (stream->reset_sent) {
        /* No data will be sent after a reset */
        return bytes;
    }
    ...
}

uint8_t* picoquic_format_available_stream_frames(picoquic_cnx_t* cnx, picoquic_path_t * path_x, uint8_t* bytes_next, uint8_t* bytes_max,
    uint64_t current_priority, int* more_data,
    int* is_pure_ack, int* stream_tried_and_failed, int* ret)
{
    ...
   while (*ret == 0 && stream != NULL && stream->stream_priority <= current_priority && bytes_next < bytes_max) {
        ...
        bytes_next = picoquic_format_stream_frame(cnx, stream, bytes_next, bytes_max, &more_stream_data, is_pure_ack, &is_still_active, ret);
        ...
    }
    ...
}    
````

The GDB callstack is:
````c
picoquic_format_available_stream_frames() at frames.c
picoquic_prepare_stream_and_datagrams() at sender.c
picoquic_prepare_packet_ready() at sender.c
picoquic_prepare_segment() at sender.c
picoquic_prepare_packet_ex() at sender.c
picoquic_prepare_next_packet_ex() at sender.c
picoquic_packet_loop_v3() at sockloop.c
`````

Best wishes
Marcus

huitema:
Damn.

Of course we have tests, and of course they did not find it. The loop can only happen if `picoquic_find_ready_stream_path` returns that stream number, which can only happen if the `has_data` test is true:

~~~
        has_data = (cnx->maxdata_remote > cnx->data_sent && stream->sent_offset < stream->maxdata_remote && (stream->is_active ||
                (stream->send_queue != NULL && stream->send_queue->length > stream->send_queue->offset) ||
                (stream->fin_requested && !stream->fin_sent)));
~~~

If the stream is reset, 'had_data' should be false. So that's definitely one bug. But it also take another condition, such as marking the stream active or queing data after the offset. Most applications don't do that.  Any indication on how you achieved that? That would be useful for getting a repro.

huitema:
[...]

PR #1923, opened 2025-08-04
Try to repro issue #1922

huitema:
Adding a simple reset test to try repro issue #1922. It doesn't. Maybe the test needs a better initialization.


huitema:
Updated the test to use multiple parallel streams, and set priorities so the "reset" stream has the highest one. It did reproduce the infinite loop (issue #1922). Fixed the test used in selecting sending streams, which fixed the issue.

Issue #1924, opened 2025-08-05
Public log headers depend on internal one


a-andre:
* `picoquic_binlog.h` includes `picoquic_internal.h`

* `picoquic_unified_log.h` needs `picoquic_packet_type_enum` which is defined in `picoquic_internal.h`

huitema:
As written, the binlog library does use a bunch of internal types. It is true that we could define a header that just contained:
~~~
/* Set the binary log folder and start generating per connection traces into it.
 * Set to NULL value to stop binary tracing.
 */
int picoquic_set_binlog(picoquic_quic_t* quic, char const* binlog_dir);

/* Enable binary logs, e.g. if autoqlog is requests */
void picoquic_enable_binlog(picoquic_quic_t* quic);
~~~
On the other hand, I am not sure that I want to encourage applications to "enable binlog", by opposition to "enable qlog".

huitema:
`picoquic_unified_log.h` defines the switches between different log providers. It is kinda internal. Do you need to include it?

a-andre:
I wanted to generate log files for https://github.com/private-octopus/picoquic/discussions/1921, that's why I tried to follow `QLOG.md`, which suggests to use `picoquic_set_binlog()` to enable logging.

Using this function, provided by `picoquic_binlog.h`, failed because of the dependency on `picoquic_internal.h`.

Later, I noticed that `picoquic_set_qlog()` from `autoqlog.h` can be used to generate log files.

My suggestion:
* Make `picoquic_logger.h`, `picoquic_binlog.h` and `picoquic_unified_log.h` private headers
* Add information about `picoquic_set_qlog()` to `QLOG.md`

huitema:
Please check the new version of QLOG.md proposed in PR #1927.

a-andre:
Thanks for updating the documentations. Looks good.

huitema:
Solved by PR #1927 and PR #1925

PR #1925, opened 2025-08-06
Make headers which depend on internal ones private

a-andre:
While at it, connect these headers with the picoquic-core target to be visible in IDEs.

This is part of #1924.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,9 +125,6 @@ set(PICOQUIC_CORE_HEADERS
      picoquic/picosocks.h
      picoquic/picoquic_utils.h
      picoquic/picoquic_packet_loop.h
-     picoquic/picoquic_unified_log.h
-     picoquic/picoquic_logger.h
-     picoquic/picoquic_binlog.h
      picoquic/picoquic_config.h
      picoquic/picoquic_lb.h
      picoquic/picoquic_newreno.h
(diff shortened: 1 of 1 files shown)


huitema:
I get the intent, but I am worried about the impact on existing projects such as:

- [picodoq](https://github.com/private-octopus/quicdoq)
- [fuzi-q](https://github.com/private-octopus/fuzi_q)
- [picoquic_ns](https://github.com/private-octopus/picoquic_ns)
- plus those that I don't know about...

In the spirit of not breaking things, we should  leave `PICOQUIC_CORE_HEADERS` as is, so old projects that
depend on the current definition still work, and add a new definition for `PICOQUIC_PUBLIC_HEADERS`,
that can be used by projects desiring to only use stable interfaces.

a-andre:
There shouldn't be any impact on existing projects. The ones you referenced use the headers from the picoquic source directory (and nothing changed there). And projects which might have used the headers from the installed include directory, couldn't do so because of the dependency issues mentioned in #1924.

huitema:
I checked the dependencies in fuzi_q and quicdoq. The main dependency to the CMake file is though the "picoquic finder", such as https://github.com/private-octopus/quicdoq/blob/master/cmake/FindPicoquic.cmake. That finder does not reference the variable PICOQUIC_CORE_HEADERS. So, yes, we are good.

huitema (approved):
OK, I can convince myself that there is no harm to existing projects.

Issue #1926, opened 2025-08-06
Example fails


doublex:
**TL;DR** Everything works when `demo_token_store.bin` and `demo_ticket_store.bin` are deleted.
The problem is 23c372add9bdb510bab3a72270a85949cc589cb8: Add configuration of ECH in QUIC

Steps to reproduce:

Terminal 1:
````sh-session
git clone https://github.com/private-octopus/picoquic.git
cd picoquic
git checkout 23c372add9bdb510bab3a72270a85949cc589cb8
cmake -DPICOQUIC_FETCH_PTLS=Y .
make
./picoquicdemo -p 8123
````

Terminal 2
````sh-session
./picoquicdemo -n test 192.168.1.123 8123 index.html
./picoquicdemo -n test 192.168.1.123 8123 index.html

# second test fails
````

Thanks so much for this amazing project.
Best wishes
Marcus

huitema:
Yes, this bug is easy to repro. The root cause is that picoquic uses ECH, and learns the ECH profile of the server in the first connection. For the connections after that, it uses ECH automatically. In your case, I assume that you used the default certificate provided for test in the build. This uses a large RSA key. The combination of this ECH configuration and the 0RTT ticket increases the size of the "ClientHello", and drives the size of the Initial packet to about 1000 bytes. At that point, a protection kicks in: don't try to coalesce a packet after the Initial if there are not 256 bytes available -- to avoid silly packet syndrome. But then, without the additional coalesced packet, the Initial packet is too short, the server ignores it, and the connection times out.

The solution in PR #1929 is to add random padding at the end of coalesced initial packets if they are too short.

huitema:
There are a couple of related issue. We may want a way to turn off ECH. We may also want to understand why the server is providing an ECH ticket at all if it only uses an RSA key. We will solve that later.

huitema:
I added test of the interaction of ECH and Session Resume or 0RTT in PR #1931, including tests of an ECH capable client. The test show that:

1. If both client and server supports ECH and if both 1st and 2nd connection use proper ECH server RR data, ECH and 0RTT work.
2. If the client does ECH Grease and the server does not support ECH, the connection work but resume/0RTT does not.

The Client Hello carries the session resume ticket as a "PSK Identifier". It does that twice, the real version in the "inner parameter" (or Grease) and a fake version in the "outer". That will certainly increase the size of the Client Hello, and is probably what triggered the latent bug in Picoquic.

PR #1929, opened 2025-08-07
Add random padding if initial is too short

huitema:
Picoquic tries hard to pad initial packets to "just the right size". The problem is that the logic is complicated. It was breaking when the Initial packet was about 1000 bytes long, which happened when trying ECH and 0RTT at the same time. This caused the Initial packet to be too short, and thus be rejected by the server, leading to the symptoms observed in issue #1926.

The solution is to add random padding at the end of the coalesced Initial packet. The padding is not a valid QUIC packet, and is thus discarded by the server without side effect.


a-andre:
The `picoquic_error_names.h` related changes shouldn't be part of this PR.

huitema:
@a-andre you are right. Moving the error names related changes to a different PR.

doublex:
Thank you for the honor.
However, I do not possess the necessary skills to be a reviewer.

suhasHere (approved):

PR #1942, opened 2025-09-17
fix for issue 1769

### LINKED ISSUE #1769: The validity period of the retry token and NEW_TOKEN
https://github.com/private-octopus/picoquic/blob/c0f0313de4b8c35ea4f3193913674361a8b32692/picoquic/packet.c#L1277C1-L1281C6

Hello,  I would like to ask why, when generating the retry token and NEW_TOKEN, the offsets PICOQUIC_TOKEN_DELAY_SHORT and PICOQUIC_TOKEN_DELAY_LONG are added in the calculation of the token validity period? This makes it seem that the validity period of the retry token is 2 minutes and 4 seconds, and the NEW_TOKEN is valid for 48 hours. However, in practical applications, wouldn't it be more appropriate to set the retry token's validity period to 4 seconds and the NEW_TOKEN's validity period to 24 hours? 
Thank you very much for your help!

victorstewart:
Fixes #1769

- Retry and NEW_TOKEN lifetimes were getting padded twice, so both ended up far longer than intended. The retry path added the short delay before calling picoquic_prepare_retry_token, which itself tacked on the configured lifetime. Same story for NEW_TOKEN frames with the long delay.

- picoquic/packet.c:1205 now passes the live current_time into picoquic_prepare_retry_token so the expiry is computed exactly once, matching the documented two‑minute retry window.

- picoquic/sender.c:2664 does the same for NEW_TOKEN generation, keeping the single long-delay application in one place.

- picoquic/tls_api.c:2878 switches the hard-coded microsecond offsets to the shared PICOQUIC_TOKEN_DELAY_* constants, ensuring both call sites derive their lifetimes from the same definitions.

- picoquictest/tls_api_test.c:3669 adjusts the validation test’s schedule so it still probes “just before” and “just after” the corrected expiry thresholds.

**Tests:** 
`cmake -S . -B build -DPROJECT_ENABLE_TESTS=ON - && cmake --build build -j && ctest --test-dir build --output-on-failure`

### DIFF
--- a/picoquic/packet.c
+++ b/picoquic/packet.c
@@ -1203,7 +1203,7 @@ int picoquic_queue_retry_packet(
 
 
     if (picoquic_prepare_retry_token(quic, addr_from,
-        current_time + PICOQUIC_TOKEN_DELAY_SHORT, &ph->dest_cnx_id,
+        current_time, &ph->dest_cnx_id,
         &s_cid, ph->pn, token_buffer, sizeof(token_buffer), &token_size) != 0) {
         ret = PICOQUIC_ERROR_MEMORY;
     }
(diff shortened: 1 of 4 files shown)


huitema (approved):
Thanks for doing this. I added a comment in the prepare token function, because we
set the lifetime as a side effect of setting the odcid parameter. Better explain that so future maintainers are not surprised!

PR #1943, opened 2025-09-17
fix for issue 1782

### LINKED ISSUE #1782: Double decryption on initial packets
It seems that on initial connection, the server decrypts the initial client packet twice, once in `picoquic_screen_initial_packet` and once again in the normal flow. I can't think of a reason why this should be, so thought I'd flag it.

victorstewart:
Fixes #1782: 

avoid double-decrypting Initial packets on the server path. We now only strip header protection during initial screening and leave payload decryption to the normal receive flow, eliminating redundant AEAD work while preserving token validation and handshake checks.

### DIFF
--- a/picoquic/packet.c
+++ b/picoquic/packet.c
@@ -126,16 +126,8 @@ int picoquic_screen_initial_packet(
 
         if (picoquic_get_initial_aead_context(quic, ph->version_index, &ph->dest_cnx_id,
             0 /* is_client=0 */, 0 /* is_enc = 0 */, &aead_ctx, &pn_dec_ctx) == 0) {
-            ret = picoquic_remove_header_protection_inner((uint8_t *)bytes, ph->offset + ph->payload_length,
+            ret = picoquic_remove_header_protection_inner((uint8_t*)bytes, ph->offset + ph->payload_length,
                 decrypted_bytes, &dph, pn_dec_ctx, 0 /* is_loss_bit_enabled_incoming */, 0 /* sack_list_last*/);
-            if (ret == 0) {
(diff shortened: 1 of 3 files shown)


picoquic/packet.c
```diff
@@ -126,16 +126,8 @@ int picoquic_screen_initial_packet(
...
                 decrypted_bytes, &dph, pn_dec_ctx, 0 /* is_loss_bit_enabled_incoming */, 0 /* sack_list_last*/);
-            if (ret == 0) {
-                size_t decrypted_length = picoquic_aead_decrypt_generic(decrypted_bytes + dph.offset,
-                    bytes + dph.offset, dph.payload_length, dph.pn64, decrypted_bytes, dph.offset, 
-                    aead_ctx);
-                if (decrypted_length >= dph.payload_length) {
-                    ret = PICOQUIC_ERROR_AEAD_CHECK;
-                }
-            }
         }
```
huitema, line 186: That does not work so well. We do not use the `decrypted bytes` here, but the test on `decrypted_length` is important. It will detect whether the packet was mangled in transit, which is rare, or whether it was ill formed by the sender, which could well happen in case of DOS attacks. Without that test, we would generate a connection context (line 183 in current code) even if the packet is malformed. Checking the checksum forces would be attackers to spend CPU cycles properly encrypting the packet, which makes attacks significantly more costly than just filling the packet with random bytes.

victorstewart:
this should be the implementation you had in mind

huitema (changes requested):
Let's stop here. This is way too much code for a small problem. I will take that PR as inspiration, and work on my own fix.

huitema:
Overtaken by events, and by PR #1947 and #1948. Also, I would rather never use the "goto exit" construct. Or "goto" in general.

PR #1945, opened 2025-09-17
fix 1804

### LINKED ISSUE #1804: Sockets on demand
As discussed in issue #1778, it would be nice if the socket loop code could create sockets on the fly if the application tries to use a new port number that does not yet have a matching socket.

victorstewart:
Fixes #1804:

- teach the socket loop to create missing UDP sockets when `picoquic_prepare_next_packet_ex` wants to send through a port that isn’t yet bound, reusing `picoquic_packet_loop_open_socket` and respecting `PICOQUIC_PACKET_LOOP_SOCKETS_MAX`

- keep all existing sockets untouched while dynamically extending the pool so new paths/ports can transmit without pre-provisioning

### DIFF
--- a/picoquic/sockloop.c
+++ b/picoquic/sockloop.c
@@ -988,6 +988,29 @@ void* picoquic_packet_loop_v3(void* v_ctx)
                     }
 
                     if (send_socket == INVALID_SOCKET) {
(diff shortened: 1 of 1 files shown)


huitema (approved):
Thanks for doing this. The code looks good.

PR #1948, opened 2025-09-20
Do not compute initial AEAD decryption key twice.

huitema:
This is the second part of the fix for issue #1782. The fix in PR #1947 ensured that packets would not be decrypted
twice, but did not pass the computed AEAD and PN decryption keys to the new context, which means the
key generation had to be done twice. This PR fixes that last point. It introduces an "internal" variant
of connection context creation that takes as additional parameters the AEAD and PN decryption contexts,
and avoids recomputing them, and it makes sure that this variant is called when creating a connection context
in the `picoquic_screen_initial_packet` function.

We have to pay attention to the way memory is handled to not have either crashes or memory leak. The convention is that if the call to `picoquic_create_cnx_internal` succeeds, the objects pointed to by the AEAD and PN decryption handles
is `owned` by the connection context, but if it fails it remains owned by the calling code and should be cleared
inside for example `picoquic_screen_initial_packet`.

### DIFF
--- a/picoquic/packet.c
+++ b/picoquic/packet.c
@@ -94,6 +94,8 @@ int picoquic_screen_initial_packet(
     picoquic_stream_data_node_t* decrypted_data)
 {
     int ret = 0;
+    void* aead_ctx = NULL;
+    void* pn_dec_ctx = NULL;
 
     /* Create a connection context if the CI is acceptable */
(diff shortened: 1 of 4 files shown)


suhasHere (approved):

victorstewart (approved):
these look great. you shrunk it down a lot! human > AI for now haha.

PR #1959, opened 2025-10-16
Testing simulation of dual_q AQM

### LINKED ISSUE #1825: L4S simulation is simplistic
The test code has minimal support for L4S-style active queue management: each link description has an `l4smax` property. If the queue size exceeds that value, the incoming packets are marked as `congestion experienced`. This allows us to test the Prague or ECN code paths, but it is not quite the "Dual-Queue Coupled AQM" described in [RFC 9332](https://www.rfc-editor.org/rfc/rfc9332.html):

* There should be two queues, the L4S queue for packets with ECT(1) or CE marks, the default queue for other packets.
* The algorithm maintains a "marking rate" to be applied to L4S packets.
     - The marking rate is driven by a proportional controller, increasing when the L4S queuing time exceeds the target, decreasing when it is lower.
     - L4S packets are marked according to the marking rate
* The algorithm derives an "early drop rate" for the default queue, computed as the square of the marking rate.
    - According to specification, packets with ECT(0) are marked with the early drop rate. Packets without ECN marks are dropped at that rate.
* If there is no L4S traffic, the marking rate returns to zero
* If the queue is completely full, packets are dropped.

The multiple queue requirement is hard to satisfy in our architecture. The current design has just one queue, serviced first-in first-out. If we introduce two queues, we have to decide a service rate for each of the queues. The L4S specification envisages giving priority to the L4S queue, but that does not seem fair if there is just one connection using the L4S queue and many using the default queue.

The handling can be simplified by using just one queue, but computing a marking rate based on the delay experienced by L4S packets. We could update the marking rate for each incoming L4S packet. This is wrong because if there is no incoming L4S packet at all, the marking rate remains constant. We could update the marking rate periodically, based on the longest delay experienced by L4S packet during that period -- or zero if there are no L4S packets. We could also update the marking rate for every incoming packet, L4S or not, which is more robust. It does have the downside of preventing non L4S traffic from building queues, which might slow it somewhat -- but that's logical if we only manage a single queue.

huitema:
Adding support in the simulations for three Active Queue Management modes: RED, Rate Controlled, and L4S. This requires reorganizing the simulation code a bit.

Also revising the implementation of the Prague CC algorithm, to try understand how L4S should really behave. This is not exactly a great success, probably because Prague is rather hard to tune. I had to loosen the test for the Prague "compete under L4S" tests because Prague it is highly sensitive to packet loss or CE mark received during the startup phase. I also loosened the Prague "satellite" test.

The L4S implementation follows RFC 9332. The tests are underwhelming. L4S is a pretty complex algorithm, very sensitive to initialization parameters. The current picoquic code is tuned to send batches of packets, which is great for performance but creates an instant queue. If the L4S marking threshold is too small, each packet train will cause a CE mark or two, and thus a performance drop. On the other hand, if the threshold is too high, there will not be enough CE marks to force the competing connections to behave. Or, the CE marks will come too late, the buffers will fill up, and packets will be dropped.

This PR is about getting support for the AQM mechanism. We will need another PR in `picoquic_ns` to add the specification syntax to declare use of AQM in specific scenarios. We probably also want a simple RED-ECN AQM for comparison with L4S, and maybe also something like FQ-Codel. But that's for another set of PRs.

Similarly, we may want to tune the existing CC algorithms so they only send long trains of packets when working at high speed, probably by tuning the way Pacing Rate and "Quantum" are computed. But that will have consequence for performance, so that's a different investigation.

Close #1825

### DIFF
--- a/picoquic/prague.c
+++ b/picoquic/prague.c
@@ -202,15 +202,10 @@ static void picoquic_prague_enter_recovery(
     if (pr_state->ssthresh < PICOQUIC_CWIN_MINIMUM) {
         pr_state->ssthresh = PICOQUIC_CWIN_MINIMUM;
     }
+    
+    path_x->cwin = pr_state->ssthresh;
+    pr_state->alg_state = picoquic_prague_alg_congestion_avoidance;
 
-    if (notification == picoquic_congestion_notification_timeout) {
-        path_x->cwin = PICOQUIC_CWIN_MINIMUM;
-        pr_state->alg_state = picoquic_prague_alg_slow_start;
-    }
-    else {
-        path_x->cwin = pr_state->ssthresh;
-        pr_state->alg_state = picoquic_prague_alg_congestion_avoidance;
-    }
     picoquic_prague_initialize_era(cnx, path_x, pr_state, current_time);
 }
 
@@ -305,12 +300,12 @@ void picoquic_prague_process_start_ack(picoquic_cnx_t* cnx,
     if (pr_state->ssthresh == UINT64_MAX) {
         /* Increase cwin based on bandwidth estimation. */
         path_x->cwin = picoquic_cc_update_target_cwin_estimation(path_x);
-        if (pkt_ctx->ecn_ce_total_remote > pr_state->l4s_epoch_ce) {
-            /* CE mark received in intitial state:
-             * exit and enter recovery.
-             */
-            picoquic_prague_enter_recovery(cnx, path_x, picoquic_congestion_notification_ecn_ec, pr_state, current_time);
-        }
+    }
+    if (pkt_ctx->ecn_ce_total_remote > pr_state->l4s_epoch_ce) {
+        /* CE mark received in intitial state:
+         * exit and enter recovery.
+         */
+        picoquic_prague_enter_recovery(cnx, path_x, picoquic_congestion_notification_ecn_ec, pr_state, current_time);
     }
     else {
         path_x->cwin += picoquic_cc_slow_start_increase_ex2(path_x, ack_state->nb_bytes_acknowledged, 0, pr_state->alpha);
--- a/picoquictest/dualq_aqm_test.c
+++ b/picoquictest/dualq_aqm_test.c
@@ -198,6 +198,10 @@ int dualq_dequeue_test()
         ret = -1;
     }
 
+    if (ret == 0 && dqt_ctx.simulated_time > max_time) {
+        ret = -1;
+    }
+
     dualq_test_release_ctx(&dqt_ctx);
 
     return ret;
--- a/picoquictest/satellite_test.c
+++ b/picoquictest/satellite_test.c
@@ -307,7 +307,7 @@ int satellite_dcubic_seeded_test()
 int satellite_prague_seeded_test()
 {
     /* TODO check max_completion_time */
-    return satellite_test_one(picoquic_prague_algorithm, 100000000, 7500000, 250, 3, 0, 0, 0, 1, 0, 0);
+    return satellite_test_one(picoquic_prague_algorithm, 100000000, 5300000, 250, 3, 0, 0, 0, 1, 0, 0);
 }
 
 /* Satellite loss interop test, as shown in https://interop.sedrubal.de/
--- a/picoquictest/l4s_test.c
+++ b/picoquictest/l4s_test.c
@@ -140,8 +140,7 @@ int l4s_prague_test()
 {
     picoquic_congestion_algorithm_t* ccalgo = picoquic_prague_algorithm;
 
-    /* TODO increased max_completion_time for 100ms, because of "app limited" changes */
-    int ret = l4s_congestion_test(ccalgo, 1, 5600000, 7, 4500, 0, NULL);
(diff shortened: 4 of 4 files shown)


picoquictest/dualq_aqm.c
```diff
@@ -30,70 +30,452 @@
...
+}
+
+
+/* Submit: implement the sim link API */
+void dualq_submit(picoquictest_aqm_t* self, picoquictest_sim_link_t* link,
+    picoquictest_sim_packet_t* packet, uint64_t current_time, int* should_drop)
+{
+    dualq_state_t* dualq = (dualq_state_t*)self;
+#if 1
+    int x = 0;
```
hfstco: What is `x` used for? Debug?

picoquictest/dualq_aqm_test.c
```diff
@@ -0,0 +1,407 @@
...
+/* Dequeue test: basic check of the dequeue API */
+int dualq_dequeue_test()
+{
+    dualq_test_ctx dqt_ctx;
+    int ret = dualq_test_get_ctx(&dqt_ctx);
+    uint8_t ecn_sequence[] = { 0, PICOQUIC_ECN_ECT_0, 0, PICOQUIC_ECN_ECT_1, PICOQUIC_ECN_CE };
+    int queue_id[] = { 0, 0, 0, 1, 1 };
+    int trials = 0;
+    int received = 0;
+    uint64_t max_time = 100000;
```
hfstco, line 154: Is this test limited by `max_time`?

picoquictest/satellite_test.c
```diff
@@ -307,7 +307,7 @@ int satellite_dcubic_seeded_test()
 int satellite_prague_seeded_test()
 {
     /* TODO check max_completion_time */
-    return satellite_test_one(picoquic_prague_algorithm, 100000000, 5300000, 250, 3, 0, 0, 0, 1, 0, 0);
+    return satellite_test_one(picoquic_prague_algorithm, 100000000, 7500000, 250, 3, 0, 0, 0, 1, 0, 0);
```
hfstco: What is the reason for this increase in completion time?

picoquic/prague.c
hfstco: We won't enter `picoquic_prague_enter_recovery()` on `picoquic_congestion_notification_timeout.`

picoquic/prague.c
```diff
@@ -194,88 +211,121 @@ static void picoquic_prague_enter_recovery(
...
+}
 
-                pr_state->alpha_shifted += delta_frac;
-                pr_state->alpha = pr_state->alpha_shifted >> PRAGUE_SHIFT_G;
-            }
+void picoquic_prague_process_start_ack(picoquic_cnx_t* cnx,
+    picoquic_path_t* path_x, picoquic_prague_state_t* pr_state, picoquic_per_ack_state_t* ack_state, uint64_t current_time)
+{
+    picoquic_packet_context_t* pkt_ctx = picoquic_prague_get_pkt_ctx(cnx, path_x);
+    if (pr_state->ssthresh == UINT64_MAX) {
```
hfstco, line 300: Maybe I have missed something.
Is it possible that `ssthresh` is set in Slow Start? Do we really want to adjust the cwin in Slow Start based solely on bandwidth estimation? 
`picoquic_cc_slow_start_increase_ex2()` is called if `ssthresh` is set only.

picoquictest/satellite_test.c
huitema: I think this is due to exiting slow start too soon due to an ECN-CE signal from the L4S AQM.

picoquictest/dualq_aqm.c
huitema: yes, debug. I need to clean that out!

picoquictest/dualq_aqm_test.c
huitema: It should be! Fixing that. Not a big issue in practice, but the test could catch a bug some day in the future.

picoquic/prague.c
huitema: Correct. This is a leftover from the previous version, in which we would do special code on "loss due to timeout". But with RACK, the only action on timeout is to send a PTO probe, and that must not be treated as a loss for congestion purpose.

huitema: Removing the unused code.

picoquic/prague.c
huitema: You are right to flag that code. Something was missed when I refactored this from the old code inline in the switch/case picoquic_congestion_notification_acknowledgement, and bits of "update alpha". There is no real point of updating alpha in slow start, because the code will exit slow start on first CE received, but that does not mean just trusting the bandwidth estimate!

picoquictest/satellite_test.c
huitema: Actually, that was due to the poor refactoring of ACK processing during startup that you also flagged. Reverting to the previous value after the fix.

picoquic/prague.c
huitema: Fixing that also fixed the perf issue in the satellite test.

hfstco (approved):

PR #1963, opened 2025-10-30
Update Webtransport version, Implement Reset Stream At, Fix H3 bugs

huitema:
Update Webtransport version from -02 to -07

Implementation of the "Reset Stream At" extension, which is required for Webtransport version -07

Fix a race condition between the queuing of H3 Settings and the decoding of the "max unidir" transport parameter, which could cause the H3 "settings" stream to not be sent.


suhasHere (approved):
I plan to test it some more. But this seems fine to land

PR #1974, opened 2025-11-11
Add sample server for wt baton

huitema:
This is meant as sample ode for developing Web Transport based apps.

### DIFF
--- a/baton_app/baton_app.c
+++ b/baton_app/baton_app.c
@@ -495,10 +495,6 @@ int wt_baton_server(char const* path, picoquic_quic_config_t* config)
     printf("Server exit, ret = 0x%x\n", ret);
 
     /* Clean up */
-    if (config->cnx_id_cbdata != NULL) {
-        picoquic_lb_compat_cid_config_free(qserver);
-    }
-
     if (qserver != NULL) {
         picoquic_free(qserver);
     }


baton_app/baton_app.c
```diff
@@ -387,3 +409,130 @@ int baton_client_loop_cb(picoquic_quic_t* quic, picoquic_packet_loop_cb_enum cb_
...
+    memset(&picoquic_file_param, 0, sizeof(picohttp_server_parameters_t));
+
+    /* Configure the web server parameters, including the "path table"
+    * with a path item describing the baton app.
+    * The "path_app_context" is null because the baton app does not use it.
+     */
+    picoquic_file_param.web_folder = config->www_dir;
+    path_item_list[0].path = path;
+    path_item_list[0].path_length = strlen(path);
+    path_item_list[0].path_callback = wt_baton_callback;
```
suhasHere, line 444: Should we check if the service was indeed baton here before assinign the path_callback ?

suhasHere (approved):

baton_app/baton_app.c
huitema: No. This is the configuration set up. The code is declaring that specified path is running the "devious baton" service. You could set the "path" to whatever you need to configure, "baton", "devious", "devious-baton", or whatever else you like. In particular, you could pick some obscure value like "l33t-bAt0n" to act as some kind of secret...

PR #1981, opened 2025-11-18
ECN support in C4 and test fixes

huitema:
Implementation of ECN support in C4. The code computes `ecn_alpha` as an exponentially weighted average of fraction 'ce_marks / (ce_marks + ect1_marks', and detects congestion if that ratio passes a sensitivity dependent threshold. The amount of congestion back-off depends of `ecn_alpha`, mimicking the behavior of TCP Prague. The amplitude of the "pushing" trials also depend on the current `ecn_alpha`, avoiding pushing too much in case of congestion.

To properly test, we had to fix two bugs:

1) Avoid taking into account the measurements done on "initial" or "handshake" packets, because they are not reliable.
2) Fix the computation of queuing delays in the dual queue ACM, which was not properly taking into account the time passed in the L4S or Classic queues.

The fixes change the results of the dual queue and L4S tests. 

This PR provides a good basis for further work on both L4S/DualQ and C4.

### DIFF
--- a/picoquic/c4.c
+++ b/picoquic/c4.c
@@ -853,7 +853,11 @@ static void c4_notify_congestion(
         beta = (C4_BETA_LOSS_1024 + MULT1024(c4_sensitivity_1024(c4_state), C4_BETA_LOSS_1024))/2;
     }
     else if (c_mode == c4_congestion_ecn) {
-        /* Apply proportional reduction. Question: should it be sensitivity related? */
+        /* Apply proportional reduction.
+         * The threshold value is a function of sensitivity, and thus the reduction
+         * incorporates a sensitivity factor. There is no obvious need for
+         * additional sensitivity related factors.
+         */
         beta = (c4_state->ecn_alpha - c4_state->ecn_threshold) * 1024 / c4_state->ecn_threshold;
         if (beta > C4_BETA_LOSS_1024) {
             /* capping beta to the standard 1/4th. */


huitema:
The latest commit comes after an intensive review of the DualQ implementation and its integration in the simulation infrastructure. It fixes multiple issues. It also add an "ecn" attribute in the logging of received packets, so we can audit the arrival of ECN marks.

huitema:
Finally, tune the L4S-ECN parameters in two ways.

First, lower the threshold for L4S "ecn_alpha" from [128..256] (depending on sensitivity) to [96..192]. Early tests show that higher threshold values like [112..224] or [128..256] result in lower data rates, and that lower threshold values like [80..160] or [64..128] may sometimes result in higher value, but also in much bigger performance variations.

Second, refine the idea of making the push level a function of the background ECN level. Leave it at 6.25% if C4 would select a "low level" push, make it vary between 25% and 6.25% if C4 would have selected a "high level" push. This does improve performance significantly, and also reduces the variation between samples.

huitema:
One final fix: C4 was setting the "pacing quantum" to a high value. Setting the quantum to 4ms avoids most marking, resulting in good performance in presence of ECN.

huitema:
Well, previous fix was not final after all. Found out that the "limit" parameter should be set as a function of the "max queue size" in the link specification, rather than the loose language in the dualq aqm draft. With that, we see much fewer packet losses, and the L4S test results are much more plausible.

picoquic/c4.c
```diff
@@ -788,6 +852,14 @@ static void c4_notify_congestion(
         */
         beta = (C4_BETA_LOSS_1024 + MULT1024(c4_sensitivity_1024(c4_state), C4_BETA_LOSS_1024))/2;
     }
+    else if (c_mode == c4_congestion_ecn) {
+        /* Apply proportional reduction. Question: should it be sensitivity related? */
```
suhasHere: don't we think it needs to be sensitivity related ?

picoquic/c4.c
```diff
@@ -294,8 +347,6 @@ static void c4_apply_rate_and_cwin(
                 pacing_rate = c4_state->seed_rate;
             }
         }
-        /* Increase pacing rate by factor 1.25 to allow for bunching of packets */
-        pacing_rate = MULT1024(1024+256, pacing_rate);
```
suhasHere, line 298: why did we need to remove this ?

picoquic/c4.c
huitema: I will update the comment. The reduction is proportional to (rate - threshold)/threshold, and the threshold itself is sensitivity related. I don't think we need an additional sensitivity term.

suhasHere: understood . makes sense

picoquic/c4.c
huitema: The increase in pacing rate was triggering early ECN notifications. L4S tries to keep the delay very low. The recommended value in the ECN RFC is 250 microseconds. Any bunch larger than that will trigger marking. I think the 250us number is unrealistic, I would be curious to know what is deployed in practice, I suspect something between 1 and 5 ms.

There is a trade-off here. We need big bunches for performance, say at least 10 packets. But 10 packets mean 10*1500 bytes, 120,000 bits. At 120 Mbps, that's just 1ms, so fine. But at data rates lower than rate, the delays pile up quickly, 12ms at 10ms. To not be blocked, we need in general to limit the "quantum" parameter used in pacing. BBR limits it to 1ms of transmission. In the C4 code, I set the limit to 4ms, which is a bit more optimistic. We may need to tune that later, once we know what is deployed in practice, e.g., what Comcast does.

I was also trying to send bug batches during the initial phase so we can quickly get an estimate to the path data rate (a form of chirping), forcing that by increasing the pacing rate. Hence the 1.25 coefficient. But that leads to big packet batches at low data rate, thus instant queues that are punished by ECN. It turns out that this chirping is only useful in long delay links (e.g. geo satellites), and on those paths we can use "safe resume" instead. So we had an unproven feature, with known negative effects in presence of L4S or other queue limiting AQM, and a very narrow range of positive effects. Better get rid of it.

PR #1984, opened 2025-12-02
Fix a bug in the estimation of max rtt

huitema:
There was a coding mistake. Instead of updating the max RTT only if the previous transmission was at most at the nominal rate, it was updating it all the time. After fixing the bug, a bunch ofC4  tests started to fail. The fixed code was estimating max RTT too tightly, which leads to a "stable mistake": underestimated max RTT and rate means the bandwidth never grows. The fix is to add a margin, so the CWND is set to `rate*(max_rtt+margin)`. 
This needed tuning, because if the margin is too high it admits too many packets in the network and "real time" tests break. After tuning, this PR sets `margin = min(max_rtt/4, 15ms)`. A companion PR is the C4 depot will update the specs accordingly.

### DIFF
--- a/picoquic/c4.c
+++ b/picoquic/c4.c
@@ -96,6 +96,7 @@
 #define C4_NB_CRUISE_BEFORE_PUSH 4
 #define C4_MAX_DELAY_ERA_CONGESTIONS 4
 #define C4_RTT_MARGIN_5PERCENT 51
+#define C4_RTT_MARGIN_DELAY 15000
 #define C4_MAX_JITTER 250000
 #define C4_KAPPA ((double)(1.0/4.0))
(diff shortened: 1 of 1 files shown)


huitema:
For information, there were 4 C4 tests that needed updates:

* the "max of 100" completion times of the ECN test rose from 5s to 5.6s
* the "max of 100" completion times of the "bad wifi" test rose from 4.3s to 4.5s
*  the "max of 100" completion times of the "compete with cubic under bad wifi" test rose from 12.1s to 12.5s
* the max latency observed in 100 "media short then long RTT" test rose from 125 to 126ms.

The changes are probably due to the stricter assessment of Max RTT.

suhasHere (approved):

PR #1987, opened 2025-12-04
Create picowt_reset_stream() API

huitema:
Add `picowt_reset_stream()` API, documented in pico_webtransport.h.
Make sure that webtransport tests negotiate reset at, using 2 APIs:
~~~
    void picowt_set_transport_parameters(picoquic_cnx_t* cnx); 
    void picowt_set_default_transport_parameters(picoquic_quic_t* quic);
~~~
Add test of API, `picowt_baton_reset`.

Discovered a bug in implementation of `picoquic_reset_stream_at`. Fix it using a compromise: the reset_stream_at frame is only sent when the guaranteed bytes are acknowledged -- the alternative was just too messy on the sender side.

### DIFF
--- a/picohttp/h3zero_common.c
+++ b/picohttp/h3zero_common.c
@@ -96,11 +96,6 @@ static void picohttp_stream_node_delete(void * tree, picosplay_node_t * node)
 
 void h3zero_delete_stream(picoquic_cnx_t * cnx, h3zero_callback_ctx_t* ctx, h3zero_stream_ctx_t* stream_ctx)
 {
-#if 1
-	if (stream_ctx->stream_id == 36) {
-		DBG_PRINTF("%s", "bug");
-	}
-#endif
 	if (cnx != NULL) {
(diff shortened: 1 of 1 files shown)


picohttp/h3zero_common.c
```diff
@@ -96,6 +96,11 @@ static void picohttp_stream_node_delete(void * tree, picosplay_node_t * node)
 
 void h3zero_delete_stream(picoquic_cnx_t * cnx, h3zero_callback_ctx_t* ctx, h3zero_stream_ctx_t* stream_ctx)
 {
+#if 1
+	if (stream_ctx->stream_id == 36) {
+		DBG_PRINTF("%s", "bug");
+	}
```
suhasHere: curious why is stream_id 36 special

huitema: Debug info. This was caused by a parsing bug when processing a repeat attempt of a "reset stream at" frame, whose frame ID is 36. Buggy code was confusing that for the stream ID. I thought I had removed all that kind of debug code...

PR #1988, opened 2025-12-08
Avoid bouncing into congestion after recovery

huitema:
When analysing ECN traces, I found several instance of "bounces", which happens as follow:

* C4 goes into "pushing" mode (+25%)
* Queue builds up rapidly because sending rate is too large
* ECN/CE marks cause an early detection of congestion, C4 enters recovery mode
* after acknowledging the packets in flight during the pushing phase, C4 enters "cruising"
* the residual queue causes a rapid influx of ECN marks, thus congestion notification and drop in nominal rate
* that drop is too large, and it takes a long time to recover and retrieve the proper rate

The root cause here is that recovery only dropped to 15/16 of the "nominal rate", but that rate was a bit over estimated so the queue is not reduced. The solution lies in two changes:

* if a second congestion is notified during recovery, reset the `alpha_current' multiplier to 7/8 (instead of 15/16), and reset the "era_sequence" to the current send pointer to increase the duration in recovery.
* in general, detect end of era by testing the era sequence against the "lowest sequence number not acked" rather than the "highest number acked". This is more robust in case of out of order deliveries, or if there are holes in the acknowledgement sequence.

### DIFF
--- a/picoquic/c4.c
+++ b/picoquic/c4.c
@@ -83,6 +83,7 @@
 #define MULT1024(c, v) (((v)*(c)) >> 10)
 #define C4_ALPHA_NEUTRAL_1024 1024 /* 100% */
 #define C4_ALPHA_RECOVER_1024 960 /* 93.75% */
+#define C4_ALPHA_RECOVER2_1024 896 /* 87.50% */
 #define C4_ALPHA_CRUISE_1024 1024 /* 100% */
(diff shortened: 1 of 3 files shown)


huitema:
With this PR, the "worst of 100" time for the ECN test drops from 5.6 seconds to 5 seconds.

PR #1995, opened 2025-12-11
Fix missing padding in mTLS server handshake

karen07:
## Summary

This fix addresses an issue where the **ServerHello packet in mTLS mode was sent without the required padding**. The problem was discovered accidentally when I enabled mandatory mTLS and generated **ED25519** certificates for both the client and the server, signed by an **ED25519 CA**. The handshake failed to complete: the client never responded to ServerHello, and `tcpdump` showed repeated retransmissions.

Further investigation revealed that the **ServerHello packet size was below 1200 bytes**, which violates QUIC’s minimum datagram size requirement.

---

## Root Cause Analysis

While studying the library behavior, I found that padding is applied inconsistently depending on whether mTLS is used:

- **Without mTLS:** After sending ServerHello, the connection transitions to `picoquic_state_server_false_start`, where padding is applied.

- **With mTLS enabled:** The connection remains in `picoquic_packet_handshake`, and padding is added **only if fewer than 256 bytes remain** in the packet.

Because **ED25519 certificates and signatures are very small**, the handshake message left **more than 256 bytes unused**, and since there was nothing else to include, the packet was sent **without any padding**.

Existing tests did not catch this problem because they used **RSA certificates**, which are large enough that the handshake packet naturally exceeded 1200 bytes.

---

## What Was Done

I added test coverage for all combinations of client, server, and CA certificate types:

```
ecdsa
ed25519
rsa
rsa-pkcs8
secp256r1
secp256r1-pkcs8
secp384r1
secp521r1
```

Test combinations:

- **pass after the fix**.

---

## Notes

I have been working with the library for less than a week, so the change in `sender.c` may not be the ideal long-term design. However, it fully resolves the issue and is validated by the expanded test suite. 

It would be greatly appreciated if you could review the approach and suggest improvements to make this fix more robust or better aligned with the library’s architecture.

image

### DIFF
--- a/picoquictest/tls_api_test.c
+++ b/picoquictest/tls_api_test.c
@@ -5662,7 +5662,7 @@ int request_client_authentication_25519_test()
     }
 
     if (ret != 0) {
-        DBG_PRINTF("%s", "mTLS client-auth test failed RSA\n");
+        DBG_PRINTF("%s", "mTLS client-auth test failed ED25519\n");
     }
 
     return ret;


karen07:
How can I fix this?
All these files are certificates generated by two scripts (which I also included).
I can remove the certificates themselves and leave only the scripts.

The scripts generate all possible combinations of:

a client certificate,

a server certificate,

and their CA.

In total, this produces 512 such sets.
The number of successful certificates is 512 × 5,
and the number of unsuccessful ones is 512 × 6.

karen07:
I want to avoid future issues similar to the one I discovered, where everything works for one certificate but not for another.
That's why I generated all possible variants, although I admit that adding 5,000 files is a lot.

huitema:
I understand that you want to achieve certainty, but this leads to some kind of explosion of combinations. Let's think harder. The bug is "initial message too small". The keys and certificates are not part of the initial messages -- in QUIC, they are sent in the "handshake" message, encrypted in the handshake key. The UDP Datagram that you worry about is the first UDP packet sent by the server. It will often contain two packets: an initial packet with the "server hello", and a handshake packet that contains the first bytes of the server first flight. Usually, the  "server hello" is short, but the encrypted data in the first flight is large, and has to be carried by several packets.

The bug that you found happens if the first server flight is unusually short. The actual value does not matter much. The extra bytes are carried in handshake packets that are not subject to size checks. You only need to reproduce the "short server flight" case. The big items there are certificates and signature. I suppose that's what is changing with mTLS. If it does not need the certificates, then the flight will be much shorter.

You really only use one test case. Use mTLS in a scenario where the first server flight is short. You may perhaps reuse the private keys that are already in the distribution, maybe ./certs/ed25519/key.pem and ./certs/ed25519/pub.pem, for minimal impact.

By the way, how do you use mTLS itself?

karen07:
I only downloaded your library a week ago and built a very simple test client–server application with it. As for mTLS, I wanted to use it as a possible option for user authentication. And when I generated the certificates, I noticed this mode. So I don’t yet have a specific plan for using mTLS.

I didn’t quite understand what you meant by
“I suppose that's what is changing with mTLS. If it does not need the certificates, then the flight will be much shorter.”

I’ll try reusing ./certs/ed25519/key.pem and ./certs/ed25519/pub.pem, but I think I tried that once, and because they are signed by an RSA CA, the bug doesn’t reproduce.

karen07:
Is there any mandatory code formatting?

karen07:
I see what you mean. When I wrote the description, I mixed up ServerHello and Handshake a bit. Yes, I actually meant that things don’t align after the Handshake.

huitema:
Here is what I would do:

1- Reformat the test `request_client_authentication_test()` into a function `request_client_authentication_test_one()` with parameters: test_server_cert_file, test_server_key_file, test_server_cert_store_file.

2- Add a new set of functions for:
- `request_client_authentication_test()` -- use the same key, cert, store as current test.
- `request_client_authentication_25519_test()` -- use key and cert per 25519 values in certs/ed25519.
- if that's not sufficient to reproduce the failure, get a different, shorter store file for 25519, and add the file to the appropriate certs folder.
- add the request_client_authentication_25519_test to the test declaration files.

3- Pursue the fixes that you are proposing in sender.c

karen07:
No, it’s not possible to reproduce it using the certificates that are already in the folder; I need to create the server, client, and root certificates myself.

I’ll create a function that takes five arguments: the paths to the server, client, and root certificates.

huitema:
On the topic of padding: the current approach to padding in sender.c is a bit too complicated.

QUIC has the concept of "packet coalescing", i.e., carrying several packets in the same UDP datagram. Typical examples are "initial, handshake", "initial, 0RTT", "handshake, 1RTT", or even "initial, handshake, 1RTT". The only rule is that all coalesced packets except the last MUST be "long header" packets (e.g., initial, handshake, 0RTT), because the long header carries the packet payload length and thus allows separating successive packets.

The padding requirement is only for "datagrams", i.e., for the content of UDP packets. The 1200 minimum requirement applies if one of the coalesced packet is an Initial packet, and if that Initial packet requires an acknowledgement. For example, padding is not required if the Initial packet only contains Acknowledgement or padding. It is required if the packet contains a padding frame.

The current code applies padding inside the packet formatting itself. That leads to extra complexity. When formatting an Initial packet, the code must predict whether the application will only send the Initial packet (padding would be required) or whether there will be a following packet after that (e.g., handshake or 0RTT). Same when formatting an handshake packet: the code must know whether there was an Initial packet before the handshake packet, and whether there might be a coalesced packet after it, e.g., a 1RTT packet. It would be wrong to pad the handshake packet if the bytes could be used to carry a useful 1RTT packet. Same logic also when sending 0RTT packets.

Other stacks have developed a simpler approach. They just try to fit as many packets as possible in the UDP datagram, without padding -- or maybe only padding to prevent finger-printing, but not to meet the Initial rule. Then, when the UDP datagram is almost ready to send, check whether it contains an Initial packet that requires the 1200 bytes minimum. In that case, just pad the UDP packets with random bytes. The receiver will get these bytes, try to parse it as a QUIC packet, fail because this is just random, and ignore it, but the 1200 bytes requirement will be met.

I did not want to do that initially, because it seemed gross. And then I didn't want to change my ways, because I am old and stubborn! But I guess that I have to swallow hard and just do it. In fact we will need that if we want to support SCONE in picoquic (see: https://datatracker.ietf.org/doc/html/draft-ietf-scone-protocol, section ). Random padding would take care of most of the corner cases, including mTLS, leaving just one: if the last packet is a 1RTT packet, padding should be inside that.

I think I can push a PR with that random padding soon, and actually fix the issue. But we still need a good test of the mTLS cases.

karen07:
I’m working on it right now — I’m writing a test for the mTLS scenarios. I’ll send it once it’s ready.

karen07:
I pushed a version with the test and commented out my changes in sender.c. With the changes commented out, the tests fail; if I uncomment my changes, the tests start passing.

picoquic/sender.c
```diff
@@ -2333,6 +2333,8 @@ int picoquic_prepare_packet_server_init(picoquic_cnx_t* cnx, picoquic_path_t * p
         else if (packet->ptype == picoquic_packet_handshake && length + checksum_overhead < send_buffer_max &&
             (cnx->crypto_context[3].aead_encrypt == NULL || length + checksum_overhead + PICOQUIC_MIN_SEGMENT_SIZE > send_buffer_max)) {
             length = picoquic_pad_to_target_length(bytes, length, send_buffer_max - checksum_overhead);
+        } else if (packet->ptype == picoquic_packet_handshake && length + checksum_overhead < send_buffer_max && picoquic_tls_client_authentication_activated(cnx->quic)) {
```
huitema:
```suggestion
        }
        else if (packet->ptype == picoquic_packet_handshake &&
            length + checksum_overhead < send_buffer_max &&
            picoquic_tls_client_authentication_activated(cnx->quic)) {
```

picoquic/sender.c
```diff
@@ -2333,6 +2333,8 @@ int picoquic_prepare_packet_server_init(picoquic_cnx_t* cnx, picoquic_path_t * p
...
+            length = picoquic_pad_to_target_length(bytes, length, send_buffer_max - checksum_overhead);
         }
     }
```
huitema, line 2343: That's a good fix. I was worried about interaction with other scenarios when the handshake packet is followed by a 1rtt packet, but your test of `picoquic_tls_client_authentication_activated` fixes that. See the suggestion: use shorter lines, and make the order of operation obvious.

picoquictest/tls_api_test.c
```diff
@@ -5576,6 +5577,93 @@ int request_client_authentication_test()
...
+                                                 test_server_cert_file, test_server_key_file,
+                                                 test_ca_cert_store_file);
+
+    if (ret != 0) {
+        DBG_PRINTF("%s", "mTLS client-auth test failed RSA\n");
+    }
+
+    return ret;
+}
+
```
huitema, line 5624: Yes. This is very much what I had in mind.

picoquic/sender.c
karen07: Why isn’t clang-format used in the project? Or did I just not notice it?

karen07: Yes, this looks better. I just usually leave all of this to the formatter.

karen07:
I don’t understand what CodeQL is complaining about. I’ve already checked all the return paths and even added a #define, but it still flags it. The code is very similar to the rest of the codebase. Could you help me understand what it’s complaining about with this variable?

huitema:
You don't need to do anything about codeQL. I dismissed the alerts. 

This is a security check. There are often security issues when a program accepts user input without checking. In our case, the test accepts names of file that come from the "user" interface without checking for things like path traversal. Of course it doesn't -- this is test input! But CodeQL is not smart enough to find out, and probably the developers are rewarded by how many alerts they throw. So they like to cry wolf, and we need to dismiss their false alarms.

huitema:
@karen07 I think this is ready. Do you want me to merge now?

karen07:
yes

karen07:
Thanks)

PR #1996, opened 2025-12-13
Random padding

huitema:
Rely on UDP-datagram level code to do padding if the `initial_sent` condition is sent.

This makes the code simpler, removing many lines in which the encoding of one packet depends on whether there will be coalesced packets.

### DIFF
--- a/picoquic/sender.c
+++ b/picoquic/sender.c
@@ -1966,7 +1966,7 @@ int picoquic_prepare_packet_client_init(picoquic_cnx_t* cnx, picoquic_path_t * p
 
     /* If context is handshake, verify first that there is no need for retransmit or ack
      * on initial context */
-    int force_handshake_padding = 0;
+    int force_handshake_ping = 0;
 
     if (ret == 0) {
         if (epoch > picoquic_epoch_initial) {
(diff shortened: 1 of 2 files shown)


huitema:
The code required changing reference log files, because if padding is done at the datagram level the Initial or Handshake packets will be shorter, thus the log changes.

By default, the "padding" bytes after the Initial or Handshake frames are logged as a "dropped" packet. This can be pretty jarring, because a common case would be logged as an error. It also poses a test issue, because the random padding values changes at each execution of the program. Logging that would cause the logs to change at each execution, which breaks tests that compare log files. The solution is to not log the content of dropped packets (because it is largely random, and thus not very useful), and also not log their type in the "padding" case (because it is meaningless).

karen07:
I will definitely take a look, but I don’t yet understand the internals of the library deeply enough, so I may need more time.

karen07:
image
I just spun up a picoquic server and connected to it with curl — is it expected that both the ServerHello and the handshake are sent with padding?

karen07:
My mistake, I was looking at the wrong thing.

karen07:
Could you explain why you decided to fill the padding with random data rather than zeros?

huitema:
> Could you explain why you decided to fill the padding with random data rather than zeros?

So the padding does not stand out, looks like encrypted data. Also because that's what everybody else does...

karen07:
Aren’t you afraid that something like Cloudflare might drop such packets, thinking they’re strange or malformed?

huitema:
No, I don't think Cloudflare will block that. Several QUIC stacks do the same trick, including that used by Firefox, with no issue. Not sure about the stack used by Cloudflare itself, but I think they do too.

karen07:
approved)

PR #1998, opened 2025-12-15
Add pqbench for perf testing.

huitema:
Adding a new executable for performance testing: "pqbench". Examples:

-  Run the pqbench server on default port 8724 with qlog option:

~~~
pqbench -q logdir
~~~

- Start a client that will launch 10 parallel connection and execute the same `perf` scenario on each:

~~~
pqbench localhost:8724 10 "=b1:*100:397:1000;"
~~~

The server is designed to accept a batch of connections and exit when all the connections are complete. The client will launch the batch of connections and exit when they are complete. We can use perf measurement tools to construct flame graphs, etc.

In the example, the client launches 10 connections, and for each connection asks 100 times to download 1000 bytes. We could of course use different scenarios. The idea is to concentrate on scenarios in which the server handles a large number of flows, with a large number of connections. This is a very different profile from the "large batch downloads" used in the previous tests, and hopefully we will learn about different ways to speed up the execution.

### DIFF
--- a/picoquic/picoquic.h
+++ b/picoquic/picoquic.h
@@ -287,7 +287,19 @@ typedef struct st_picoquic_connection_id_t {
 
 
 /* forward definition to avoid full dependency on picotls.h */
-typedef struct st_ptls_iovec_t ptls_iovec_t; 
+typedef struct st_ptls_iovec_t ptls_iovec_t;
+
+/* Alternate structure when applications need to access 
+ * content of an iovec without developing a dependency on picotls.h.
+ * They can use something like:
+ *     ....... ptls_iovec_t * list ...
+ *     picoquic_iovec_t * my_list = (picoquic_iovec_t *) list;
+ */
+
+typedef struct st_picoquic_iovec_t {
+    uint8_t* base;
+    size_t len;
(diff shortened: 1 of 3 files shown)


picoquic/quicctx.c
```diff
@@ -846,7 +925,7 @@ void picoquic_set_default_multipath_option(picoquic_quic_t* quic, int multipath_
     quic->default_multipath_option = multipath_option;
 
     if (multipath_option & 1) {
-        quic->default_tp.is_multipath_enabled = 1;
+        /* quic->default_tp.is_multipath_enabled = 1;*/
```
a-andre, line 928: Why has this line been removed?

huitema: The previous version of the multipath draft had two separate Transport Parameters: enable multipath, and maximum number of paths. The latest version has just one: maximum path ID, which has to be set to at least 1 to negotiate the multipath option. The transport parameter "is_multipath_enabled" is not used.

I did not remove it entirely because the transport parameter tests have a list of values of the transport parameter structure, and removing one member in the structure requires editing that list. It should be done asap...

PR #2001, opened 2025-12-18
Try use poll instead of select

huitema:
In theory, poll() gives better performance than select and is available on all Unix systems. Testing that.

### DIFF
--- a/picoquic/sockloop.c
+++ b/picoquic/sockloop.c
@@ -618,7 +618,7 @@ int picoquic_packet_loop_poll(
     /* Picoquic expresses times in microseconds, but the timeout 
      * parameter of poll() is in milliseconds. The code below converts
      * the microsecond delay to the nearest millisecond, which is a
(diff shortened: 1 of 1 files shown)


picoquic/sockloop.c
```diff
@@ -572,18 +577,119 @@ int picoquic_packet_loop_wait(picoquic_socket_ctx_t* s_ctx,
...
+    struct sockaddr_storage* addr_dest,
+    int* dest_if,
+    unsigned char* received_ecn,
+    uint8_t* buffer, int buffer_max,
+    int64_t delta_t,
+    int* is_wake_up_event,
+    picoquic_network_thread_ctx_t* thread_ctx,
+    int* socket_rank)
+{
+    int delta_t_ms = (int)((delta_t + 500) / 1000);
```
suhasHere, line 627: may be a comment will be helpful here

PR #2006, opened 2025-12-22
Bug fix from picoquic_frame_type_datagram_l to picoquic_frame_type_datagram

karen07:
Fix DATAGRAM payload offset when squeezing DATAGRAM_LEN When converting a DATAGRAM_LEN frame into a DATAGRAM frame to fit MTU, the payload was copied from an incorrect offset. The code skipped only the length varint but not the frame type byte, causing part of the length field to be included in the payload and corrupting the datagram. This change fixes the offset by copying from frame_content + 1 + header_length, ensuring only the actual payload is transmitted.

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -5303,8 +5303,8 @@ uint8_t * picoquic_format_first_datagram_frame(picoquic_cnx_t* cnx, uint8_t* byt
 
         if (frame_content[0] == picoquic_frame_type_datagram_l) {
             /* It might be possible to squeeze the size by removing the length */
-            size_t header_length = picoquic_varint_skip(frame_content + 1);
-            size_t data_length = cnx->first_datagram->length - header_length - 1;
+            size_t header_length = 1 + picoquic_varint_skip(frame_content + 1);
+            size_t data_length = cnx->first_datagram->length - header_length;
             size_t min_length = data_length + 1;
 
             if (bytes + min_length <= bytes_max) {


picoquic/frames.c
```diff
@@ -5312,7 +5312,7 @@ uint8_t * picoquic_format_first_datagram_frame(picoquic_cnx_t* cnx, uint8_t* byt
                     *bytes++ = picoquic_frame_type_padding;
                 }
                 *bytes++ = picoquic_frame_type_datagram;
-                memcpy(bytes, frame_content + header_length, data_length);
+                memcpy(bytes, frame_content + 1 + header_length, data_length);
```
huitema: That works, but it is a bit cryptic. I think the bug is in the computation of "header_length" on line 5306. The "header" is logically defined as "length of type + length of type parameter", but the computation on lines 5306-5307 omits "length of type". These two lines should be changed to:

~~~
            size_t header_length = 1 + picoquic_varint_skip(frame_content + 1);
            size_t data_length = cnx->first_datagram->length - header_length;
~~~

If we do that, we can leave line 5315 unchanged. We also make the code a bit easier to understand.

karen07: Yes, you’re right. Should I rebase and push an updated commit, or would you prefer to handle it on your side?

Regarding the test: I found this issue while running a high datagram load, but I agree that it should be covered by a proper unit test.

huitema: I am busy with the family and the grandchildren today and tomorrow, traveling Friday. So feel free to do it!

karen07:
Fixed)

huitema (approved):
Thanks!

PR #2007, opened 2026-01-05
Revise the rate measurement and the Initial phase of C4

huitema:
Revise the rate measurement of C4 to use the value computed by picoquic instead of doing the computation anew in less precise way. The tighter rate measurement prevents some amount of "RTT creeping up" seen in the previous version.

Also revise the initial phase to use a "CWND based" logic, similar to Reno, instead of the "rate based" logic similar to BBR.

The reasons for these changes are explained in papers added in a [parallel PR in the C4 project](https://github.com/private-octopus/c4/pull/88)

### DIFF
--- a/picoquic/c4.c
+++ b/picoquic/c4.c
@@ -576,6 +576,17 @@ static void c4_initial_handle_loss(picoquic_path_t* path_x, c4_state_t* c4_state
 static void c4_initial_handle_ack(picoquic_path_t* path_x, c4_state_t* c4_state, picoquic_per_ack_state_t* ack_state, uint64_t current_time)
 {
     c4_state->nb_packets_in_startup += 1;
+    /* We implement Reno style slow start, doubling the CWND every RTT, by
+    * incrementing CWND by the number of new bytes acknowledged.
+    * However, this is too aggressive at the end of the initial phase.
+    * The exit test is "3 successive RTT with no data rate increase".
+    * If the CWND keeps increasing after noticing the first increase,
+    * it grows too much and the queues build up too much. But then, the first
+    * notice can also be due to a jitter event, which causes rate measurment to be low,
+    * in which case we need  to increase the CWND to eventually catch with a new RTT.
+    * The 'shift' in the formula is a compromise, kinda similar to "Hystart++",
+    * causing the CWND to increase slower if we may be close to the exit.
+    */
     c4_state->initial_cwnd += ack_state->nb_bytes_acknowledged >> (3 * c4_state->nb_eras_no_increase);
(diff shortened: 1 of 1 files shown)


picoquic/c4.c
```diff
@@ -796,9 +820,10 @@ void c4_handle_ack(picoquic_path_t* path_x, c4_state_t* c4_state, picoquic_per_a
         if (c4_era_check(path_x, c4_state)) {
             /* Update max rtt and running min rtt */
             c4_update_min_max_rtt(path_x, c4_state);
-            /* test need to reenter initial if conditions did change */
+            /* test needed to reenter initial if conditions did change */
             if (!c4_state->initial_after_jitter &&
                 c4_state->nominal_max_rtt > 50000 &&
+                c4_state->nominal_rate < 1000000 &&
```
suhasHere, line 845: just to understand, this clamps it under 1 gig connections right

picoquic/c4.c
```diff
@@ -542,6 +576,7 @@ static void c4_initial_handle_loss(picoquic_path_t* path_x, c4_state_t* c4_state
 static void c4_initial_handle_ack(picoquic_path_t* path_x, c4_state_t* c4_state, picoquic_per_ack_state_t* ack_state, uint64_t current_time)
 {
     c4_state->nb_packets_in_startup += 1;
+    c4_state->initial_cwnd += ack_state->nb_bytes_acknowledged >> (3 * c4_state->nb_eras_no_increase);
```
suhasHere, line 590: can you please add some text to epxain the intuition here

picoquic/c4.c
huitema: The reasoning is that in some conditions, the initial phase will exit too early -- especially if we have both high jitter and competition from other flows. Finding an RTT higher than the previous max, is indication that the previous initial might have exited too soon. This could be due to high jitter that caused rate estimate to be too low, because estimate is basically CWND/RTT, and jitter causes RTT to be higher than expected. Competition can cause congestion and packet losses, which may also cause early exit. Reentering Initial is a bit of a hack: it will cause a packet burst, probably some congestion and packet loss. That's OK if the issue is actually high jitter and competition: the congestion will force the competing flow to back up a little. But that's detrimental if the connection was going fine. So we don't do that if the RTT is low (lower than 50ms) or if the data rate is high enough (higher than 1Mbps, i.e., 8Mbps). And we only do that once per connection.

picoquic/c4.c
huitema: Done.

picoquic/c4.c
huitema: Added comment in the code.

PR #2012, opened 2026-01-18
Add SETTINGS_ENABLE_WEBTRANSPORT for Chrome compatibility

mondain:
Chrome's WebTransport implementation requires SETTINGS_ENABLE_WEBTRANSPORT (0x2b603742) from an older draft, in addition to the current draft's SETTINGS_WT_MAX_SESSIONS (0x14e9cd29).

Changes:
- Add h3zero_settings_enable_webtransport constant (0x2b603742)
- Include in default SETTINGS frame (length increased from 22 to 27 bytes)
- Encode when webtransport_max_sessions > 0
- Decode and set webtransport_max_sessions = 1 if received

Without this, Chrome returns ERR_METHOD_NOT_SUPPORTED when attempting WebTransport CONNECT requests.

### DIFF
--- a/picohttp/h3zero.c
+++ b/picohttp/h3zero.c
@@ -1106,23 +1106,29 @@ void h3zero_delete_data_stream_state(h3zero_data_stream_state_t * stream_state)
 static uint8_t const h3zero_default_setting_frame_val[] = {
     0, /* Control Stream ID, varint = 0 */
     (uint8_t)h3zero_frame_settings, /* var int frame type ( < 64) */
-    22, /* Length of setting frame content */
+    27, /* Length of setting frame content */
     (uint8_t)h3zero_setting_header_table_size, 0, /* var int type ( < 64), then var int value (0) */
     (uint8_t)h3zero_qpack_blocked_streams, 0, /* var int type ( < 64),  then var int value (0) Control*/
     /* enable_connect_protocol = 0x8 */
     (uint8_t)h3zero_settings_enable_connect_protocol, 1,
     /* datagram support */
     (uint8_t)h3zero_setting_h3_datagram, 1,
-    /* Declare max 1 web transport session */
+    /* Declare max 1 web transport session (draft-14: SETTINGS_WT_MAX_SESSIONS) */
     ((h3zero_settings_webtransport_max_sessions >> 24)&0xff)|0x80,
     (uint8_t)((h3zero_settings_webtransport_max_sessions >> 16)&0xff),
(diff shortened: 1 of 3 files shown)


huitema (commented):
This is cool, thanks. Two questions, though: where is the settings `h3zero_settings_enable_webtransport 0x2b603742` defined? Adding a draft name and  number to the comments would be nice. Also, what are the limit in interoperability? Are there some other messages or functions that need to be modified, or avoided?

mondain:
The setting `SETTINGS_ENABLE_WEBTRANSPORT (0x2b603742)` is in several of the draft-ietf-webtrans-http3 drafts, up-to 06 (section 8.2) as far as I'm aware. 

```
The SETTINGS_ENABLE_WEBTRANSPORT parameter indicates that the specified HTTP/3 connection
is WebTransport-capable.

   Setting Name:  ENABLE_WEBTRANSPORT
   Value:  0x2b603742
   Default:  0
```

I use the latest Chrome on Linux `Version 144.0.7559.59 (Official Build) (64-bit)` and without this option, WT does not work; unless one adds flags to startup, but no regular user, is going to change their startup IMHO.

This comment is for moqt draft-14 `Declare max 1 web transport session (draft-14: SETTINGS_WT_MAX_SESSIONS)`

I don't have answers for your other questions at this time, but I'll look.

huitema:
OK. I did a little analysis of the protocol changes through the web transport drafts. I get this:

1. Unify "enable" and "max sessions" transport parameter: draft 7.

2. Add a "Drain" capsule type: draft 7

3. Prioritization handling: draft 8

4. Sub-protocol negotiation: draft 9, becomes ALPN negotiation in draft 12

5. Specify key exporters per session: draft 10

6. Specify flow control transport parameter types, capsule types, error type:draft 12

Your PR addresses the first change. It does not address the DRAIN capsule, which means a picoquic implementation may end up sending a DRAIN capsule to an unsuspecting Chrome implementation. On the other hand, DRAIN is just a different way of closing the session. If Chrome closes the session with a "protocol error", so be it.

I am not concerned with prioritization handling, but we do have implementations that rely on ALPN selection to pick the actual protocol running on top of web transport. I suppose that if Chrome does not send the parameter, picoquic will end up picking a default. That's probably OK, but not ideal. It means that for now, servers should use the URL field to specify the protocol.

We do not expose key exporters yet in picoquic, and we only support one session per connection and thus never negotiate use of flow control, so that's not a problem.

PR #2014, opened 2026-01-20
Performance work for C4 on loopback

huitema:
To assess C4 performance, we ran tests on the loopback interface on Windows, repeating [a previous effort done for BBR](https://www.privateoctopus.com/2024/10/13/RandomLoopbackDelaysSlowBbr.html). The results were not impressive. The best trials show a throughput of about 1Gbps, more than twice less the results for BBR or Cubic. Due to a bug in the initial estimation, many results were even worse, with throughput of maybe 100 kbps.

We fixed the initial bug, making sure that C4 only exited the "initial" phase after getting a proper estimate of the bandwidth. We also improved the support for "memory log" in `pqbench`, and collected detailed statistics. The statistics showed that the RTT was erratic, between a few microseconds and 1 millisecond, and that the estimate of the "max RTT" was closely following the RTT, because the filtering period was reduced with the low RTT. As a result, the CWND value was often much too small, and was slowing down the transfer. The first fix was to set a floor of 1 millisecond for the max RTT. This did improve the results, with the measured throughput reaching 2.2 to 2.7 Gbps: about 88.7% of Cubic, and better than BBR.

Looking at the traces, we inferred that the gap with BBR was probably due to inefficient pacing. In "cruising" periods, C4 attempts to pace the data at the "nominal rate". This forces periodic wait for new rate control credits. When the CPU is nearly overloaded, the system will take a bit longer than expected to return from the "wait". At high data rate, the "pacing quantum" is not sufficient to smooth over these extra delays, and the connection is slowed down. Our solution is to loosen pacing "a bit", increasing the pacing rate by 3/64th in "cruising" state when the RTT is less than 1ms.

image

This worked! As seen on the graph above, the C4 trials deliver rates of a bit more than 3Gbps, which is better than Cubic.


picoquic/c4.c
```diff
@@ -719,7 +726,16 @@ static void c4_enter_cruise(
...
-    c4_state->alpha_1024_current = C4_ALPHA_CRUISE_1024;
+    if (path_x->smoothed_rtt < C4_MAX_RTT_MIN) {
+        /* When operating in a CPU limited environment, pacing is too
+         * conservative, and should be loosened. Ideally, we would detect
+         * the CPU limited condition by comparing pacing rate and the actual
+         * send rate, but that requires some amount of testing. In practice,
+         * we have only tested this loosening in loopback tests, so we only
+         * apply it if the RTT is below 1ms.
+         */
+        c4_state->alpha_1024_current = C4_ALPHA_CRUISE_1024 + 48;
```
suhasHere, line 737: I wonder if this needs to be a transport config ?

huitema: For now, it is a very narrow applicability. There are not that many paths with an RTT below 1ms. And delay measurements really are bad when the delay is that low: system call delay very easily vary between 10 and 100 microseconds. But yes, we will need to revisit this, as the comment says.

PR #2018, opened 2026-01-27
Fix flow control attack

### LINKED ISSUE #2015: Flow Control Bypass via Offset Manipulation
**Location:** sender.c

```C
if (ret == 0){
    if (cnx->quic->max_data_limit != 0) {
        if (cnx->data_received + ((3 * cnx->quic->max_data_limit) / 4) > cnx->maxdata_local) {
            uint64_t max_data_increase = cnx->data_received + cnx->quic->max_data_limit - cnx->maxdata_local;
            bytes_next = picoquic_format_max_data_frame(cnx, bytes_next, bytes_max, &more_data, &is_pure_ack,
                max_data_increase);
        }
    }
```

**Issue:** The `max_data` update logic uses `data_received` which tracks the highest received offset across all streams, not actual consumed bytes.

**Attack:** An attacker can send a single 1-byte STREAM frame with a large offset (e.g., offset=1KB, length=1), causing `data_received` to increase by 1KB while only consuming 1 byte of actual data. This triggers `max_data` to increase proportionally, giving the attacker a much larger send window at minimal cost.

**Impact:** 
- Flow control bypass
- Application-layer DoS (streams blocked waiting for missing data)
- Connection resource exhaustion

**Suggested Fix:** Base `max_data` updates on total `consumed_offset` across all streams, not `data_received`.

huitema:
Close #2015 

Verified that the issue #2015 is valid. When flow control is engaged, base increase on bytes delivered to the application rather than maximum offset. This is a bit slower in presence of packet losses, but less vulnerable to abuse.


huitema:
The satellite tests show a big dip in performance in case of packet losses. This is expected: the flow control window only expands if bytes are delivered. In the simulation of 1% packet loss, we are constantly waiting for a packet correction, so the time to delivery is twice the RTT, and the flow control window is too small. We just changed the expected result, but another possibility would be to increase the flow control window.

doublex:
Could you give a recommendation on how much to increase the flow control window?

huitema:
How much to increase the flow control window? Well, it depends on the refresh algorithm used by the peer, and on the expected packet loss rate. In the absence of packet loss, we want to ensure that the peer always has credits available.

A popular algorithm is to refresh the window by sending a MAX DATA frame when a fraction `alpha` of the window has been received. In picoquic, `alpha` is set to 1/4th, and the code is:

~~~
if (cnx->data_received + ((3 * cnx->quic->max_data_limit) / 4) > cnx->maxdata_local) {
    uint64_t max_data_increase = cnx->data_received + cnx->quic->max_data_limit - cnx->maxdata_local;
    bytes_next = picoquic_format_max_data_frame(cnx, bytes_next, bytes_max, &more_data, &is_pure_ack,
                                max_data_increase);
}
~~~

We can see the following time line:

1- sender sends the packet that brings the "consumed" part over the limit
2- receiver gets that packet after `RTT/2`
3- sender receives the new `MAX_DATA` after another `RTT/2`

At that point the server has consumed the old value, the fraction `alpha` of the maximum flow control window, and is consuming another fraction `1-alpha` in the RTT between the triggering packet and the arrival of `MAX_DATA`. If we want to guarantee a rate R, we must ensure: `(1-alpha)*increase > RTT*rate`. For picoquic, since alpha is 1/4, that means the `increase` should be at least `BDP*4/3`. But you are not guaranteed to interact with picoquic. Other implementations have different `alpha`, maybe as large as `1/2`, so a reasonable value would be, set the increase to `2*BDP`.

If we have packet losses, as in our satellite test, the equation changes. The value `cnx->data_received` only increases when the packets up to `cnx->data_received` have been received in order. It takes at least `1 RTT` for losses to be corrected. If we want to enable continuous transmission in the presence of loss, the increase should probably be at least `3*BDP`.

This can get worse if we consider second losses, the loss of the packet that corrects the previous loss. For a loss rate of 1%, the probability of a second loss is 0.01%. This seems small, but in the case of a high speed satellite link, we can easily have more than 10,000 packets in flight. That means more than a 100 could be lost per RTT, and more than 1 per RTT will be lost twice. If we want to sustain transmission in presence of two losses, the increase should probably be at least `4*BDP`.

Of course, all that is a compromise. Small devices may not have enough memory to allocate `4*BDP` to the reordering queue. They may decide to use a smaller window, even if that causes an occasional slow down.

huitema:
The latest commit adds comment in `picoquic.h` to provide guidance on how to set the `max_data` parameter.

doublex:
Thank you for the detailed explanation. It is really helpful.
You really know your stuff.
It is a pleasure to work with you.
Greetings from Vienna

PR #2019, opened 2026-01-27
Add flow control initial max data parameter to configuration options.

hfstco:
For my experiments, it’s useful to increase the initial flow control max data parameter. Right now, I achieve this by modifying picoquicdemo in the source code.

It would be helpful if flow control could be disabled (or configured) via a command-line parameter instead. Some other QUIC implementations (e.g. quiche, mvfst) also provide an option for that.

I used the '-Z' flag for this option, since there weren’t many unused letters left. 

Currently, only the global max_data parameter can be set. We can extend this to max_stream_data parameters too. 

Flow control initial max default:
Screenshot 2026-01-27 at 22 27 32

Flow control initial max set to 10 MB (-Z 10000000):
Screenshot 2026-01-27 at 22 28 28

### DIFF
--- a/picoquictest/config_test.c
+++ b/picoquictest/config_test.c
@@ -317,8 +317,6 @@ static config_error_test_t config_errors[] = {
     { 2, { "-d", "idle" }},
     { 1, { "-Z" }},
     { 2, { "-Z", "0123456789abcdexyedcba9876543210" }},
-    /* INT_MAX + 1. config_atoi() currently only supports 32-bit integers. */
-    { 2, { "-Z", "2147483648"}}
 #ifdef PICOQUIC_WITHOUT_SSLKEYLOG
     { 1, {"-8"}},
 #endif


picoquic/config.c
```diff
@@ -958,6 +971,7 @@ void picoquic_config_init(picoquic_quic_config_t* config)
     config->initial_random = 3;
     config->cwin_max = UINT64_MAX;
     config->idle_timeout = PICOQUIC_MICROSEC_HANDSHAKE_MAX / 1000;
+    config->flow_control_max = PICOQUIC_INITIAL_FLOW_CONTROL_MAX;
```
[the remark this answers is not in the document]
hfstco: Fixed.

picoquic/config.c
```diff
@@ -469,7 +470,17 @@ static int config_set_option(option_table_line_t* option_desc, option_param_t* p
...
-
+    case picoquic_option_FLOW_CONTROL_MAX: {
+        int v = config_atoi(params, nb_params, 0, &ret);
+        if (ret != 0 || v < 0) {
+            fprintf(stderr, "Invalid flow control max option: %s\n", config_optval_param_string(opval_buffer, 256, params, nb_params, 0));
+            ret = (ret == 0) ? -1 : ret;
+        }
+        else {
+            config->flow_control_max = (v==0)?UINT64_MAX:v;
+        }
```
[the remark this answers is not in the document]
hfstco: Fixed. Set flow_control_max to `PICOQUIC_INITIAL_FLOW_CONTROL_MAX` in case of an int overflow.

picoquic/picoquic_config.h
```diff
@@ -78,6 +78,7 @@ typedef enum {
     picoquic_option_ECH_server,
     picoquic_option_ECH_client,
     picoquic_option_ECH_init,
+    picoquic_option_FLOW_CONTROL_MAX,
     picoquic_option_HELP
```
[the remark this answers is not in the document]
hfstco: I don't think anyone uses the options directly with numeric values.

picoquictest/config_test.c
```diff
@@ -437,6 +443,7 @@ int config_test_compare(const picoquic_quic_config_t* expected, const picoquic_q
     ret |= config_test_compare_string("ech_key_file", expected->ech_key_file, actual->ech_key_file);
     ret |= config_test_compare_string("ech_config_file", expected->ech_config_file, actual->ech_config_file);
     ret |= config_test_compare_string("ech_public_name", expected->ech_public_name, actual->ech_public_name);
+    ret |= config_test_compare_int("flow_control_max", expected->flow_control_max, actual->flow_control_max);
```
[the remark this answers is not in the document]
hfstco: Right now, the test suite doesn’t include a `uint64_t` comparison method. We could consider adding one. At the moment, `picoquic_option_CWIN_MAX` also relies on the int comparison method, which means setting these options could fail if the configured value exceeds `INT_MAX`.

picoquic/config.c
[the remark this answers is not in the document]
hfstco: But values larger than `INT32_MAX` will end in setting flow_control_max to `PICOQUIC_INITIAL_FLOW_CONTROL_MAX`.

hfstco:
@huitema Can you take a look at the appveyor error? I don't have a Windows machine available right now. On macos everything compiles fine.

Edit: AppVeyor failed because a uint64_t value was being compared using the int32 comparison method. This should be fixed now.

picoquictest/config_test.c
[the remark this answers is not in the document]
hfstco: Ok, we have a uint64 comparison method in the test suite! I didn't know that. Fixed.

picoquic/config.c
```diff
@@ -35,6 +35,7 @@
 #include "tls_api.h"
 #include "picoquic_config.h"
 #include "picoquic_bbr.h"
+#include "picoquic_internal.h"
```
huitema, line 38: Why did you need to include the internal header?

picoquic/picoquic_internal.h
```diff
@@ -50,6 +50,7 @@ extern "C" {
 #define PICOQUIC_NB_PATH_DEFAULT 2
 #define PICOQUIC_MAX_PACKETS_IN_POOL 0x2000
 #define PICOQUIC_STORED_IP_MAX 16
+#define PICOQUIC_INITIAL_FLOW_CONTROL_MAX 0x100000
```
huitema, line 54: It makes sense to use a named constant instead of "0x100000" in quicctx.c, but I would not make that visible in the configuration program.

picoquic/config.c
```diff
@@ -958,6 +972,7 @@ void picoquic_config_init(picoquic_quic_config_t* config)
     config->initial_random = 3;
     config->cwin_max = UINT64_MAX;
     config->idle_timeout = PICOQUIC_MICROSEC_HANDSHAKE_MAX / 1000;
+    config->flow_control_max = PICOQUIC_INITIAL_FLOW_CONTROL_MAX;
```
huitema: Please leave that to 0 by default!

picoquic/config.c
```diff
@@ -469,7 +471,17 @@ static int config_set_option(option_table_line_t* option_desc, option_param_t* p
...
     }
-
+    case picoquic_option_FLOW_CONTROL_MAX: {
+        int v = config_atoi(params, nb_params, 0, &ret);
+        if (ret != 0 || v < 0) {
+            fprintf(stderr, "Invalid flow control max option: %s\n", config_optval_param_string(opval_buffer, 256, params, nb_params, 0));
+            ret = (ret == 0) ? -1 : ret;
+        }
+        else {
+            config->flow_control_max = (v==0)?PICOQUIC_INITIAL_FLOW_CONTROL_MAX:v;
```
huitema: Why do you need to transform "config->flow_control_max = 0" to a large constant? It would be much simpler to keep the existing meaning, in which 0 means "act as if it is not configured". 

The max value supported by `config_atoi` is INT_MAX, i.e., 0x7FFFFFF, 2 Gigabytes. I think that's large enough for the current scenarios. If we want more, we will indeed need to define `config_atoi64`, and a 64 bit comparison test in the test programs. We will do that if there is a use case.

I did not see the part where you are calling `picoquic_set_max_data_control`? Did I miss something?
I would have expected to see code added to the function `picoquic_create_and_configure` in `config.c` to configure the flow control max if the option is set. Something like:
~~~
    if (config->flow_control_max > 0){
        picoquic_set_max_data_control(quic, config->flow_control_max);
    }
~~~

huitema (commented):
Sorry, but I think we need a bit more work.

First, see the comments. I would tread "-Z 0" just the same as "not setting -Z in the first place". That obviates the need to use default values: just leave "0" to mean "not set".
You also need to apply the parameter in the function `picoquic_create_and_configure`, but of course only if set to a non zero value.

Second, I see that you updated the config tests to check the parameter value. That's great. After adding the setup code in  `picoquic_create_and_configure`, we may want to check that the option is actually applied. Maybe add some code in `config_quic_test_one` in `config_test.c` to check that the value of `quic->max_data_limit` and `quic->default_tp.initial_max_data` match the value set in the "-Z" param? That will ensure that we have coverage of the configuration.

picoquic/config.c
hfstco: > Why do you need to transform "config->flow_control_max = 0" to a large constant? It would be much simpler to keep the existing meaning, in which 0 means "act as if it is not configured".

>The max value supported by config_atoi is INT_MAX, i.e., 0x7FFFFFF, 2 Gigabytes. I think that's large enough for the current scenarios. If we want more, we will indeed need to define config_atoi64, and a 64 bit comparison test in the test programs. We will do that if there is a use case.

Yes, that fine.

>I did not see the part where you are calling picoquic_set_max_data_control? Did I miss something?
I would have expected to see code added to the function picoquic_create_and_configure in config.c to configure the flow control max if the option is set. Something like:

Line number 954 in `config.c`.

picoquic/config.c
hfstco: I have changed that.

picoquic/picoquic_internal.h
hfstco: Removed `PICOQUIC_INITIAL_FLOW_CONTROL_MAX` in config.c.

picoquic/config.c
hfstco: See above. Copilot suggested it. But that's no longer an issue, as we don't use `PICOQUIC_INITIAL_FLOW_CONTROL_MAX` in `config.c`.

hfstco:
>First, see the comments. I would tread "-Z 0" just the same as "not setting -Z in the first place". That obviates the need to use default values: just leave "0" to mean "not set".
You also need to apply the parameter in the function picoquic_create_and_configure, but of course only if set to a non zero value.

I have changed that.

>Second, I see that you updated the config tests to check the parameter value. That's great. After adding the setup code in picoquic_create_and_configure, we may want to check that the option is actually applied. Maybe add some code in config_quic_test_one in config_test.c to check that the value of quic->max_data_limit and quic->default_tp.initial_max_data match the value set in the "-Z" param? That will ensure that we have coverage of the configuration.

Added two separate test cases for -Z > 0 and -Z == 0. I also extended `config_quic_test_one` to verify that `quic->max_data_limit` and `quic->default_tp.initial_max_data` are set correctly. Additionally, I added some error cases for: -Z without a parameter, -Z with a non-numeric parameter and -Z with a value > `INT_MAX`

Edit: Ok, I removed the -Z > `INT_MAX` error case, because otherwise we have a memory leak.

huitema (approved):
Sorry for missing the line in config.c in the first review. Looks really good now! Thanks for doing that work.

Issue #1829, opened 2025-02-03
Duplicate code between logconvert.c and logger.c


huitema:
The functions that produce names for packet types and frame types are duplicated in `logconvert.c` and `logger.c` (`ptype2str` vs. `textlog_ptype_name`, `ftype2str` vs `textlog_frame_names`). Removing the duplication will remove 187 lines of code, improve code coverage, and make creation of new frame types simpler.

The complicating factor is that `logger.c` is in the `picoquic-core` library, while `logconvert.c` is in `bin_log`. We should consider moving the file `log_convert` to the  `picoquic-core` library, while making sure that it is described in its own header and not loaded by default if not used.

huitema:
Resolved in PR 2021, defining common functions and adding a way to do qlog without conversions.

PR #2021, opened 2026-02-02
Rewrite qlog generation for streaming and common code

### LINKED ISSUE #1252: QLOG Streaming and new version
The QLOG definition has eveolved. Picoquic needs to support the V2, which may mean some minor changes in format. Main change is the support for streaming, which would be very nice and would allow direct support of qlog without an intermediary binary step.

huitema:
Close #1252 
Start addressing #1829

The goal is to enable direct generation of qlog, without adding the intermediate step of writing a binary log.

Before updating qlog.c, we want to make sure that we use as much common code as possible (see #1829).

Then, we test that the new qlog writing functions provide exactly the same output as the old one. This forces us to do a few natural things in order to be "bug for bug compatible".

After verifying that we can reach bug for bug compatibility, we rewrite the "autoqlog" definition to point to the new code, and we update the QLOG tests accordingly, removing the tests that were generating a binlog file before converting it.

After updating the trace tests, we replace all the references to "picoquic_get_binlog" in tests by "picoquic_get_qlog". We had to write a helper function to retrieve the "max byte in flight" from the qlog instead of the binlog in some congestion and flow control tests.


loglib/qlog_frames.c
```diff
@@ -0,0 +1,647 @@
...
+    uint64_t recon_bytes_in_flight;
+    uint64_t recon_min_rtt;
+    uint64_t saved_ip_length;
+    const uint8_t* saved_ip;
+
+    if ((bytes = picoquic_frames_varint_decode(bytes, bytes_max, &lifetime)) == NULL ||
+        (bytes = picoquic_frames_varint_decode(bytes, bytes_max, &recon_bytes_in_flight)) == NULL ||
+        (bytes = picoquic_frames_varint_decode(bytes, bytes_max, &recon_min_rtt)) == NULL ||
+        (bytes = picoquic_frames_varint_decode(bytes, bytes_max, &saved_ip_length)) == NULL ||
+        (saved_ip_length != 4 && saved_ip_length != 4) ||
```
hfstco, line 454: `(saved_ip_length != 4 && saved_ip_length != 4)`

loglib/picoquic_qlog.h
```diff
@@ -0,0 +1,37 @@
...
+* DISCLAIMED. IN NO EVENT SHALL Private Octopus, Inc. BE LIABLE FOR ANY
+* DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+* (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+* LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+* ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+* (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+* SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+#ifndef PICOQUIC_QLOG_FNS_H
```
hfstco, line 22: Same identifier as in `picoquic_qlog_fns.h`.

loglib/qlog_fns.c
```diff
@@ -0,0 +1,1049 @@
...
+    unsigned int key_phase_sent : 1;
+    unsigned int key_phase_received_last : 1;
+    unsigned int key_phase_received : 1;
+    unsigned int spin_bit_sent_last : 1;
+    unsigned int spin_bit_sent : 1;
+
+    int state;
+} qlog_fns_context_t;
+
+#define QLOG_DECLARE_CONTEXT(ctx, cnx) qlog_fns_context_t * ctx = (qlog_fns_context_t*)cnx->qlog_ctx
```
hfstco, line 78: Not used?

loglib/qlog_fns.c
```diff
@@ -0,0 +1,1049 @@
...
+
+    qlog_fns_event_start(ctx, NULL, 0, picoquic_get_quic_time(cnx->quic), "info", "message");
+#ifdef _WINDOWS
+    size_t written = vsnprintf_s(message_text, sizeof(message_text),
+        _TRUNCATE, fmt, vargs);
+
+    message_len = (written < 0) ? sizeof(message_text) : written;
+#else
+    size_t written = vsnprintf(message_text, sizeof(message_text), fmt, vargs);
+    if (written < 0 || written >= sizeof(message_text)) {
```
hfstco, line 280: `size_t` is defined as unsigned int.

loglib/qlog_fns.c
```diff
@@ -0,0 +1,1049 @@
...
+    UNREFERENCED_PARAMETER(ticket);
+    UNREFERENCED_PARAMETER(ticket_length);
+#endif
+}
+
+/* log congestion control parameters
+* The congestion control is per path. 
+* We get an event per path id. If multipath is not enabled, only path[0] is logged.
+*/
+void qlog_fns_cc_dump_path(picoquic_cnx_t* cnx, picoquic_path_t* path, picoquic_packet_context_t* pkt_ctx,
```
hfstco, line 797: Perhaps we can log ssthresh here as well.

hfstco:
First, I think it's a great idea to write the qlog directly rather than converting it from the binlog at the end of each connection.

I find the`congestion_state_updated` event very interesting to log. However, difficult to implement due to the current implementation.

huitema:
@hfstco yes, we could take a deep look at the "congestion updated" event. Or rather, the event  "recovery/metrics_updated". This has been implemented by simply dumping a record each time key values like CWIN, bytes in flight or RTT are updated. Yes, it would be nice to also include the state of the congestion control algorithm, or key parameters such as for example the bottleneck rate and the min RTT for BBR. We could do that by having the Qlog code call an API exposed by the congestion algorithm. Or we could completely change the implementation and instead of issuing the "recovery/metrics_updated" event from the main picoquic code, let the congestion algorithm call a qlog logging API.

But this will have to be a different PR. The goal here is streamline the generation of qlog while keeping parity with existing implementation. Maybe we should open an issue for logging congestion events, so we do not forget.

PR #2023, opened 2026-02-08
Clang warning fix

karen07:
Without these fixes, it doesn’t build under Clang.

### DIFF
--- a/picoquic/picoquic.h
+++ b/picoquic/picoquic.h
@@ -429,7 +429,7 @@ typedef struct st_picoquic_tp_t {
 * quic context was initialized.
 */
 
-uint64_t picoquic_current_time(); /* wall time */
+uint64_t picoquic_current_time(void); /* wall time */
 uint64_t picoquic_get_quic_time(picoquic_quic_t* quic); /* connection time, compatible with simulations */
 
 /* Callback function for providing stream data to the application,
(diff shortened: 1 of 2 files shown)


huitema (commented):
@karen07 Picoquic does in fact build with Clang -- this is checked in the CI actions "CITestsClang", see github action ci-tests-clang.yml. I suppose that you are using Clang in a slightly different way than what the test does. Could you explain that? Do we need to tighten the test?

karen07:
I use Clang 21.1.6 on Arch Linux and build picoquic using this CMake file:
https://github.com/karen07/QUICTun/blob/main/CMakeLists.txt

The issue is probably related to -Wall -Wextra -Werror -Wpedantic.

huitema:
I can indeed reproduce the issue using  -Wall -Wextra -Werror -Wpedantic, I get a set of errors of type:

~~~
/Users/christianhuitema/Documents/projects/picoquic/picoquic/picoquic.h:432:31: error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]
uint64_t picoquic_current_time(); /* wall time */
                              ^
                               void
~~~

These issues are address in your PR, but I also get a few more, e.g.:
~~~
In file included from /Users/christianhuitema/Documents/projects/picoquic/picoquic/bbr.c:25:
/Users/christianhuitema/Documents/projects/picoquic/picoquic/cc_common.h:146:7: error: no newline at end of file [-Werror,-Wnewline-eof]
#endif
~~~

Did you fix those too?

karen07:
Probably we should add a mode either in tests or in CI where the whole **picoquic** is built with `-Wall -Wextra -Werror -Wpedantic`. I’m not sure yet where exactly to plug this in. If you point me to the right place, I can come with a PR.

huitema:
If you want, you can do that as another git action -- compile with setting the flags as a parameter to the cmake call. You probably do not need to execute the full test suite, just one basic test would be enough.

PR #2024, opened 2026-02-08
Expose TLS exporter support and secret export API

karen07:
Adds a public API to enable the TLS exporter feature on the master TLS context and to export keying material from an established TLS connection using a caller-provided exporter label.

### DIFF
--- a/picoquictest/tls_api_test.c
+++ b/picoquictest/tls_api_test.c
@@ -2332,6 +2332,96 @@ int tls_api_inject_hs_ack_test()
     return ret;
 }
 
+int tls_exporter_test()
+{
+    uint64_t simulated_time = 0;
+    uint64_t loss_mask = 0;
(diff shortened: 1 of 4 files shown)


karen07:
Yes, of course, I’ll add it)

karen07:
Done)

picoquictest/tls_api_test.c
```diff
@@ -2332,6 +2332,96 @@ int tls_api_inject_hs_ack_test()
...
+        }
+    }
+
+    if (test_ctx != NULL) {
+        tls_api_delete_ctx(test_ctx);
+    }
+
+    return ret;
+}
+
```
huitema, line 2424: Looks good!

karen07:
i use exporter in my project
https://github.com/karen07/QUICTun/blob/main/src/QUICTun.c

PR #2025, opened 2026-02-08
Fix shellcheck warnings and format

karen07:
I ran shellcheck, and it produced a couple of warnings.

### DIFF
--- a/ci/build_picotls.sh
+++ b/ci/build_picotls.sh
@@ -1,17 +1,21 @@
 #!/bin/sh
-#build last picotls master (for Travis)
+# build last picotls master (for Travis)
 
 # Build at a known-good commit
-COMMIT_ID= f350eab60742138ac62b42ee444adf04c7898b0d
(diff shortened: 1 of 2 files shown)


ci/build_picotls.sh
```diff
@@ -1,17 +1,21 @@
...
 
 # Build at a known-good commit
-COMMIT_ID= f350eab60742138ac62b42ee444adf04c7898b0d
+COMMIT_ID=f350eab60742138ac62b42ee444adf04c7898b0d
 
-cd ..
-# git clone --branch master --single-branch --shallow-submodules --recurse-submodules --no-tags https://github.com/h2o/picotls
+cd .. || exit
+# git clone --branch master --single-branch --shallow-submodules --recurse-submodules \
+#     --no-tags https://github.com/h2o/picotls
```
huitema, line 9: That code is commented out. We could just delete the comments...

PR #2027, opened 2026-02-13
Revisit dualqm

huitema:
Revisit the implementation of the dual Q algorithm to match details of the RFC.

It is very unclear that we need to do this -- the proposed modifications in this PR look a bit experimental.

The fundamental issue is that the default parameters proposed in the RFC appear to be designed for very low latency paths. They propose a short ramp up with the max threshold maybe a few milliseconds larger than the min threshold. In practice, if the RTT is large enough, the queue only starts decreasing after at best 1 RTT. If the case of a single connection, that means that if the queue starts increasing, its size will quickly run higher than the min RTT. We would expect the feedback to ramp up, with higher rates signalling worse congestion, but in practice the system behaves like an on-off controller.

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -209,11 +209,11 @@ int picoquic_flow_control_check_stream_offset(picoquic_cnx_t* cnx, picoquic_stre
 
         uint64_t new_bytes = new_fin_offset - stream->fin_offset;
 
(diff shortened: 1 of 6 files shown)


huitema:
Abandoning this PR. The proposed changes are not well justified, and they break the existing tests of the Prague CC.

PR #2028, opened 2026-02-14
Try io uring

huitema:
testing the implementation of IO-uring in the socket loop.


huitema:
At this point, the code compiles and link if we use the option "WITH_IO_URING=ON" in `cmake`. However, the socket tests do not yet work with that option. Checking this as a work in progress.

PR #2030, opened 2026-02-16
Make outgoing ECN function of default CC

huitema:
Document the default ECN mark for congestion control algorithms (CCA):

- ECN ECT 1 if the CCA supports both the generic ECN logic and the L4S logic.
- ECN ECT 0 if the CCA supports the generic ECN logic but not the L4S logic.
- Null if the CCA does not supports the ECN logic.

Change the tests definitions to only apply ECN marks in packets if the simulation requires ECN support, and in that case to set the ECN marks according to the default CCA.

Change the socket loop to set the ECN marks in packets according to the default CCA.

### DIFF
--- a/picoquic/picoquic_packet_loop.h
+++ b/picoquic/picoquic_packet_loop.h
@@ -308,7 +308,7 @@ int picoquic_packet_loop_win(picoquic_quic_t* quic,
 /* Following declarations are used for unit tests. */
 void picoquic_packet_loop_close_socket(picoquic_socket_ctx_t* s_ctx);
 int picoquic_packet_loop_open_sockets(uint16_t local_port, int local_af, int socket_buffer_size, int extra_socket_required,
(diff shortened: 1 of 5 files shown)


huitema:
The following graph shows two connection competing on the same path. The connection using Cubic starts first, the connection using C4 starts second. The path is managed by an L4S AQM.

image

Without the changes in this PR, both would be setting the outgoing ECN to ECT1, i.e., selecting the L4S mode. However, Cubic does not really work well with L4S, and these trials used to show the throughput of Cubic dropping to very little. With the change, Cubic selects ECT0. The L4S AQM puts the Cubic packets in the "classic" queue, and the L4S packets in the "L4S" queue. After a few RTT, we see the connections stabilizing, and sharing the bandwidth almost evenly.

huitema:
Close #2030

suhasHere (approved):

PR #2034, opened 2026-02-20
C4 ecn on off

huitema:
Modify the ECN response of C4 to "stop pushing" if the pushing phase resulted in two many ECN-EC marks.

This PR replaces the previous logic based on the number of successful pushes by an equivalent logic based on a "probe level". This simplifies the code a little, and allows for specific response in case of ECN feedback during push:

- if the measured data rate is growing, only move to the next probe level if there was no ECN-CE signalling of congestion during push.
- if the data rate was not growing and there was ECN-CE signalling, move to the newly created "probe level 0".

The probe level zero only pushes at 3.125% instead of 6.25% for the "default" mode (now probe level 1). This avoids pushing over the limit too hard if the AQM starts marking ECN packet with short queue levels, and thus avoids causing too many packet losses. Since the push rate is shorter, the probe level 0 uses shorter cycles to keep the same overall level of aggressiveness.

image

This ECN test shows the new algorithm in action. The link utilisation is very good, and the probing causes only very small periodic delay increases.


picoquic/c4.c
```diff
@@ -104,6 +125,13 @@
 #define C4_KAPPA ((double)(1.0/4.0))
 #define C4_ECN_SHIFT_G 4 /* g = 1/2^4, gain parameter for alpha EWMA */
 
+#define C4_PROBE_LEVEL_MAX 3
+#define C4_PROBE_LEVEL_DEFAULT 1
+
+uint64_t c4_push_rate_by_probe_level[C4_PROBE_LEVEL_MAX + 1] = {
+    C4_ALPHA_PUSH_VERY_LOW_1024, C4_ALPHA_PUSH_LOW_1024, C4_ALPHA_PUSH_1024, C4_ALPHA_PUSH_1024
+};
```
suhasHere, line 133: just confirming, the last 2 elements are exactly the same ?

huitema: Yes. The idea is that we want 3 successful increases (low, push, push) before detecting that the capacity has probably changed and starting "initial" again.

PR #2045, opened 2026-03-04
Add tests of migration disabled downloads.

huitema:
We had a report of a failure of interop against a server that was disabling migration. Adding a couple of tests of that scenario just to be sure.


huitema:
I performed a series of tests using picoquicdemo. I could force the issue by forcing the client to disable migration. I applied two fixes: make sure that picoquic demo only attempt to create a new path if the CLI arguments require testing multipath or migration, or if the server requires preferred address migration; preventing migration but not returning an error if the path creation attempt was prevented by the no-migration option. I then verified that the various migration scenarios still work if migration is enabled, but that the connections succeed even if migration is disabled.

Issue #2046, opened 2026-03-05
Embedding picoquic as a component in another CMake project


afrind:
When picoquic is used via add_subdirectory or FetchContent inside another CMake project, several problems arise:

1. Leaking paths to parent projects — public include directories use bare relative paths without BUILD_INTERFACE/INSTALL_INTERFACE generator expressions, so they leak into any parent project that links against picoquic-core.
2. Hardcoding library names that do not exist — LIB_PATH is built from a hardcoded list that includes picotls-fusion, which is not built on platforms without AES-NI support.
3. Need to export libraries to parent scope — when embedded via add_subdirectory, picoquic does not export the variables (picoquic_FOUND, picoquic_INCLUDE_DIR, picoquic_LIBRARIES) that picoquic-config.cmake would provide after a normal install, making it difficult for the parent project to re-export picoquic to its own consumers.

huitema:
Can we have some kind of test project that uses picoquic as a component. We can then build this project in a github action, and verify that symbols and libraries are properly exported.

afrind:
See https://github.com/afrind/picoquic/pull/2

Which has a series of commits each fixing a specific problem, a sample, and a CI workflow to validate.  CI is green there.  If it looks good I can make a PR in the main repo.

huitema:
I am trying to understand these fixes.

For "build interface", I think I get it -- found [a Stack Overflow entry explaining the issue](https://stackoverflow.com/questions/69890017/when-to-use-build-interface-and-install-interface-generator-expressions-with-tar).

The LIB_PATH fix is also cool. We should have done it long ago.

I am *not* sure about the way the dependencies on "threads" is handled, and the dependency on OpenSSL is definitely wrong. CMake already has code and options to decide whether OpenSSL is needed or not, and is present or not. We should be using the results of that code. There are build scenarios in which we do not use OpenSSL and use MbedTLS instead, and we definitely do not want to bring openssl in the build, even if it is available on the platform.

The embedding sample and the embedding CI are cool.

So, yes, I guess we can prepare a PR in the picoquic depot along those lines. We just need to fine tune the handling of openSSL.

[...]

PR #2049, opened 2026-03-05
Fix CMake embedding support for FetchContent/install/re-export scenario

afrind:
Addresses the three problems that arise when picoquic is embedded in another CMake project that is itself installed and re-exported to downstream consumers.

**Changes:**

- Wrap public include directories with `BUILD_INTERFACE`/`INSTALL_INTERFACE` generator expressions so build-tree paths don't leak into parent projects
- Build `LIB_PATH` dynamically from `PTLS_LIBRARIES` instead of hardcoding library names that may not exist on all platforms
- Add `picoquic::picoquic-core` as a namespaced alias and export it via an installed CMake targets file (`picoquic-targets.cmake`)
- Fix `picoquic-config.cmake` to call `find_dependency(Threads)` and `find_dependency(OpenSSL)` and include the targets file, so consumers get transitive dependencies resolved automatically

**Test:**

Adds `ci/embedding-test/` with a three-stage scenario: `sample_lib` embeds picoquic via FetchContent and installs itself; `sample_app` consumes `sample_lib` via `find_package` with no direct knowledge of picoquic. A new GitHub Actions workflow runs this chain on every push and PR.

The local test can be run with:
```
ci/embedding-test/run.sh
```

Fixes: #2046

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,12 @@ project(picoquic
         DESCRIPTION "picoquic library"
         LANGUAGES C CXX)
 
-find_package(Threads REQUIRED)
+option(WITH_THREADS "Build with Threads support" ON)
+if(WITH_THREADS)
+    find_package(Threads REQUIRED)
+else()
+    message(FATAL_ERROR "WITH_THREADS=OFF is not yet fully implemented")
+endif()
 
 option(DISABLE_DEBUG_PRINTF "Disable Picoquic debug output" OFF)
 option(ENABLE_ASAN "Enable AddressSanitizer (ASAN) for debugging" OFF)
--- a/picoquic-config.cmake.in
+++ b/picoquic-config.cmake.in
@@ -1,8 +1,12 @@
 @PACKAGE_INIT@
 
 include(CMakeFindDependencyMacro)
-find_dependency(Threads)
-find_dependency(OpenSSL)
+if(@WITH_THREADS@)
+    find_dependency(Threads)
+endif()
+if(@WITH_OPENSSL@)
+    find_dependency(OpenSSL)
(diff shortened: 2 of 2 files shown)


afrind:
@huitema let me know if there's anything more you would like here.

huitema:
It seems all good. I am ready to approve it, but I would prefer to have another review from people who are already using picoquic embedded in another app. Let's set a deadline for those reviews to end of day today!

TimEvens:
+1

CMakeLists.txt
```diff
@@ -12,7 +12,12 @@ project(picoquic
         DESCRIPTION "picoquic library"
         LANGUAGES C CXX)
 
-find_package(Threads REQUIRED)
+option(WITH_THREADS "Build with Threads support" ON)
```
suhasHere, line 15: Let's not make it default ON given  esp32 deployments. @TimEvens also agree on it.  Otherwise the PR looks fine

afrind: It has to default ON for now because the cmake doesn't actually support building without it right now.  Christian suggested a follow up PR:

> We may at some point want to make the dependency on threads optional. It should be possible to build a picoquic library that does not depend on threads. The dependency is mostly in the "socket loop" code, and there is already an option for application to plug in function pointers for their own library. There are definitely platforms (e.g., ESP32) for which we do not want to use threads. That is some work of course, maybe a different PR.

huitema: Yes, let's open an issue for that, and do a second PR to verify the build with the thread option off.

huitema: I opened issue #2058 to track the "build without threads" issue.

Issue #2044, opened 2026-03-04
Stream leak on server side when  unidirectional stream is recieved


fbussery:
**Description**

On the client side, I create a unidirectional stream using picoquic_get_next_local_stream_id().
I send data to the server and properly set the FIN flag.

On the server side, I receive the data and correctly get the stream FIN notification.

However, I noticed that upon reception, even if I reset or discard the stream, it does not seem to be released automatically. The stream remains allocated.

The only way I managed to release it was by manually calling:

```
picoquic_stream_head_t* stream = picoquic_find_stream(cnx, stream_id); 
picoquic_delete_stream(cnx, stream);
```
But since this relies on internal APIs, I assume this is not the correct approach.

So my question is:
What is the proper way to release a remotely-initiated unidirectional stream after receiving its FIN on the server side?

For reference, I am using picoquic at commit:
(2ccf08ff5b94a5a39b1692accddf09ac4953e240    Merge pull request #1679 from private-octopus/media_test_check_reset)

fbussery:
in fact deleting the stream is not the good approach. I see sometime crashes later when calling picoquic_incoming_packet in picoquic_delete_stream_if_closed

huitema:
The unidir steam context should indeed be deleted when receiving a FIN from the sender *and delivering it and all pending data*.

The first step there is to add a test that reproduces the issue.

fbussery:
so, means that on picoquic_callback_stream_fin reception, except may be doing a picoquic_unlink_app_stream_ctx, the stream resources should close automatically. I will see how to make a test to reproduce. thanks.

fbussery:
Just one thing I may not undestood. In my mind, the deletion of the stream should have been done after the callback which is called from function picoquic_stream_network_input
I see as expected the call to picoquic_delete_stream_if_closed()

My concern is that picoquic_delete_stream_if_closed never delete the stream because picoquic_is_stream_acked() is always null.

But I suppose that this can't be else since for unidirectional, on server side, there is no stream ack (or may be I do not understood)

Do we really need to check picoquic_is_stream_acked(stream) for unidirectional streams when server?

Here a proposition of fix that seems to work. Still testing
[0001-Fix-stream-leak-on-FIN-reception-for-receive-only-un.patch](https://github.com/user-attachments/files/25769521/0001-Fix-stream-leak-on-FIN-reception-for-receive-only-un.patch)

huitema:
[...]

PR #2051, opened 2026-03-06
Fix stream leak on FIN reception for receive-only unidirectional streams

fbussery:
picoquic_delete_stream_if_closed gates deletion on picoquic_is_stream_acked, which checks whether all sent data has been acknowledged. For a receive-only unidirectional stream, the local side never sends anything So there is no ack.

The ACK check is only meaningful when the local side has actually transmitted data. Skip it for receive-only streams and delete immediately once is_closed is set.

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -141,9 +141,22 @@ int picoquic_delete_stream_if_closed(picoquic_cnx_t* cnx, picoquic_stream_head_t
         ret = 1;
     }
     
-    /* We only delete the stream if there are no pending retransmissions */
-    if (stream->is_closed && picoquic_is_stream_acked(stream)) {
-        picoquic_delete_stream(cnx, stream);
+    /* for bidirectional streams, We only delete the stream if there are no pending retransmissions
+     * for unidirectonal, in server mode, we won't have any ack. So we delete stream
+     */
+    if (stream->is_closed)
+    {
+        if ((picoquic_is_stream_acked(stream)) ||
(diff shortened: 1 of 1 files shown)


fbussery:
Pull request related to https://github.com/private-octopus/picoquic/issues/2044

We don't notice anymore stream leaks on server side with this fix proposition. 
We did test on 10 client/servers for more than 10 hours

fbussery:
sorry, commit mistake... need update...

picoquic/frames.c
```diff
@@ -141,8 +141,10 @@ int picoquic_delete_stream_if_closed(picoquic_cnx_t* cnx, picoquic_stream_head_t
         ret = 1;
     }
     
-    /* We only delete the stream if there are no pending retransmissions */
-    if (stream->is_closed && picoquic_is_stream_acked(stream)) {
+    /* for bidirectional streams, We only delete the stream if there are no pending retransmissions
+     * for unidirectonal, in server mode, we won't have any ack. So we delete stream
+     */
+    if (stream->is_closed && ((!cnx->client_mode && !IS_BIDIR_STREAM_ID(stream->stream_id) && IS_CLIENT_STREAM_ID(stream->stream_id))  || picoquic_is_stream_acked(stream))) {
```
huitema: This is not quite the right test. We can have unidir streams initiated by either client or server, and we should make sure that both are closed.

huitema:
Superseded by PR #2052

Issue #2055, opened 2026-03-09
qlog recovery metrics_updated event not written to qlogs


hfstco:
I think there is a bug in the streaming qlog implementation. The qlogs directly created by picoquicdemo no longer include any recovery::metrics_updated events. 

.qlog files converted from the .log files via the `picolog_t` tool include them.

So the events will be written to the .log file, but not to the .qlog file.

Example:
`output-picoquicdemo.server.qlog` and `output-picoquicdemo.server.log` are the files written by `picoquicdemo`
`output-picolog_t.qlog` is the converted qlog of `output-picoquicdemo.server.log` via `./picolog_t -o . -f qlog ./output-picoquicdemo.server.log` tool
[qlogs.zip](https://github.com/user-attachments/files/25842479/qlogs.zip)

I am using the master branch of picoquic.
server: `./picoquicdemo -k {keyfile} -c {certfile} -T /tmp/{current_time}/{impl}/ticket_store.bin -N /tmp/{current_time}/{impl}/token_store.bin -D -p 50011 -Z 21474836 -G cubic -q /tmp/{current_time}/{impl}/`
client: `./picoquicdemo -G cubic -q /tmp/{current_time}/{impl}/ -Z 21474836 quic.hfst.dev 50011 /10000000`

hfstco:
Looks like, `binlog_cc_dump()` is called first and sets the `path->is_cc_data_updated` to zero. https://github.com/private-octopus/picoquic/blob/07b2bad4a81c8c828b6bba12cd77e00095cc226b/picoquic/logwriter.c#L1172

In the following call of `qlog_fns_cc_dump()` `path->is_cc_data_updated` is always set to zero. https://github.com/private-octopus/picoquic/blob/07b2bad4a81c8c828b6bba12cd77e00095cc226b/loglib/qlog_fns.c#L886

hfstco:
As a simple fix, we can skip setting `path->is_cc_data_updated = 0` if `cnx->qlog_ctx != NULL` is set in `binlog_cc_dump()`.
And set `path->is_cc_data_updated = 0` in `qlog_fns_cc_dump()` afterwards.

But to be honest, that is a very dirty fix.

hfstco:
I think if the binlog feature is removed in the future, the first PR is the better fit.

But if both logging options will remain in the codebase, something like the second PR is better suited.

huitema:
[...]

PR #2057, opened 2026-03-09
Reset is_cc_data_updated in unified_log.c

hfstco:
Another draft to fix https://github.com/private-octopus/picoquic/issues/2055.

Instead of setting `is_cc_data_updated = 0` inside `binlog_cc_dump()` and `qlog_fns_cc_dump()`, we set it to 0 in `picoquic_log_cc_dump()`.

### DIFF
--- a/picoquictest/multipath_qlog_ref.txt
+++ b/picoquictest/multipath_qlog_ref.txt
@@ -195,7 +195,7 @@
     "frame_type": "path_ack", "path_id": 0, "ack_delay": 0, "acked_ranges": [[0, 4]]}, { 
     "frame_type": "path_ack", "path_id": 1, "ack_delay": 0, "acked_ranges": [[0, 0]]}, { 
     "frame_type": "stream", "id": 4, "offset": 4222, "length": 1394, "fin": false , "has_length": false, "begins_with": "7e7f808182838485"}]}],
-[44904, 0, "recovery", "metrics_updated", {"cwnd": 15360,"pacing_rate": 7097966,"bytes_in_flight": 5870,"smoothed_rtt": 21637,"min_rtt": 20286,"latest_rtt": 20286,"app_limited": 1}],
+[44904, 0, "recovery", "metrics_updated", {"bytes_in_flight": 5870,"smoothed_rtt": 21637}],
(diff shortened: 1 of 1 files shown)


picoquictest/multipath_test.c
```diff
@@ -1840,10 +1840,8 @@ int multipath_trace_test_one(int use_qlog_streaming)
         if (use_qlog_streaming) {
             picoquic_set_qlog(test_ctx->qserver, ".");
         }
-        else {
-            (void)picoquic_file_delete(MULTIPATH_TRACE_BIN, NULL);
-            picoquic_set_binlog(test_ctx->qserver, ".");
-        }
+        (void)picoquic_file_delete(MULTIPATH_TRACE_BIN, NULL);
+        picoquic_set_binlog(test_ctx->qserver, ".");
```
huitema: Please do not do that. Yes, we do have a problem with the multipath qlog tests, but the solution is not to add binlog. We really want to verify what is happening when binlog is not initiated.

I did a little investigation, and I think that the solution is a bit more convoluted:

1- verify manually that the output of `multipath_qlog_fns` is good.
2- remove the "bug compat" hack
3- replace `multipath_qlog_ref.txt` by the output of  `multipath_qlog_fns`
4- remove the current code for `multipath_qlog` and replace it with `multipath_qlog_fns`

hfstco:
There is a problem getting the `path_ctx` in `qlog_fns_get_path_context()`. Maybe you can fix it.

loglib/qlog_fns.c
```diff
@@ -125,10 +125,10 @@ qlog_fns_path_context_t* qlog_fns_get_path_context(qlog_fns_context_t* ctx, pico
         if (path_ctx->unique_path_id == unique_path_id) {
             return path_ctx;
         }
+        path_ctx_prev = path_ctx;
         path_ctx = path_ctx->next;
         nb_ctx++;
     }
-    path_ctx_prev = path_ctx;
```
huitema, line 131: Misplaced instruction -- that was the cause of the memory leak.

huitema: Fixing the memory leak changes the resulting log. The cc_dump logs only contain the values that differ from previous. The bug caused the caches of previous values to be unlinked (but leaking in memory), so the previous values were zero. With the fix, the previous values are correct, so the cached values are not repeated. Fixing that by updating the reference file.

PR #2064, opened 2026-03-12
End to end implementation and test of preferred address migration

huitema:
This is a big PR, because proper end to end implementation of the preferred address migration feature requires defining a "local" port on the server in addition to the server port, so that connection can migrate from the server port (which may be shared) to a local port specific to the server instance and thread. This required adding configuration options for the server and testing them with unit test, and then doing manual end to end test using `picoquicdemo`.

The format of the "server port" option changes from just specifying the server port number to:

```
-p ['S']<server_port>[':' <local_port>]
```
Where the letter S, when present, indicates that the port is shared. The socket loop code is modified to open sockets for the local port and the additional server port if both are specified, and to open the server port in shared mode (`SO_REUSEADDR`) if the port is shared. The local port will be used as the default for outgoing connections.

We add two new configuration parameters:

```
-4 preferred_ipv4_address
-6 preferred_ipv6_address
```

If any of them is present, the configuration will set a Preferred Address parameter in the QUIC context, setting the port value for that parameter to the local port.

The end to end tests found some bugs in the handling of preferred address (byte order of port values in preferred address parameter), and in the QLOG rendering of the preferred address. These were of course fixed.

### DIFF
--- a/loglib/qlog_fns.c
+++ b/loglib/qlog_fns.c
@@ -625,11 +625,16 @@ void qlog_fns_preferred_address(FILE* f, const uint8_t* bytes, uint64_t len)
     uint64_t cid_len;
     const uint8_t* end_bytes = bytes + len;
 
-    fprintf(f, "\"ip_v4\": \"");
-    bytes = qlog_frame_hex_string(f, bytes, end_bytes, 4);
+    fprintf(f, "{");
+    if (len < 4) {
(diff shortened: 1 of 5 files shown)


huitema:
@hfstco I will be happy to get your review, but I don't want to burden you with that. Do as you like!

PR #2066, opened 2026-03-13
Extend CONNECTION_CLOSE APIs to support reason string

lminiero:
As discussed in #2040, this is an attempt to provide ways for users for both providing a reason when closing a connection themselves, and get access to the remote reason when receiving a CONNECTION_CLOSE instead.

On the sending side, I reused the existing `char const * local_error_reason` that the context has already, and created a new public `picoquic_close_ex` function where you can also provide a reason:

    int picoquic_close_ex(picoquic_cnx_t* cnx, uint64_t application_reason_code, char const* error_reason);

Internally, it's a clone of what `picoquic_close` did, which also keeps track of the reason: I only copy the pointer, which means that it's the application responsibility to ensure the string remains available until the connection is freed. I thought about whether this should use `picoquic_string_duplicate`, but considering that `local_error_reason` was already internally used in different places I decided not to touch that.  `picoquic_close` now calls `picoquic_close_ex` with `NULL` as a reason.

I also had to modify `picoquic_format_application_close_frame`, since (unlike `picoquic_format_connection_close_frame`) that was never encoding the reason string, and was always just writing a 0 reason length.

On the receiving side, I added a new `char * remote_error_reason` to the context, which is supposed to be a string that will be allocated: as such, it's freed when the connection is destroyed. Considering that there wasn't a helper method to decode a string, I added a `picoquic_frames_charz_decode` method (a counterpart to `picoquic_frames_charz_encode`) that uses the parsed length to conditionally allocate a string. That helper is now invoked by both `picoquic_decode_connection_close_frame` and `picoquic_decode_application_close_frame` in place of the `picoquic_frames_length_data_skip` that was there before.

I haven't added a way yet to access the remote reason string from the application, meaning that, as done in this patch, it's only available if you include `picoquic_internal.h`. Considering that there isn't a way to get the local reason either, it may be worthwhile to provide an extended version of `picoquic_get_close_reasons` too.

I tested all this into my own library and it seems to work as expected, but of course a proper review will be needed, especially considering this now also allocates a string. It may be a good idea to add a check in both `picoquic_decode_connection_close_frame` and `picoquic_decode_application_close_frame` to check whether `remote_error_reason` exists, and in case free it and NULL-ify first, as otherwise I guess we risk a memory leak if two CONNECTION_CLOSE frames containing a reason are parsed in a row.

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -4389,12 +4389,20 @@ uint8_t * picoquic_format_connection_close_frame(picoquic_cnx_t* cnx,
 
 const uint8_t* picoquic_decode_connection_close_frame(picoquic_cnx_t* cnx, const uint8_t* bytes, const uint8_t* bytes_max)
 {
-    bytes = picoquic_frames_varint_decode(bytes + 1, bytes_max, &cnx->remote_error);
+    uint64_t reason_length = 0;
 
-    if (bytes == NULL ||
-        (bytes = picoquic_frames_varint_skip(bytes, bytes_max)) == NULL ||
-        (bytes = picoquic_frames_charz_decode(bytes, bytes_max, &cnx->remote_error_reason)) == NULL)
-    {
+    if ((bytes = picoquic_frames_varint_decode(bytes + 1, bytes_max, &cnx->remote_error)) != NULL &&
+        (bytes = picoquic_frames_varint_skip(bytes, bytes_max)) != NULL &&
+        (bytes = picoquic_frames_varint_decode(bytes, bytes_max, &reason_length)) != NULL){
+        const uint8_t* first_byte = bytes;
+        if (reason_length != 0 &&
+            (bytes = picoquic_frames_fixed_skip(bytes, bytes_max, reason_length)) != NULL &&
+            cnx->remote_error_reason == NULL) {
+            cnx->remote_error_reason = picoquic_string_create((char*)first_byte, (size_t)reason_length);
+        }
+    }
+
+    if (bytes == NULL) {
         picoquic_connection_error(cnx, PICOQUIC_TRANSPORT_FRAME_FORMAT_ERROR,
             picoquic_frame_type_connection_close);
     }
--- a/picoquic/util.c
+++ b/picoquic/util.c
@@ -861,30 +861,6 @@ const uint8_t* picoquic_frames_cid_decode(const uint8_t* bytes, const uint8_t* b
     return bytes;
 }
 
-const uint8_t* picoquic_frames_charz_decode(const uint8_t* bytes, const uint8_t* bytes_max, char ** s)
-{
-    uint64_t length;
-    if ((bytes = picoquic_frames_varint_decode(bytes, bytes_max, &length)) != NULL && length > 0) {
-        if (length > PICOQUIC_CONNECTION_ID_MAX_SIZE ||
-            bytes + length > bytes_max) {
-            bytes = NULL;
-        }
-        else {
-            *s = picoquic_string_create((char const *)bytes, (size_t)length);
-            if (s == NULL) {
-                bytes = NULL;
-            }
-            else {
-                memcpy(*s, bytes, length);
-                char *str = *s;
-                str[length] = 0;
-                bytes += length;
-            }
-        }
-    }
-    return bytes;
-}
-
 /* Predict length of a varint encoding */
 size_t picoquic_frames_varint_encode_length(uint64_t n64)
 {
--- a/picoquic/picoquic_utils.h
+++ b/picoquic/picoquic_utils.h
@@ -177,7 +177,6 @@ const uint8_t* picoquic_frames_uint32_decode(const uint8_t * bytes, const uint8_
 const uint8_t* picoquic_frames_uint64_decode(const uint8_t * bytes, const uint8_t * bytes_max, uint64_t * n);
 const uint8_t* picoquic_frames_length_data_skip(const uint8_t * bytes, const uint8_t * bytes_max);
 const uint8_t* picoquic_frames_cid_decode(const uint8_t * bytes, const uint8_t * bytes_max, picoquic_connection_id_t * cid);
-const uint8_t* picoquic_frames_charz_decode(const uint8_t* bytes, const uint8_t* bytes_max, char ** s);
 
 #define VARINT_LEN(bytes) (((uint8_t)1) << ((bytes[0] >> 6)&3))
 #define VARINT_LEN_T(bytes, t_len) (((t_len)1) << ((bytes[0] >> 6)&3))


picoquic/util.c
```diff
@@ -861,6 +861,29 @@ const uint8_t* picoquic_frames_cid_decode(const uint8_t* bytes, const uint8_t* b
...
+const uint8_t* picoquic_frames_charz_decode(const uint8_t* bytes, const uint8_t* bytes_max, char ** s)
+{
+    uint64_t length;
+    if ((bytes = picoquic_frames_varint_decode(bytes, bytes_max, &length)) != NULL && length > 0) {
+        if (length > PICOQUIC_CONNECTION_ID_MAX_SIZE ||
+            bytes + length > bytes_max) {
+            bytes = NULL;
+        }
+        else {
+            *s = picoquic_string_create(NULL, length + 1);
```
huitema:
```suggestion
*s = picoquic_string_create(NULL, (size_t)(length + 1));
```
Fix warning on platforms where size_t is not 64 bits.

picoquic/util.c
```diff
@@ -861,6 +861,29 @@ const uint8_t* picoquic_frames_cid_decode(const uint8_t* bytes, const uint8_t* b
...
+const uint8_t* picoquic_frames_charz_decode(const uint8_t* bytes, const uint8_t* bytes_max, char ** s)
+{
+    uint64_t length;
+    if ((bytes = picoquic_frames_varint_decode(bytes, bytes_max, &length)) != NULL && length > 0) {
+        if (length > PICOQUIC_CONNECTION_ID_MAX_SIZE ||
+            bytes + length > bytes_max) {
+            bytes = NULL;
+        }
+        else {
+*s = picoquic_string_create(NULL, (size_t)(length + 1));
```
huitema:
```suggestion
            *s = picoquic_string_create(NULL, (size_t)(length + 1));
```
proper alignment.

picoquic/util.c
```diff
@@ -861,6 +861,29 @@ const uint8_t* picoquic_frames_cid_decode(const uint8_t* bytes, const uint8_t* b
...
+const uint8_t* picoquic_frames_charz_decode(const uint8_t* bytes, const uint8_t* bytes_max, char ** s)
+{
+    uint64_t length;
+    if ((bytes = picoquic_frames_varint_decode(bytes, bytes_max, &length)) != NULL && length > 0) {
+        if (length > PICOQUIC_CONNECTION_ID_MAX_SIZE ||
+            bytes + length > bytes_max) {
+            bytes = NULL;
+        }
+        else {
+            *s = picoquic_string_create(NULL, (size_t)(length + 1));
```
huitema: You could achieve the same results with:
```suggestion
            *s = picoquic_string_create((char const *)bytes, (size_t)length);
```
Then you would not need the code that copies the string and you could delete the next lines.

lminiero:
Hi @huitema, thanks for the quick review and the first fixes!

Yep, I had assumed the malloc could be problematic. I'm not sure what you're suggesting as a fix, though. Do you mean that I should get rid of `picoquic_frames_charz_decode` and revert that to `picoquic_frames_length_data_skip` in both `picoquic_decode_connection_close_frame` and `picoquic_format_application_close_frame`, and then depending on whether `picoquic_frames_length_data_skip` found something, create the string at an upper level instead? Or that it's ok to allocate the string in `picoquic_frames_charz_decode`, but that it should first skip the data and only then allocate the string using the constructor?

(as a side note, I just noticed there's a `length > PICOQUIC_CONNECTION_ID_MAX_SIZE` check that is an ugly copy/paste error, I should probably fix that too :grin: )

lminiero:
> Or that it's ok to allocate the string in `picoquic_frames_charz_decode`, but that it should first skip the data and only then allocate the string using the constructor?

Assuming this is what you meant, would changing the function like this work for you?

```
const uint8_t* picoquic_frames_charz_decode(const uint8_t* bytes, const uint8_t* bytes_max, char ** s)
{
    uint64_t length;
    const uint8_t *reason;
    if ((bytes = picoquic_frames_varint_decode(bytes, bytes_max, &length)) == NULL) {
        return NULL;
    }
    if (length > 0) {
        reason = bytes;
        bytes = picoquic_frames_fixed_skip(bytes, bytes_max, length);
        if (bytes != NULL && reason + length == bytes) {
            *s = picoquic_string_create((const char *)reason, length);
        }
    }
    return bytes;
}
```

lminiero:
For the other approach (charz_decode just tracks start of reason, and skips the rest), the patch could look like this. It has some duplicated code in the two decode functions, but is probably cleaner, as that's where the reason string is allocated, and only if a reason string was found. Any time one of the "decode close frame" functions is called, we also free any existing reason string we may have had from before.

```
diff --git a/picoquic/frames.c b/picoquic/frames.c
index 0e3daa85..6704b453 100644
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -4389,16 +4389,25 @@ uint8_t * picoquic_format_connection_close_frame(picoquic_cnx_t* cnx,
 
 const uint8_t* picoquic_decode_connection_close_frame(picoquic_cnx_t* cnx, const uint8_t* bytes, const uint8_t* bytes_max)
 {
+    const uint8_t *reason = NULL;
+    if (cnx->remote_error_reason != NULL) {
+        free((void*)cnx->remote_error_reason);
+        cnx->remote_error_reason = NULL;
+    }
+
     bytes = picoquic_frames_varint_decode(bytes + 1, bytes_max, &cnx->remote_error);
 
     if (bytes == NULL ||
         (bytes = picoquic_frames_varint_skip(bytes, bytes_max)) == NULL ||
-        (bytes = picoquic_frames_charz_decode(bytes, bytes_max, &cnx->remote_error_reason)) == NULL)
+        (bytes = picoquic_frames_charz_decode(bytes, bytes_max, &reason)) == NULL)
     {
         picoquic_connection_error(cnx, PICOQUIC_TRANSPORT_FRAME_FORMAT_ERROR,
             picoquic_frame_type_connection_close);
     }
     else {
+        if (reason != NULL && bytes > reason) {
+            cnx->remote_error_reason = picoquic_string_create((const char *)reason, bytes-reason);
+        }
         picoquic_state_enum old_state = cnx->cnx_state;
         cnx->cnx_state = (cnx->cnx_state < picoquic_state_client_ready_start || cnx->crypto_context[picoquic_epoch_1rtt].aead_decrypt == NULL) ? picoquic_state_disconnected : picoquic_state_closing_received;
 
@@ -4433,16 +4442,25 @@ uint8_t * picoquic_format_application_close_frame(picoquic_cnx_t* cnx,
 
 const uint8_t* picoquic_decode_application_close_frame(picoquic_cnx_t* cnx, const uint8_t* bytes, const uint8_t* bytes_max)
 {
+    const uint8_t *reason = NULL;
+    if (cnx->remote_error_reason != NULL) {
+        free((void*)cnx->remote_error_reason);
+        cnx->remote_error_reason = NULL;
+    }
+
     bytes = picoquic_frames_varint_decode(bytes + 1, bytes_max, &cnx->remote_application_error);
 
     if (bytes == NULL ||
         /* TODO, maybe: skip frame type for compatibility with draft-13 */
-        (bytes = picoquic_frames_charz_decode(bytes, bytes_max, &cnx->remote_error_reason)) == NULL)
+        (bytes = picoquic_frames_charz_decode(bytes, bytes_max, &reason)) == NULL)
     {
         picoquic_connection_error(cnx, PICOQUIC_TRANSPORT_FRAME_FORMAT_ERROR,
             picoquic_frame_type_application_close);
     }
     else {
+        if (reason != NULL && bytes > reason) {
+            cnx->remote_error_reason = picoquic_string_create((const char *)reason, bytes-reason);
+        }
         cnx->cnx_state = (cnx->cnx_state < picoquic_state_client_ready_start) ? picoquic_state_disconnected : picoquic_state_closing_received;
         if (cnx->callback_fn) {
             (void)(cnx->callback_fn)(cnx, 0, NULL, 0, picoquic_callback_application_close, cnx->callback_ctx, NULL);
diff --git a/picoquic/picoquic_utils.h b/picoquic/picoquic_utils.h
index e81785c9..297186fb 100644
--- a/picoquic/picoquic_utils.h
+++ b/picoquic/picoquic_utils.h
@@ -177,7 +177,7 @@ const uint8_t* picoquic_frames_uint32_decode(const uint8_t * bytes, const uint8_
 const uint8_t* picoquic_frames_uint64_decode(const uint8_t * bytes, const uint8_t * bytes_max, uint64_t * n);
 const uint8_t* picoquic_frames_length_data_skip(const uint8_t * bytes, const uint8_t * bytes_max);
 const uint8_t* picoquic_frames_cid_decode(const uint8_t * bytes, const uint8_t * bytes_max, picoquic_connection_id_t * cid);
-const uint8_t* picoquic_frames_charz_decode(const uint8_t* bytes, const uint8_t* bytes_max, char ** s);
+const uint8_t* picoquic_frames_charz_decode(const uint8_t* bytes, const uint8_t* bytes_max, const uint8_t ** s);
 
 #define VARINT_LEN(bytes) (((uint8_t)1) << ((bytes[0] >> 6)&3))
 #define VARINT_LEN_T(bytes, t_len) (((t_len)1) << ((bytes[0] >> 6)&3))
diff --git a/picoquic/util.c b/picoquic/util.c
index 9075d458..61c1dc76 100644
--- a/picoquic/util.c
+++ b/picoquic/util.c
@@ -861,27 +861,15 @@ const uint8_t* picoquic_frames_cid_decode(const uint8_t* bytes, const uint8_t* b
     return bytes;
 }
 
-const uint8_t* picoquic_frames_charz_decode(const uint8_t* bytes, const uint8_t* bytes_max, char ** s)
+const uint8_t* picoquic_frames_charz_decode(const uint8_t* bytes, const uint8_t* bytes_max, const uint8_t ** s)
 {
     uint64_t length;
-    if ((bytes = picoquic_frames_varint_decode(bytes, bytes_max, &length)) != NULL && length > 0) {
-        if (length > PICOQUIC_CONNECTION_ID_MAX_SIZE ||
-            bytes + length > bytes_max) {
-            bytes = NULL;
-        }
-        else {
-            *s = picoquic_string_create(NULL, (size_t)(length + 1));
-            if (s == NULL) {
-                bytes = NULL;
-            }
-            else {
-                memcpy(*s, bytes, length);
-                char *str = *s;
-                str[length] = 0;
-                bytes += length;
-            }
-        }
+    const uint8_t *reason = NULL;
+    if ((reason = bytes = picoquic_frames_varint_decode(bytes, bytes_max, &length)) != NULL) {
+        bytes = picoquic_frames_fixed_skip(bytes, bytes_max, length);
     }
+    if(s != NULL)
+        *s = reason;
     return bytes;
 }
 ```

**Review by huitema — COMMENTED** [review]
I would rather not have a function `picoquic_frames_charz_decode` at all. A function like that will have to allocate memory to copy the string and add a final zero, and memory allocation in these low level functions is problematic.

If we do a function, I would define it like:

```
const uint8_t* picoquic_frames_length_data_decode(const uint8_t* bytes, const uint8_t* bytes_max,
           size_t * l, const uint8_t ** s);
```

With:

- l: pointer to size_t variable. on successful decode, contains the length of the string.
- s: pointer to the first character of the string. The string is l character long, *not* null terminated.
- return NULL is error, the input byte after the string if success.

That is, do not attempt to NULL-terminate the string, do not allocate any memory.

If the application does want to allocate memory, it will have to do something like:
```
if ((bytes = picoquic_frames_length_data_decode(bytes, bytes_max, &l, &b)) != NULL){
    s = picoquic_string_create((const char*)b, l);
}
```
You can do that for the connection close reason. Just be careful with data handling. Just to be prudent, it may be possible to receive the 'connection close` frame several times, so we should avoid memory leaks by either only taking the first string, or freeing the old string before copying.

File: picoquic/frames.c [hunk 4]
```diff
@@ -4433,12 +4433,19 @@ uint8_t * picoquic_format_application_close_frame(picoquic_cnx_t* cnx,
...
+            (bytes = picoquic_frames_fixed_skip(bytes, bytes_max, reason_length)) != NULL &&
+            cnx->remote_error_reason == NULL) {
+            cnx->remote_error_reason = picoquic_string_create((char*)first_byte, (size_t)reason_length);
+        }
+    }
+
+    if (bytes == NULL) {
         picoquic_connection_error(cnx, PICOQUIC_TRANSPORT_FRAME_FORMAT_ERROR,
             picoquic_frame_type_application_close);
     }
```
huitema, line 4459: I guessed it would be simpler to just write what I had in mind. This code version has three big changes:

1. It does not use charz decode, which means we don't really need to add that new function
2. The allocation of memory happens in the code specific to application frame decode, which
    matches the freeing of the reason string in the destructor of the connection object
3. We test that the reason is not already set, this keep the first version and avoid memory leak.

picoquic/util.c
huitema: I think that we don't need that function. I have tried really hard to not call malloc from decoding functions, and would prefer to not add it to util.c

huitema (commented):
I suppose that you are very busy with IETF 125, so I did a bit of a rewrite of the decoding function for the application close frame. I would like to remove the code creating the function `picoquic_frames_charz_decode`, because less code is better, but I will let you do that unless you object.

Otherwise we are almost done.

picoquic/frames.c
lminiero: This sounds close to the sample patch I had shared in my last comment, where I moved the string management to the frame decoding code: in that patch I always updated the string any time the function was called, but it does make sense to only do it once and ignore it for further calls. Ack on not needing the extra charz_decode function.

I see you only updated one of the two decode functions, so I'll do the same for the other one.

picoquic/util.c
```diff
@@ -861,6 +861,29 @@ const uint8_t* picoquic_frames_cid_decode(const uint8_t* bytes, const uint8_t* b
...
+            bytes + length > bytes_max) {
+            bytes = NULL;
+        }
+        else {
+            *s = picoquic_string_create((char const *)bytes, (size_t)length);
+            if (s == NULL) {
+                bytes = NULL;
+            }
+            else {
+                memcpy(*s, bytes, length);
```
huitema:
```suggestion
                memcpy(*s, bytes, (size_t)length);
```
That will fix the uint64_t/size_t warning for 32 bit builds.
Of course not needed if we decide to remove the function...

lminiero: Yep, I'm removing it now. I'm testing if everything still works and I'll push a commit in the next few minutes.

lminiero:
@huitema I _think_ I did everything correctly, but please double check :pray: 
Thanks again for the precious feedback and the patience!

huitema:
Thanks for this work, and thanks for keeping calm with my chaotic feedback!

PR #2070, opened 2026-03-18
H09 as wt protocol

huitema:
Work in progress.


CMakeLists.txt
```diff
@@ -260,6 +262,8 @@ set(PICOHTTP_HEADERS
      picohttp/h3zero_common.h
      picohttp/h3zero_uri.h
      picohttp/h3zero_url_template.h
+     picohhtp/h09_common.h
```
hfstco: Here is a typo. You wrote "picohhtp" instead of "picohttp". Should fix https://github.com/private-octopus/picoquic/issues/2071.

`"/home/runner/work/picoquic/picoquic/picohhtp/h09_common.h": No such file`

afrind: @huitema: I confirmed this is the problem

huitema: Thanks! I will fix that asap.

H09 is still work in progress. I need to add the proper handling of files, etc., and verify that I can use the code of the existing H09 over QUIC and be compatible with the interop runner specification...

huitema:
Abandon. There has to be another way to participate in the WT interops.

PR #2080, opened 2026-04-06
Fix WebTransport header handling for browser compatibility

suhasHere:
Three fixes for WebTransport over HTTP/3:

1. Use lowercase header names (RFC 9114 compliance)
   
2. Case-insensitive header matching when parsing
   
3. Strip quotes from wt-available-protocols values
   - HTTP Structured Fields (RFC 8941) encode strings with quotes
   - Browsers send: "moqt-16", "moqt-15" (quoted per RFC 8941)
   - Protocol matching now strips quotes before comparison

These fixes enable interoperability with browser WebTransport clients (Chrome, Firefox) that follow the HTTP/3 and Structured Fields specs.

### DIFF
--- a/picohttp/h3zero.h
+++ b/picohttp/h3zero.h
@@ -50,8 +50,8 @@ extern "C" {
 #define H3ZERO_WEBTRANSPORT_APPLICATION_ERROR(code) (0x52e4a40fa8dbull + code) /* see spec for skipping grease points when mapping codes */
 #define H3ZERO_USER_AGENT_STRING "H3Zero/1.0"
 
-#define H3ZERO_WT_AVAILABLE_PROTOCOLS "WT-Available-Protocols"
-#define H3ZERO_WT_PROTOCOL "WT-Protocol"
+#define H3ZERO_WT_AVAILABLE_PROTOCOLS "wt-available-protocols"
+#define H3ZERO_WT_PROTOCOL "wt-protocol"
 
 #define H3ZERO_CAPSULE_CLOSE_WEBTRANSPORT_SESSION 0x2843
 #define H3ZERO_CAPSULE_DRAIN_WEBTRANSPORT_SESSION 0x78ae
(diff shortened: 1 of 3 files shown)


picohttp/webtransport.c
```diff
@@ -286,9 +286,12 @@ int picowt_select_wt_protocol(h3zero_stream_ctx_t* stream_ctx, char const* suppo
             a++;
         }
         while (*a != ',' && *a != 0 && *a != ' ' && *a != '\t' && candidate_length < 254) {
-            candidate[candidate_length] = *a;
+            /* Skip quotes - HTTP structured fields use quoted strings */
+            if (*a != '"') {
+                candidate[candidate_length] = *a;
+                candidate_length++;
+            }
```
huitema, line 293: That code will strip out any '"' within the candidate, not just the first and the last. I assume that this works in our context.

picohttp/h3zero.h
```diff
@@ -50,8 +50,8 @@ extern "C" {
 #define H3ZERO_WEBTRANSPORT_APPLICATION_ERROR(code) (0x52e4a40fa8dbull + code) /* see spec for skipping grease points when mapping codes */
 #define H3ZERO_USER_AGENT_STRING "H3Zero/1.0"
 
-#define H3ZERO_WT_AVAILABLE_PROTOCOLS "WT-Available-Protocols"
-#define H3ZERO_WT_PROTOCOL "WT-Protocol"
+#define H3ZERO_WT_AVAILABLE_PROTOCOLS "wt-available-protocols"
+#define H3ZERO_WT_PROTOCOL "wt-protocol"
```
huitema, line 55: Can we make sure that the WT draft is updated to reflect that preference for lower case? The current values were copied directly from the draft.

huitema (approved):
Doing case independent comparison of header names is a good idea. The "de-quoting" code is a bit rough, but it works as long that the ALPN values do not include a quote character.

PR #2081, opened 2026-04-06
Fix WT-Protocol header: quote response value, fix length field, add test

afrind:
Three bugs fixed in WebTransport protocol negotiation:

1. h3zero_create_response_header_frame_ex sent WT-Protocol as a bare token (e.g. moqt-16) instead of a Structured Header String (e.g. "moqt-16") as required by the WebTransport spec.

2. The wt_protocol case in h3zero_parse_qpack_header_block wrote the parsed length into wt_available_protocols_length instead of wt_protocol_length, leaving wt_protocol_length always zero.

3. The parser now strips the surrounding quotes from the WT-Protocol response header value so callers receive the bare protocol name.

Add h3zero_wt_protocol_response_test to verify encode/decode round-trip produces the correct unquoted protocol name.

### DIFF
--- a/picohttp/h3zero.c
+++ b/picohttp/h3zero.c
@@ -31,6 +31,7 @@
  * The server will start the connection by sending a setting frame, which will
  * specify a zero-length dynamic dictionary for QPACK.
  */
+#include <stdio.h>
 #include <string.h>
 #include <stdlib.h>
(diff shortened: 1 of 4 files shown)


huitema (approved):
Sorry for the delay, and thanks for the PR. This is good.

PR #2082, opened 2026-04-06
Fix build failures under FreeBSD

zi0r:
- FreeBSD, like Apple, doesn't have sys/prctl.h: Follow APPLE logic and don't try to include it
- pthread_setname_np() requires 2 args under FreeBSD: pthread_setname_np(pthread_t thread, const char *name)

### DIFF
--- a/picoquic/sockloop.c
+++ b/picoquic/sockloop.c
@@ -92,7 +92,7 @@
 #endif
 #endif
 
-#ifndef __APPLE__
+#if !defined(__APPLE__) && !defined(__FreeBSD__)
 #ifdef __LINUX__
 #include <linux/prctl.h>  /* Definition of PR_* constants */
 #else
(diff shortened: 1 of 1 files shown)


picoquic/sockloop.c
```diff
@@ -1735,6 +1735,8 @@ void picoquic_internal_thread_setname(char const * thread_name)
 #else
 #ifdef __APPLE__
```
huitema:
```suggestion
#if defined(__APPLE__)
```

On some platforms, the pre-processor protest when seeing a "#elif" clause after #ifdef.

huitema (approved):
Looks good, with a minor nit.

huitema:
Thanks for the fix, @zi0r ! And yes, we want picoquic to work well on BSD, not just on Apple devices!

PR #2083, opened 2026-04-07
Picoquic ns n times

huitema:
Add an option in the simulator to repeat a simulation N times. while collecting separate qlog files and media traces for each simulation in the batch.


huitema:
@hfstco, @suhasHere  This is something I found useful during the development of C4, which includes many simulation cases. Simulations always have a part of randomness, and the results can vary from one run to the next. In the C4 development, I used a python script to launch the same simulation N times, but I only retained a pass/fail result for each run. In fact, it would be very nice to have statistics showing the spread of results, so as to make comparisons between code versions, or with different algorithms.

The next step would be to integrate the change in picoquic_ns, with a "number of runs" option, and to develop a series of python scripts to draw visualization of the statistics from the corresponding qlog files. But to be logical, that would mean somehow redefining the picoquic_ns and c4 projects. I would like to:

* Move the executable `picoquic_ns` to the main `picoquic` project, so that it would get built each time we update picoquic, e.g., after modifying one of the supported CC algorithms.

* Focus the project picoquic_ns on developing a collection of python scripts that can drive or explain simulations, using the executable developed in the picoquic project.

* Focus the project C4 on just writing the C4 specs and the test definitions for C4, using the simulation code developed in picoquic and the python scripts developed in picoquic_ns.

What do you think?

suhasHere:
I agree with this approach

huitema:
Well, if someone could help me there, that would be nice. I added an executable target "pico_sim" to the CMakeList. It mostly worked, except that the "embedded" test fail. That test failed because pico_sim had a dependency on "picohttptest", which was not built in the embedded build. There might be some CMake magic to make that work, but I could not, so I decided to just break the dependency.

I did that by moving "picoquic_ns.c" to the "picoquichttp" library (because it has a dependency on Qperf, which is part of the HTTP library, and by moving the "dualq_aqm" to the "picoquic" library, which already contains the definition of simulator links. That allows me to break the dependency on the piocquictest library, and indeed that works well in the VS project on Windows, and also  in the CMake project on Mac.

But it does not work on Ubuntu. Which is weird, because we have multiple projects with the same pattern.

So if someone understands what is going on, please help me!

hfstco:
I also agree with this approach.

Here is a patch that should fix the error.

[Refactor_build_configuration__adjust_PICO_SIM_dependencies,_introduce_conditional_BUILD_HT.patch](https://github.com/user-attachments/files/26571226/Refactor_build_configuration__adjust_PICO_SIM_dependencies._introduce_conditional_BUILD_HT.patch)
[Refactor_build_configuration__adjust_PICO_SIM_dependencies,_introduce_conditional_BUILD_HT.patch](https://github.com/user-attachments/files/26571226/Refactor_build_configuration__adjust_PICO_SIM_dependencies._introduce_conditional_BUILD_HT.patch)

I am not too deep into compilers. But I think the problem occurs because GCC uses a single-pass linker. We have to link the libraries in the correct order for GCC's left-to-right library resolution. I have tested it under Ubuntu 25.10 and 24.04 LTS. 

We have the following dependency tree with this PR:
picoquic-log → picoquic-core (for picoquic_packet_type_name, picoquic_frame_name, etc.)
picohttp-core → picoquic-core + picoquic-log (for picoquic_set_qlog)
picoquic-test → picoquic-core + picoquic-log + picohttp-core

So we end up with the following order:
picohttp-core → picoquic-log → picoquic-core

In addition, I have added a CMAKE option to enable the picoquic_sim executable and moved `picoquic/dualq_aqm.c` from the `PICOQUIC_CORE_HEADERS` to the `PICOQUIC_LIBRARY_FILES` files.

Edit: I forgot to revert the -O0 option back to -O3 within the patch. Please change it back to -O3.

huitema:
Thanks very much. @hfstco. The patch did indeed work!

hfstco:
Glad to hear that.

But please revert the optimization flag back to -O3. Forgot to revert it back.

huitema:
I think that I did reverse the option -- I replaced all -O0 by -O3.

huitema:
All clear now, just waiting for approval of at least one reviewer!

hfstco (approved):

PR #2084, opened 2026-04-08
cmake: export picohttp-core as picoquic::picohttp-core

afrind:
picohttp-core was installed as a static library but never added to the picoquic-targets export set, so find_package(picoquic) provided picoquic::picoquic-core but not picohttp-core. Consumers needing H3/HTTP support had no cmake target to link against.

- h3zero.h: add missing #include <stdint.h> / <stddef.h> so the header is self-contained and can be included without preconditions.

- CMakeLists.txt: 
  - add ALIAS picoquic::picohttp-core so the namespaced target works during FetchContent builds (matching picoquic::picoquic-core).
  - Add EXPORT picoquic-targets + INCLUDES DESTINATION to the install rule so find_package(picoquic) exposes picoquic::picohttp-core. 
  - Fix picohttp-core's INTERFACE_INCLUDE_DIRECTORIES to use BUILD_INTERFACE / INSTALL_INTERFACE so the source path isn't baked into installed targets.

- ci/embedding-test: extend install_test to build against picoquic::picohttp-core, validating the export end-to-end.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -438,8 +438,9 @@ if (BUILD_LOGLIB)
         PRIVATE
             ${PTLS_INCLUDE_DIRS}
         PUBLIC
-            picoquic
-            loglib)
+            $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/picoquic>
+            $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/loglib>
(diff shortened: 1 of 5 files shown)


huitema (approved):
Thanks for the fix. Looks good.

We should discuss how to handle the commits by "Claude".

huitema:
@afrind I think I broke the PR when trying to resolve the merge conflict. Sorry. I am not great at programming CMake! Can you help me cleanup this mess?

huitema (approved):
Thanks for the quick fixes!

PR #2085, opened 2026-04-09
Add wildcard '*' path matcher for WebTransport CONNECT dispatch

afrind:
Register a path entry with path "*" and path_length 1 to catch any incoming WT CONNECT that has no specific path match. Specific paths still take priority; the wildcard is recorded as a fallback during the scan and returned only when no exact prefix match is found.

Comparing only on paths discards the possibility that multiple :authority's could share the same picoquic, with different webtransport paths.  Also, applications may want to do different types of matching besides exact matching.

### DIFF
--- a/UnitTest1/unittest1.cpp
+++ b/UnitTest1/unittest1.cpp
@@ -3512,6 +3512,12 @@ namespace UnitTest1
             Assert::AreEqual(ret, 0);
         }
 
+        TEST_METHOD(picowt_baton_wildcard) {
+            int ret = picowt_baton_wildcard_test();
+
+            Assert::AreEqual(ret, 0);
+        }
+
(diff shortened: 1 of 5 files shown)


huitema (approved):
I like the idea, and it will certainly help deploying WT protocols, not to mention the weird requirements of the WT interop runner. We have a unit test, and that makes me happy. My only doubt is that the same `h3zero_find_path_item` command is used to locate a path usable for POST and one usable for CONNECT (and extended connect). That was not an issue before the wildcard, because the deployments could use separate URL for separate services, but there is no way to do that separation with a wildcard.

I guess that's still not much of an issue, because nobody uses POST, and because any endpoint could do a POST to an URL handled by MoQ. We should treat that as a separate issue, and I opened issue #2088.

PR #2086, opened 2026-04-09
Expose authority in WebTransport requests

afrind:
Add authority/authority_length fields to h3zero_header_parts_t, parse the :authority pseudo-header in incoming requests, and expose it via picowt_get_authority() in the pico_webtransport API.

### DIFF
--- a/picoquictest/h3zerotest.c
+++ b/picoquictest/h3zerotest.c
@@ -713,116 +713,116 @@ typedef struct st_qpack_test_case_t {
 static qpack_test_case_t qpack_test_case[] = {
     {
         qpack_test_get_slash, sizeof(qpack_test_get_slash),
-        { h3zero_method_get, qpack_test_string_slash, 1, NULL, 0, 0, 0, NULL, 0,
+        { h3zero_method_get, qpack_test_string_slash, 1, NULL /* authority */, 0 /* authority_length */, NULL, 0, 0, 0, NULL, 0,
         NULL /* wt_available_protocols */, 0 /* wt_available_protocols_length */, NULL /* wt_protocol */, 0 /* wt_protocol_length */, 0 }
     },
     {
         qpack_test_get_slash_null, sizeof(qpack_test_get_slash_null),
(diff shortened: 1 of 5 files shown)


picoquictest/h3zerotest.c
```diff
@@ -713,116 +713,116 @@ typedef struct st_qpack_test_case_t {
...
+        { h3zero_method_get, qpack_test_string_long, sizeof(qpack_test_string_long), NULL /* authority */, 0 /* authority_length */, NULL, 0, 0, 0, NULL, 0,
         NULL /* wt_available_protocols */, 0 /* wt_available_protocols_length */, NULL /* wt_protocol */, 0 /* wt_protocol_length */, 0 }
     },
     {
         qpack_connect_webtransport, sizeof(qpack_connect_webtransport),
-        { h3zero_method_connect, qpack_test_string_wtp, sizeof(qpack_test_string_wtp), NULL, 0, 0, 0,
+        { h3zero_method_connect, qpack_test_string_wtp, sizeof(qpack_test_string_wtp), NULL /* authority */, 0 /* authority_length */, NULL, 0, 0, 0,
         (uint8_t *)web_transport_str, CONNECT_TEST_PROTOCOL_WTP_LEN,
         NULL /* wt_available_protocols */, 0 /* wt_available_protocols_length */, NULL /* wt_protocol */, 0 /* wt_protocol_length */, 0 }
     },
```
huitema, line 814: We only have one test case with a WT connect method, and it does not test the encoding of "authority". It would be nice to add a case -- say, a modified version of this "qpack_connect_webtransport" that encodes the authority.

huitema (approved):
Looks reasonable. 

I wish we had added a test case for a WT Connect with the authority set. However, the code in `h3zero_create_connect_header_frame` is already setting the authority parameter, so that code path will be used by the end to end tests, so I guess it is OK.

afrind:
@huitema do you want me to follow up with more tests?

huitema:
If that's easy, yes. On a couple occasions, I was able to capture the H3 frames that happen during interop testing and add them to the list of test cases. If you can do some of that, that would be a great way to increase coverage + provide some assurance of non regression.

PR #2087, opened 2026-04-09
Qmux development

huitema:
QMUX support will be developed in stages:

* do a basic QMUX send test, to show that we can format an outgoing packet,
* do a basic QMUX receive test, to show that we can process an incoming packet,
* formatting of transport parameter frame
* processing of transport
* simulate a "back to back" QMUX connection
* simulate a "back to back" QMUX connection using TLS
* simulate the QMUX/TLS connection over network simulation
* add support for TCP in socket loop
* test end to end QMUX connection

There will be at least one PR for each of these stages.


huitema:
At this point, we have QMUX working on simulated TCP/TLS connections, using the same data structure and callbacks as QUIC. The missing parts are:

1- Using QX-Ping to monitor the quality of the connection and provide the feedback on data rates, RTT and window sizes, using the same API as QUIC.

2- Integrate QLOG -- but we may have to wait for guidance from the WG.

3- and of course, plug in a "real" socket loop for deployments.

huitema:
The previous series of commit gave us a refactoring of the socket loop, to make it easier to change for adding support of TCP sockets. That work is still pending. The first step is to be able to create "qmux" contexts from the description of configuration,
then configure them in clients, servers and socket loop.

huitema:
Ah, the painful wait... The QMUX sockloop is not quite complete yet, three functions remain to be written, and then we need to have the loop test pass. Couple of days to wait!

victorstewart (approved):
we addressed all offline. looks good!

PR #2090, opened 2026-04-16
picoquicdemo: add the possibility to specify server address when using multipath

### LINKED ISSUE #2073: [Multipath] Secondary path is activated but not utilized for data transfer (Bandwidth aggregation failed)
1. Network Environment
  1) Client (Windows 11):
    - Interface 1 (WLAN): 192.168.30.200
    - Interface 2 (Ethernet): 192.168.200.33
  2) Server (Rock 5T / Debian):
    - Interface 1 (WLAN): 192.168.30.209
    - Interface 2 (Ethernet): 192.168.200.34
  3)Networking
    - Asymmetric routing verified via ping -S. Cross-interface ICMP is fully functional.

2. Execution Commands
  1) Server: ./picoquicdemo -q /path/to/save/qlog/ -M -p 4443 -w .
  2) Client: .\picoquicdemo -q C:\\path\\to\\save\\qlog\\  -M -A "192.168.200.33/24" -n localhost 192.168.30.209 4443 test_vdo.mp4

3. Main Issue: No Bandwidth Aggregation
 - Path Utilization: Logs confirm both Path 0 and Path 1 are Active/Validated.
 - Observation: However, actual data traffic only flows through a single path (WLAN).
 - Failure: There is no combined bandwidth aggregation observed between the two interfaces.

4. Secondary Issue: Intermittent Transfer Failures
 - Abnormal Exit (One-time):  In one specific instance, the client received the session ticket but failed with Client exit with code = -1.
- This particular symptom has only been observed once so far.

5. Supporting Data
Attached: qlogs for both the successful (single-path) and failed (timeout) cases.
[qlog.zip](https://github.com/user-attachments/files/26107211/qlog.zip)

resowifix:
This submission is an attempt to resolve #2073.

It thus add the possibility to extend -A option with the targeted server ip :
`-A "10.0.0.1/3/10.1.1.1"`

It still supports the previous syntax, so `./picoquicdemo -A "10.0.0.1/3" -M 192.0.2.1 4433 index.html` is equivalent to `./picoquicdemo -A "10.0.0.1/3/192.0.2.1" -M 192.0.2.1 4433 index.html`

It was tested with namespaces on linux.


picoquicfirst/picoquicdemo.c
```diff
@@ -471,20 +474,27 @@ int picoquic_parse_client_multipath_config(char *mp_config, int *src_if, struct
...
+                valid_new_entry = 1;
+                if (token && picoquic_store_text_addr(&ip, token, server_port) == 0) {
+                  memcpy(alt_server_ip + (*nb_alt_paths), &ip, sizeof(struct sockaddr_storage));
+                }
+                else {
+                  memcpy(alt_server_ip + (*nb_alt_paths), default_server_ip, sizeof(struct sockaddr_storage));
+                }
+              }
             }
+          }
```
huitema, line 494: In that block, you are parsing "token2", which could be:

    NULL: interface index set to 0, destination address undefined.
    interface: interface index set to interface value, error if the value is not numeric, destination address undefined.
    interface/IPaddress: interface index set to interface value, error if the value is not numeric, destination address must be valid, server port set to default value.

I wonder whether we should also specify the port number, as in:

    interface/IPaddress/serverIP/port, or maybe
    interface/IPaddress/serverIP:port

but this extra function can probably wait.

The code uses atoi, which parses the leading digits of a string and ignore the following characters. It is a bit weak: if the user specifies -A "10.0.0.1/eth1/10.1.1.1" instead of -A "10.0.0.1/3/10.1.1.1", the parser will return the equivalent of -A "10.0.0.1/0/10.1.1.1" without detecting an error, and the user will be wondering why the new path is routed on the default interface. Of course that's also what the current code does, so that would not be a new bug. But still, it would be nice if we could fix that.

huitema (commented):
Thank you for this submission. We do need that kind of functionality. I have a minor comment on the parsing code. I think that as we are changing the code we should detect errors such as setting  -A "10.0.0.1/eth1/10.1.1.1" instead of -A "10.0.0.1/3/10.1.1.1". If you think that you can do that easily that would be nice.

picoquicfirst/picoquicdemo.c
resowifix: Thank you for your answer !
I think I should be able to do it, I'm going to try.

resowifix: This should be done now.

The "error" triggered by an invalid interface index is a message on stdout, and the code skips the entry.
If the error is different (invalid address for example) the code just skip, just like the current code. Should it be changed ?

I didn't do anything for the port number, but I should be able to do it if you think it is relevant.

huitema (approved):
Thanks for this PR, and for fixing the verification of the interface index!

PR #2095, opened 2026-05-04
perf: split due-now connections out of the wake tree

victorstewart:
The connection wake list currently uses one picosplay tree keyed by next_wake_time. That works for future timers, but it is a poor fit for large batches of connections that are already due.

In a 10K persistent-client server profile, the application main thread stalled around 7,272 live clients with 92.18% of sampled user cycles in picosplay_delete_hint. The hot path was:

  picoquic_prepare_next_packet_ex
  -> picoquic_get_earliest_cnx_to_wake
  -> picosplay_first
  -> picoquic_prepare_packet_ex
  -> picoquic_reinsert_by_wake_time
  -> picosplay_delete_hint

Added a separate ready FIFO for connections whose wake time is due now, kept the existing picosplay tree for future timers, and routed reinsertion based on the new wake time. This preserves timer ordering for future wakeups while making the common many-ready scheduling path O(1) and FIFO-fair. Keep the one-connection case cheap by avoiding the tree in the due-now path.

This ready/future split improves the scheduler microbench without relying on unfair ordering:

  1 connection all-ready: 6.87 ns/op -> 0.24 ns/op
  10K all-ready:         26.51 ns/op -> 1.11 ns/op
  50K mixed10:           37.16 ns/op -> 7.73 ns/op
  100K mixed10:          36.07 ns/op -> 3.56 ns/op

Real profile evidence:

  baseline 10K shaped, stalled at 7,272 live:
    picosplay_delete_hint 92.18% flat / 92.176% folded

  ready-FIFO prototype, 7,280 live:
    picosplay_delete_hint 0.50% flat / 0.269% folded

  later 10K and 41K live captures:
    picosplay_delete_hint stayed at 0.000% in the main app profile

Tests:

  cmake --build ../picoquic-pr-build-shared -j48

  ../picoquic-pr-build-shared/picoquic_ct getter splay stress ready_to_send ready_to_skip ready_to_zfin two_connections keep_alive

  ctest --test-dir ../picoquic-pr-build-shared --output-on-failure

01-baseline-stalled-7272
02-ready-fifo-7280
03-post-scheduler-wall-7388
04-maxadmit-10001
05-maxadmit-41013

### DIFF
--- a/picoquic/picoquic_internal.h
+++ b/picoquic/picoquic_internal.h
@@ -642,6 +642,8 @@ typedef struct st_picoquic_quic_t {
 
     struct st_picoquic_cnx_t* cnx_list;
     struct st_picoquic_cnx_t* cnx_last;
+    struct st_picoquic_cnx_t* cnx_wake_ready_first;
+    struct st_picoquic_cnx_t* cnx_wake_ready_last;
     picosplay_tree_t cnx_wake_tree;
 
     struct st_picoquic_cnx_t* cnx_in_progress;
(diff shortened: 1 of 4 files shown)


huitema (approved):
That's a very cool fix! Most of the performance tests that we did so far assess data throughput or message rate with a somewhat small number of connections. You are exploring a specific profile with a large number of connections, each of them with a relatively small load. I would not have guessed that the splay of active connections would become a bottleneck, and it is great that you found that!

I like the fix, using an O(1) ready list for the active connections in complement to the splay. I see that we are getting code coverage with a unit test. That's cool.

I wonder whether the existing performance test (pqbench) can be used to simulate the "many connections" scenario. We should probably try that once this PR is merged.

PR #2098, opened 2026-05-14
Reinsert on zero length fin add

RichLogan:
This is more of a question and a potential solution than a normal PR. 

In our application we typically close a stream after the last bytes that will be sent have already been sent, and so we close using add with a 0 length data and fin set to true.

One of our tests ensuring timely stream close has been quite flaky, and on investigating it seems like a 0 length does not today cause the stream context to be reinserted with a new wake time, so a fin might linger longer than expected(?). For us I think this is really only a test related issue, but it seems like we may want to reinsert when the fin is set even if it's a zero length byte array?

### DIFF
--- a/picoquic/sender.c
+++ b/picoquic/sender.c
@@ -248,6 +248,7 @@ int picoquic_add_to_stream_with_ctx(picoquic_cnx_t* cnx, uint64_t stream_id,
     int ret = 0;
     picoquic_stream_head_t* stream;
     PICOQUIC_THREAD_CHECK(cnx->quic);
+    unsigned int reinsert = 0;
         
     stream = picoquic_find_stream_for_writing(cnx, stream_id, &ret);
     if (ret == 0 && set_fin) {
@@ -258,6 +259,7 @@ int picoquic_add_to_stream_with_ctx(picoquic_cnx_t* cnx, uint64_t stream_id,
             }
         } else {
             stream->fin_requested = 1;
+            reinsert = 1;
         }
     }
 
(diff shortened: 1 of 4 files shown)


picoquic/sender.c
```diff
@@ -248,6 +248,7 @@ int picoquic_add_to_stream_with_ctx(picoquic_cnx_t* cnx, uint64_t stream_id,
     int ret = 0;
     picoquic_stream_head_t* stream;
     PICOQUIC_THREAD_CHECK(cnx->quic);
+    unsigned int reinsert = 0;
```
huitema:
```suggestion
```

The flag is probably not needed.

picoquic/sender.c
```diff
@@ -258,6 +259,7 @@ int picoquic_add_to_stream_with_ctx(picoquic_cnx_t* cnx, uint64_t stream_id,
             }
         } else {
             stream->fin_requested = 1;
+            reinsert = 1;
```
huitema:
```suggestion
```

The flag is not needed, see modification of final test.

picoquic/sender.c
```diff
@@ -296,10 +298,13 @@ int picoquic_add_to_stream_with_ctx(picoquic_cnx_t* cnx, uint64_t stream_id,
             }
         }
 
-        picoquic_reinsert_by_wake_time(cnx->quic, cnx, picoquic_get_quic_time(cnx->quic));
+        reinsert = 1;
```
huitema:
```suggestion
```

See next suggestion.

picoquic/sender.c
```diff
@@ -296,10 +298,13 @@ int picoquic_add_to_stream_with_ctx(picoquic_cnx_t* cnx, uint64_t stream_id,
...
     }
 
     if (ret == 0) {
+        if (reinsert) {
```
huitema:
```suggestion
        if (length > 0 || set_fin) {
```

This would test the condition without requiring the 'reinsert' flag,

huitema (commented):
Thank you for this fix! And specifically, thank you for adding a test of the issue. The scenario that you are using obviously was overlooked!

I am proposing a small change to make the code a bit simpler.

RichLogan:
Thanks @huitema, lgtm, not sure why I didn't just do that in the first place!

huitema (approved):
Thanks for this fix!

PR #2102, opened 2026-05-17
Implement draft-denis-tls-aegis-06 AEGIS TLS/QUIC

victorstewart:
Adds experimental AEGIS support for picoquic per [draft-denis-tls-aegis-06](https://www.ietf.org/archive/id/draft-denis-tls-aegis-06.html), when built with compatible picotls/libaegis support.

  Summary:
  - Add opt-in `WITH_AEGIS` CMake support, including fetched libaegis `0.10.1`.
  - Register `TLS_AEGIS_256_SHA512` (`0x1306`) and `TLS_AEGIS_128L_SHA256` (`0x1307`) for clients and servers.
  - Implement QUIC header protection for AEGIS per `draft-denis-tls-aegis-06`.
  - Preserve RFC 9001 AES-GCM behavior for QUIC Initial and Retry.
  - Add AEGIS negotiation, fallback, 0-RTT, Retry, Initial, and header-protection tests.
  - Add `aegisperf` for AEAD and in-process picoquic transfer benchmarks.

  Verification:
  - AEGIS targeted tests: 8/8 passed.
  - Default guard targeted tests: 9/9 passed.
  - AEGIS CTest: 2/2 passed.
  - Default CTest: 2/2 passed.
  - `git diff --check origin/master..HEAD`: clean.

Performance notes:
- AEGIS-128L: `14687.546 MB/s` (`1.53x` AES-128-GCM, `1.57x` AES-256-GCM, `7.00x` ChaCha20)
- AEGIS-256: `10106.602 MB/s` (`1.05x` AES-128-GCM, `1.08x` AES-256-GCM, `4.82x` ChaCha20)
- In the 4 MiB in-process picoquic transfer benchmark paired with that run:
    - AEGIS-128L: `843.431 MB/s` (`1.20x` AES-128-GCM, `1.19x` AES-256-GCM, `1.82x` ChaCha20)
    - AEGIS-256: `807.731 MB/s` (`1.15x` AES-128-GCM, `1.14x` AES-256-GCM, `1.74x` ChaCha20)
- At 16 KiB AEAD-only payloads, AES-GCM was faster than both AEGIS suites, so this PR does not claim blanket performance superiority across all packet sizes.


huitema (commented):
That great, thank you for the PR. AEGIS is cool. A modern AEAD cipher that uses a sponge constructand update the sponge using AES rounds. That has very nice properties.

Two small points, however:

1. Do we have a Github action that builds picoquic "with AEGIS"? It would ensure that we build AEGIS and verify non regression for all new PRs.

2. Could we make sure that the AEGIS tests are available in the Windows development environment, see unittest1.cpp.

Speaking of Windows, I suppose that the AEGIS library is available on Windows. Did you try to build it, or should we defer that and the Visual Studio Unit tests to a follow-up PR?

victorstewart:
addressed everything

huitema (approved):
Nice and clean. Thanks for addressing the review comments. And once we can, let's start publishing improved performance reports!

PR #2105, opened 2026-05-22
Update pico_sim to handle repetitions

huitema:
Update the pico_sim application to log independent qlog traces for each repetition of a simulation, and to provide a summary of each execution on the standard output.

### DIFF
--- a/p_sim/pico_sim.c
+++ b/p_sim/pico_sim.c
@@ -43,36 +43,30 @@ char const* get_spec_name(char const* s, char * buffer, size_t l_buf)
 {
     int last_dot = -1;
     int last_slash = -1;
-    int spec_len = 0;
     int x = 0;
 
     for (int i = 0; s[i] != 0; i++) {
-        if (s[i] == '/' || s[i] == '\\') {
-            last_slash = i;
-        }
-        else if (s[i] == '.')
+        if (s[i] == '.')
         {
-            last_dot = i;
+            last_dot = x;
+        }
+        if ((x + 1) < (int)l_buf) {
+            buffer[x] = s[x + last_slash];
+            x++;
+        }
+        if (s[i] == '/' || s[i] == '\\') {
+            last_slash = i+1;
+            last_dot = -1;
+            x = 0;
         }
     }
-
-    if (last_slash >= 0) {
-        last_slash += 1;
+    if (last_dot > 0 &&
+        last_dot <= x) {
+        buffer[last_dot] = 0;
     }
     else {
-        last_slash = 0;
+        buffer[x] = 0;
     }
-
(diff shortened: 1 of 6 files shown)


huitema:
This PR does just one thing, expose the results of a series of simulations in individual reports and qlog files. The early tests did expose an issue: when doing series of 100 tests, the vast majority of those provide nearly identical results. Variations, when they happen, seem due to minor computing issues, such as variations in the last decimals when converting integers to floating point. This is unfortunate, and we should strive to incorporate sources of randomness, such as for example using variable random seeds for the simulation of jitter. ~But this injection of randomness is a work by itself, and should come in a separate PR.~ _(Edited 5/24/2026 -- adding a random seed to jitter is both practical and useful.)_

huitema:
I think that we really need to add some form of randomization. This graph of 100 "bad wifi" connections using C4, Cubic and BBR is done using the current code, with fixed random seeds:

image

Compare now with the result of the same experiment, using random seeds:

image

There are huge differences. With fixed seeds about 90% of each trial have the same results, and the overall variations are very small, but with random seeds we see very large variations. Clearly, the random seeding results in interesting lessons, that are just absent in the fixed seeds variants. When our goal is non-regression testing, the fixed seeds are probably more convenient, but if we want to evaluate congestion control algorithms, we clearly must use the random seeds!

huitema:
The basic randomization only applies if the simulation specifies to apply some jitter on the link. There may be other forms of randomization, maybe in some AQM, but this is left for further study.

hfstco:
Looks good to me.  I like the idea of adding some randomization.

p_sim/pico_sim.c
```diff
@@ -39,6 +39,37 @@ void usage(void)
...
+    int last_slash = -1;
+    int x = 0;
+
+    for (int i = 0; s[i] != 0; i++) {
+        if (s[i] == '.')
+        {
+            last_dot = x;
+        }
+        if ((x + 1) < (int)l_buf) {
+            buffer[x] = s[x + last_slash];
```
hfstco: One small thing I have noticed. I have not tested the function. But looks suspicious to me.
```
last_slash = -1
x = 0
i = 0
buffer[0] = s[0 + (-1)] -> undefined behaivor
```

huitema:
@hfstco do you actually approve this PR?

hfstco (approved):

suhasHere (approved):

p_sim/pico_sim.c
huitema: Good catch. Fixed that, init should be `last_slash=0`. It is correctly set to (index of last / + 1) after finding a separator. Renamed to after_slash for clarity.

PR #2107, opened 2026-05-28
Install picosplay.h with HTTP headers

rwl4:
`h3zero_common.h` is installed as part of `PICOHTTP_HEADERS`, but it tries to include `picosplay.h`, which was only listed as a private core header.

This adds `picosplay.h` to the installed HTTP header set so downstream consumers can include installed HTTP/WebTransport headers such as `pico_webtransport.h` without needing the picoquic source tree in their include path.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -263,6 +263,7 @@ set(PICOHTTP_LIBRARY_FILES
     picohttp/wt_baton.c)
 
 set(PICOHTTP_HEADERS
+     picoquic/picosplay.h
      picohttp/h3zero.h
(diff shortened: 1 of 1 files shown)


huitema (approved):
Thanks for fixing this!

rwl4:
Hey, if you could get this merged somewhat soonish I'd really appreciate it. I have a library releasing soon that uses picoquic's WebTransport, and it'd be nice to not have to point at my fork.

huitema:
Sorry, I had approved it, I should have merged it immediately!

PR #2108, opened 2026-05-30
Faster find ready tream

huitema:
The function `picoquic_find_ready_stream_path` is shown to use 11% of the available CPU when managing 4 incoming video streams per connection at a media over QUIC relay managing 250 connections. This seems excessive. The root cause seems to be the requirement to check all the streams that share a priority level before selecting the one with the oldest (lower) time of last data sent. This makes the algorithm O(N). It would be nice to drive that to O(1).

That function is used for every packet, which means that any change carries a high risk of regression. The first series of fixes prepare the resolution:

* Move the stream related code to a separate source file, "stream.c", to facilitate reasoning on stream scheduling functions.
* Decouple sending "stop sending" and "reset frame" from the `picoquic_find_ready_stream_path`, which helps streamline the code in that function.
* Make the output list dynamic, instead of reserving a fixed size.
* Only add stream in the output list if they have data to send, which reduces the number of streams to check for every invocation.
* Fix the handling of the legacy "high priority stream".

These are just initial steps. More changes will come as we finalize this PR.


huitema:
The commit [call find stream ready once](https://github.com/private-octopus/picoquic/pull/2108/commits/61d5fabaa72f8a6b30149e8abe901adcc5a7be42) is fixing a simple problem: the code was calling the "find ready stream" function once to find the first stream to process, then once more before formatting the stream data frames. This is an easy an safe fix which ought to improve performance.

For some unknown reason that fix triggered a latent "memory leak" bug in the testing of WT-Baton. To identify that issue, we had to run ASAN/UBSAN, which detected three other issues. Those were fixed outside of the PR, as they are independent, but had to be merged into the PR to avoid merging issues later, and also to not have spurious failures of the CI tests.

huitema:
There is a small issue with the optimization to only keep "ready" stream in the output list. If we also kept in that list the flows that are blocked by flow control, we could set the "blocked by flow control" flag in the memory log report each time the blocked frame would have been at the head of the list. But that's generally inefficient, because that creates extra tests for each packet until the flow control condition is resolved. Instead, we only set the flag when we try to add the stream to the output list and find it blocked.

huitema:
@suhasHere This PR reworks the implementation of "find_ready_stream_path" to make it less sensitive to the number of available streams. The key is to trim and sort the "output_streams" list:

* Only include it it the streams that can send data, thus reducing the size of the list
* Maintain the list in order of priorities
* As the list is ordered, return the first available stream (normally the first in the list) instead of visiting all streams.

This combination of improvements and a little bit of code rewrite drives down the cost of the "find_ready_stream_path" function -- to the point that it does not show in the list of hot functions in may tests.

A small issue remains: my tests do not exactly reproduce your use case. At that point, we have two choices:

* redo the measurements by running your tests against this code branch, and learning potential issues,
* or, merge this PR as is, redo the measurements later against the master branch, and fix later whatever issue appears.

Please review and decide!

TimEvens:
This does significantly reduce find stream ready CPU usage.

picoquic/streams.c
```diff
@@ -664,6 +730,57 @@ picoquic_stream_head_t* picoquic_find_stream(picoquic_cnx_t* cnx, uint64_t strea
     return (picoquic_stream_head_t*)picosplay_find(&cnx->stream_tree, (void*)&target);
 }
 
+int picoquic_find_ready_stream_has_data(picoquic_cnx_t* cnx, picoquic_stream_head_t* stream)
+{
+    int has_data = 0;
+    /* TODO: the condition cnx->maxdata_remote > cnx->data_sent
+     * applies to all streams and should be moved to the top of the loop. */
```
suhasHere: +1, should we do this in this PR?

huitema: I have updated the comment. I checked the other comment in that function, and updated it too.

I had to do some extra work to resolve a merge conflict on a parallel PR. Can you review again? I would like to merge this and avoid further conflicts!

suhasHere (approved):
Looks good to me.

PR #2113, opened 2026-06-01
Per-stream retirement callback (stream_released) + h3zero forwarding

gmarzot:
Apps with per-stream context have no way to know when picoquic has fully retired a stream. `picoquic_unlink_app_stream_ctx` only nulls the pointer; it doesn't tell the app *when*. `picoquic_is_stream_acked` is internal to frames.c. Best the app can do today is bulk-free at `picoquic_callback_close` — accumulating state for the full connection lifetime, pathological for workloads with high mid-session stream churn.

Two commits:

**1. Add `picoquic_callback_stream_released` + h3zero forwarding**

Fires from `picoquic_delete_stream_if_closed` just before the stream is deleted, when `app_stream_ctx` is non-NULL. Pure receivers (FIN/RESET received), senders, and bidi all get notified at the right moment — when picoquic is confident no further callback will reference this `stream_ctx`. Callback return value is ignored: the stream is about to be deleted regardless.

`h3zero_callback` forwards the new event as `picohttp_callback_free` to the registered path callback, then deletes its own `stream_ctx` (mirrors the connection-close `picosplay_empty_tree` path). Without this, apps registered via path callbacks can only release per-stream state at session close.

**2. `picoquic_mark_active_stream`: NULL ctx means "do not change"**

Layered stacks hit a clobber path: h3zero (or picowt) sets `app_stream_ctx` at stream-create time, then a higher layer that uses prepare_to_send calls `picoquic_mark_active_stream(.., NULL)` to wake the worker (no ctx opinion). `mark_active_stream`'s unconditional assignment wipes h3zero's pointer. The new `stream_released` callback then fires with `app_stream_ctx == NULL` and the layered stack never learns about the retirement.

Treating NULL as "do not change" fixes the clobber. `picoquic_unlink_app_stream_ctx` remains the explicit-clear API. Documented behaviour of `mark_active_stream` (picoquic.h:1373) does not specify NULL semantics, so this is not a documented-behaviour change.

Does NOT modify `picoquic_add_to_stream_with_ctx` (same clobber pattern). Its NULL-via-wrapper `picoquic_add_to_stream` IS documented at picoquic.h:1500 to "erase the app_stream_ctx value" — that one stays.

**Note for quicperf:** `picohttp/quicperf.c:1000` calls `mark_active_stream(.., 0, NULL)` as an explicit deactivate-and-clear-ctx before delete. After this change `app_stream_ctx` is preserved, leaving a dangling pointer to the about-to-be-freed `quicperf_stream_ctx_t`. quicperf should add a `picoquic_unlink_app_stream_ctx` alongside. Not bundled here so the policy change stays single-purpose; happy to add it to this PR on request.

**Alternate approach:** keep `mark_active_stream` semantics, push the lookup burden to apps — they call `h3zero_find_stream` before `mark_active` and pass the resulting `stream_ctx`. Per-callback O(log N) splay-tree cost; ~0.05% CPU on a 2.5 Gbps WT workload, so negligible. Doable but every layered-stack app would have to repeat the same workaround.

**Tested:** `picoquic_ct` 515/515 + `picohttp_ct` 82/82 pass on this branch.

### DIFF
--- a/picoquic/streams.c
+++ b/picoquic/streams.c
@@ -246,6 +246,36 @@ int picoquic_mark_active_stream(picoquic_cnx_t* cnx,
     return ret;
 }
 
+int picoquic_mark_active_stream_v2(picoquic_cnx_t* cnx,
+    uint64_t stream_id, int is_active)
+{
+    int ret = 0;
+    picoquic_stream_head_t* stream;
+    PICOQUIC_THREAD_CHECK(cnx->quic);
+
+    stream = picoquic_find_stream_for_writing(cnx, stream_id, &ret);
+    if (ret == 0) {
+        if (is_active) {
+            if (!stream->fin_requested &&
+                (!stream->reset_requested || picoquic_check_sack_list(&stream->sack_list, 0, stream->reliable_size) == 0) &&
+                cnx->callback_fn != NULL) {
+                if (!stream->is_active) {
+                    stream->is_active = 1;
+                    picoquic_reinsert_by_wake_time(cnx->quic, cnx, picoquic_get_quic_time(cnx->quic));
+                }
+            }
+            else {
+                ret = PICOQUIC_ERROR_CANNOT_SET_ACTIVE_STREAM;
+            }
+        }
+        else {
+            stream->is_active = 0;
+        }
+    }
+
+    return ret;
+}
+
 int picoquic_set_stream_not_coalesced(picoquic_cnx_t* cnx, uint64_t stream_id, int is_not_coalesced)
 {
     int ret = 0;
--- a/picohttp/h3zero_common.c
+++ b/picohttp/h3zero_common.c
@@ -1967,6 +1967,25 @@ int h3zero_callback(picoquic_cnx_t* cnx,
 			* after the limits for unidir streams are known */
 			ret = h3zero_protocol_init_safe(cnx, ctx);
 			break;
+		case picoquic_callback_stream_released:
+			/* picoquic patch 0003 fires this per-stream when fully
+			 * retired. Forward to the registered path callback as
+			 * picohttp_callback_free so the app can release per-
+			 * stream context mid-session, then delete h3zero's own
+			 * stream_ctx. Without this, apps can only clean at
+			 * session close (picosplay_empty_tree). */
+			if (stream_ctx == NULL) {
+				stream_ctx = h3zero_find_stream(ctx, stream_id);
+			}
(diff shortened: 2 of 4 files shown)


huitema:
I like the idea of `picoquic_callback_stream_released`, that's a nice improvement, but can we do the change without breaking compatibility with existing applications?

You want to change the semantic of `picoquic_mark_active_stream: NULL ctx` to mean "do not change". I would rather add an alternate version, say `picoquic_mark_active_stream_v2` which simply does not have context parameter, making it obvious to the app that the context will not change.

There have been tons of discussions about the semantics of acknowledgements and how to surface it at the API. The general conclusion is to not do that, because it is misleading. Especially in the webtransport context, we can have implementations that interpose a queue before transport and application. ACK means "it was placed in the queue", not "it was delivered to the application". Picoquic does not have such a queue, but you cannot predict that the peer is using picoquic. If we provided the API, we would get issues when the remote application sends an ACK, and then later clears the queue between transport and app.

gmarzot:
Thanks @huitema — agreed on the back-compat point. Branch updated: dropped the NULL-semantic change to picoquic_mark_active_stream; added picoquic_mark_active_stream_v2 (no ctx parameter) as you suggested. v1 unchanged.

picoquic/streams.c
```diff
@@ -246,6 +246,36 @@ int picoquic_mark_active_stream(picoquic_cnx_t* cnx,
...
+            }
+        }
+        else {
+            stream->is_active = 0;
+        }
+    }
+
+    return ret;
+}
+
```
huitema, line 278: Just on small issue, motivated by code coverage and also robustness to merge conflicts -- we have another PR in progress that will modify `picoquic_mark_active_stream`, and by default the merge would miss the existence of "_v2". I would split `picoquic_mark_active_stream` and `picoquic_mark_active_stream` in three components:

- a common core, static, with a stream context as argument, that does the mark active and the wake time, and return the stream pointer or NULL.
- `picoquic_mark_active_stream` that calls that core and update the `app_stream_ctx` if the returned stream pointer is not NULL,
- `picoquic_mark_active_stream_v2` that calls the core function and set the return code if the returned pointer is NULL.

huitema: Or, we could just merge as is, and I will handle the merge conflict later!

huitema (approved):
Nice improvement!

There will be a merge conflict with Pr #2108. I will handle that conflict when merging the other PR.

PR #2114, opened 2026-06-03
h3zero: bound partial frame buffering and QPACK indexes

victorstewart:
this https://blog.calif.io/p/codex-discovered-a-hidden-http2-bomb inspired the below bug finds: h3zero parser bugs where tiny attacker-controlled inputs could either pin disproportionate memory or read past the QPACK static table.

Baseline repro evidence was collected before the production fixes. The unfixed code accepted or triggered all of the following repros:

* HEADERS resource pin: request-stream bytes `01 80 01 00 00 00` declare a 0x10000-byte HEADERS frame and provide one payload byte. The baseline test observed a 65536-byte `current_frame` allocation with `current_frame_read == 1`, pinning 64 KiB from six wire bytes per open stream.

* SETTINGS resource pin: peer unidirectional bytes `00 04 80 01 00 00 00` declare a 0x10000-byte SETTINGS frame and provide one payload byte. The baseline test observed a 65536-byte `current_frame` allocation for one payload byte on the control stream path.

* Duplicate control stream: a second peer-initiated unidirectional control stream was accepted with error 0 before SETTINGS was received, despite HTTP/3 allowing only one peer control stream.

* QPACK indexed-static OOB: payload `00 00 ff 24` uses static index 99. The baseline debug test accepted it, and ASan/UBSan reported a global-buffer-overflow at the static-table lookup.

* QPACK static-name-reference OOB: payload `00 00 5f 54 00` uses static name index 99 with an empty value. The baseline debug test accepted it, and ASan/UBSan reported a global-buffer-overflow at the static-table lookup.

The resource-pinning issue is bounded by stream limits, but it is asymmetric: a small amount of request/control-stream data reserved the maximum 64 KiB frame buffer and kept it attached until the stream completed or was torn down. The QPACK issue is a classic off-by-one bounds check: `h3zero_qpack_nb_static` is the entry count, so index `h3zero_qpack_nb_static` must be rejected instead of read.

Fixes:

* Change both QPACK static table checks from `>` to `>=`.

* Add grow-on-receive buffering for processed HEADERS and SETTINGS/control frames. The parser now reserves only the bytes needed for content received so far, grows geometrically up to the declared frame length, and still enforces the existing 0x10000 processed-frame cap.

* Track aggregate buffered partial-frame bytes in the h3zero callback context and reject excess buffered load with `H3ZERO_EXCESSIVE_LOAD`.

* Release frame-buffer accounting through a shared cleanup helper so stream teardown and control-stream reset cannot leak the aggregate budget.

* Enforce a single remote control stream by recording the first peer control stream id and rejecting any distinct second control stream with `H3ZERO_STREAM_CREATION_ERROR` before frame parsing/allocation.

* Use `H3ZERO_EXCESSIVE_LOAD` for oversized HEADERS/processed-frame load failures instead of treating peer-driven load as an internal error.

* Preserve empty-frame behavior while avoiding zero-length copies against a NULL backing buffer.

Regression coverage:

* Added focused tests for the partial HEADERS pin, partial SETTINGS pin, duplicate remote control stream rejection, indexed-static index 99, static-name-reference index 99, and the combined QPACK static-bounds wrapper.

* The partial-frame tests assert that the final fixed path does not allocate the full declared 0x10000-byte frame after receiving only one payload byte.

* The SETTINGS test also verifies aggregate buffered-byte accounting is released when the stream state is deleted.

Proof and final verification:

* Baseline proof logs were recorded under `.experiments/artifacts/` before fixes and showed the two 65536-byte pins, duplicate control-stream acceptance, and ASan/UBSan QPACK global-buffer-overflows.

* Final focused Debug repros passed: partial HEADERS, partial SETTINGS, duplicate remote control stream, combined QPACK bounds, indexed-static index 99, and static-name-reference index 99.

* Final ASan/UBSan focused repros passed for both QPACK crash inputs, QPACK parse/fuzz coverage, partial HEADERS, partial SETTINGS, and duplicate remote control stream rejection.

* Final full-suite gates passed: `picohttp_ct` 93/93, `picoquic_ct` 538/538, `ctest` 2/2, and `git diff --check` clean.


huitema:
Wow. That's a lot of code. Please give me at least a day to properly review!

huitema:
My first request would be to remove the "grow-on-receive buffering" code. That's a lot of code, and I don't think that we need it. We may want to use a smaller limit for the max header length.

huitema:
For the setting frames, it would be easy to replace the current "global load and parse" by a progressive parser that loads the individual TLV components, without allocating memory.

victorstewart:
superseeded

PR #2115, opened 2026-06-04
Proof of concept for ACK rate reduction

huitema:
This PR is just a proof of concept, showing that higher min and max ack delay can drive lower ACK rates. Run this PR if you want to evaluate the performance effect on an application. It does cause a number of regressions that the test suite finds, and is not in any way ready for merging in the main branch.


suhasHere:
Add we planning to close this PR ?

huitema:
Yes, we should close this PR. Way too many regressions to proceed with that.

PR #2116, opened 2026-06-04
WebTransport: draft refresh, conformance suite, browser gates, and CI

victorstewart:
## Summary

Minimal WebTransport-over-HTTP/3 draft refresh for picoquic:

- Advertise and parse current WebTransport SETTINGS needed for draft-15 interop.
- Use `webtransport-h3` for current peers and preserve compatibility with older `webtransport` peers.
- Defer client CONNECT until peer SETTINGS and transport requirements are known.
- Fix DRAIN capsule send/receive behavior and zero-length capsule accumulation.
- Keep WebTransport session-level flow control out of picoquic; QUIC flow control remains the resource bound.
- Split `wt_baton` out of `picohttp-core` so demo/test code is not compiled into the core HTTP library.
- Add focused native tests and documentation for the supported compatibility policy.

This intentionally does not add WebTransport session flow-control SETTINGS/capsules or in-tree browser E2E infrastructure.


huitema (dismissed):
I like some of this PR. Updating to the latest code points, fixing interop bugs, adding documentation, all that is great. But this way too much code for a single PR, which makes it essentially impossible to review. I would much prefer a set of much smaller PRs, that could be reviewed individually. Also, remember the design philosophy of picoquic: keep it simple, keep it small!

There is also a ton of code to do interop testing with web browsers. That's a great idea, but it feels a lot like a project within the project, adding a lot of code. I don't know whether that should be separated.

A lot of the new code comes from the implementation of web transport flow control. Yes, it is not done in the current builds, but this is deliberate. Flow control at the web transport layer duplicates the flow control at the Quic layer and introduces a lot of complexity. The picoquic design choice is to not use that and stick to one webtransport connection per Quic connection. Please remove the web transport flow control code from this PR. We should only consider the addition of such features after opening an issue, explaining the motivation, and getting consensus.

So, two requirements: remove all the code tied to web transport flow control, including all the flow control capsules and the related negotiations and tests; and, align with the latest picoquic and ensure that the CI tests run.

picoquictest/h3zero_stream_test.c
```diff
@@ -554,6 +555,22 @@ uint8_t* h3zero_client_data_get_response(uint8_t * bytes, uint8_t * bytes_max)
...
+        }
+        else {
+            *bytes++ = 0;
+            *bytes++ = 0;
+            *bytes++ = 0xC0 | 7; /* etag static header, no pseudo header in trailers. */
+        }
+    }
+    return bytes;
+}
+
```
huitema, line 573: 1) Why do we need this function?

2) The name "data_get" is a bit misleading, as "get" generally means fetching something out of some context. For functions that simply prepare a message we should use a name like "xxx_format_yyy".

picoquic/tls_api.c
```diff
@@ -2390,7 +2391,21 @@ int picoquic_export_secret(picoquic_cnx_t *cnx, const char *label, uint8_t *out,
     }
 
     ptls_t* tls = tls_ctx->tls;
-    return ptls_export_secret(tls, out, outlen, label, ptls_iovec_init(NULL, 0), 0);
+    return ptls_export_secret(tls, out, outlen, label, context, 0);
+}
+
```
huitema: Curious: why do we need this?

picoquic/quicctx.c
```diff
@@ -5206,6 +5206,17 @@ uint64_t picoquic_get_remote_stream_error(picoquic_cnx_t* cnx, uint64_t stream_i
...
+{
+    uint64_t remote_stop_error = 0;
+    picoquic_stream_head_t* stream;
+    PICOQUIC_THREAD_CHECK(cnx->quic);
+    if ((stream = picoquic_find_stream(cnx, stream_id)) != NULL) {
+        remote_stop_error = stream->remote_stop_error;
+    }
+    return remote_stop_error;
+}
+
```
huitema: Not sure about that one. The stream is going to be deleted shortly after the error is noticed. Are we setting ourselves for a race condition?
Maybe add a comment that this function should only be used within the callback announcing the error?

picoquic/picoquic.h
```diff
@@ -743,6 +743,10 @@ void picoquic_set_use_exporter(picoquic_quic_t* quic, int use_exporter);
 /* Export keying material from the TLS connection using the given exporter label */
 int picoquic_export_secret(picoquic_cnx_t *cnx, const char *label, uint8_t *out, size_t outlen);
 
+/* Export keying material from the TLS connection using the given exporter label and context */
+int picoquic_export_secret_with_context(picoquic_cnx_t* cnx, const char* label,
+    const uint8_t* context, size_t context_len, uint8_t* out, size_t outlen);
+
```
huitema: Curious: where are we using this exported secret?

picohttp/h3zero_common.h
```diff
@@ -48,6 +48,7 @@ extern "C" {
         picohttp_callback_provide_datagram, /* Ready to send datagram in this context */
         picohttp_callback_reset, /* Stream has been abandoned by peer. */
         picohttp_callback_stop_sending, /* Peer asking to reset the stream. */
+        picohttp_callback_drain, /* Peer initiated graceful WebTransport drain. */
```
huitema: This is a change in a published API. It is probably OK, but we will have implementations that have not implemented drain. If they do not ignore the call, they are going to break. How can we mitigate that?

CMakeLists.txt
```diff
@@ -680,7 +680,9 @@ if(BUILD_TESTING AND picoquic_BUILD_TESTS)
     target_include_directories(picohttp_ct PRIVATE picohttp)
     set_picoquic_compile_settings(picohttp_ct)
 
-    add_executable(pico_baton baton_app/baton_app.c)
+    add_executable(pico_baton
+        baton_app/baton_app.c
+        picohttp/wt_baton.c)
```
huitema: We are compiling wt_baton.c a couple times. Should it be part of a library?

huitema (commented):
Yes, that's a vast improvement from the previous version.

I am concerned that we are introducing changes in published APIs. For example, we do require that the web transport "connect" be issued after the HTTP connection is ready, while it was OK to queue before that in the previous version. We also add a new Drain callback, which may trip some implementations. We do have existing developments that use Picoquic and Webtransport. They may be surprised by the change. we need to find a way to mitigate that. For example, it might be a good idea to rev the version number.to "1.1.50.0' (in both picoquic.h and CMakeList.txt.

On the other hand, these existing implementations will directly benefit from the improved interoperability provided by the PR.

I will ask a couple of key developers to review the PR before we merge.

See the comments in the text.

huitema:
@gmarzot @suhasHere @lminiero This PR makes a series of changes to ensure interop between webtransport in picoquic and the implementations in Chrome, Firefox and Safari. That's very useful, but I am concerned that the interop changes may affect your implementations. Please take a look and comment if that's needed!

victorstewart:
i'll just maintain the same connect dynamics while still adhering to the latest draft spec by deferring the CONNECT bytes internally until after the peer SETTINGS and WebTransport transport requirements are validated

victorstewart:
every concern should be addressed now

picoquictest/h3zero_stream_test.c
victorstewart: it’s test input generation: it formats a minimal valid trailer HEADERS frame so h3zero_client_data_test covers HEADERS -> DATA -> trailer and rejects DATA after a trailer. agree on naming; I made the trailer helper static and renamed it to h3zero_client_data_format_trailer(). i also renamed the response helper from get to format for consistency

huitema:
@gmarzot @suhasHere @lminiero  -- My previous comment expressed concern about possible changes in the public API of web transport, and how they would affect your implementations. @victorstewart has revised the PR, and the existing code will continue to work -- which is great. However, this PR moves picoquic to the latest web transport draft (draft 15), which is incompatible with the old code that supported interop with draft 7 to 14. Going to draft 15 has lots of benefits, including verified interoperability with the latest version of Chrome, which is great. Would it be easy to upgrade your implementation and your deployments to that new version?

lminiero:
@huitema actually I haven't started using the picoquic native WebTransport stack yet: in my library, I'm still using my own WebTransport layer on top of picoquic's QUIC implementation. I have it in my plans to switch, but I haven't found the time yet. As such, there definitely isn't anything this PR can break for me at this stage :smile:

gmarzot:
> @gmarzot Would it be easy to upgrade your implementation and your deployments to that new version?

@huitema Thanks for mentioning it. I have not looked closely enough to know what migrating will look like. But nice that there is nothing breaking all of a sudden. My instinct would be to try to support the older WT drafts as well as the new draft 15 (already in that muck w/ MOQ).  I just got CC/FC to behave reasonably for WT (h3zero based).  I wonder how this PR might change things in that area.. and if I update to a version w/ this PR i will lose older draft compatibility even if the APIs stay working? I can just pin my version until I figure it out.

huitema:
I looked at the changes between drafts. The key elements are:

|                    | draft-15   | draft-14    | Draft-12    | Draft 11  | Draft 10
|--------------|------------|-------------|------------|----------|---------
| Negotiation   | WT_ENABLED  | WT_MAX_SESSIONS | WEBTRANSPORT_MAX_SESSIONS | same
| Nego error   | WT_REQUIREMENTS_NOT_MET | nothing)  | nothing | nothing
| Connect protocol| "webtransport-h3" | "webtransport" | same | same
| Close session   | capsule 0x2843    | capsule 0x2843 | same | same
| Drain session   | capsule 0x78ae    | capsule 0x78ae | same | same
| Flow control    | Yes.              |  Yes           | Yes   | No


Drafts 13-14 use same codes.
Drafts 7-8-9-10-11 use the same codes

So, it boils down to adding support for 3 different settings. The client should offer all three settings, probably in the order from draft 15 to draft 11. The server should read the client settings, and should negotiate the most up to date version offered by the client. Or vice versa.

If they negotiate draft 15, they should set the "protocol" in the extended connect to "webtransport-h3". Otherwise, just use "webtransport". Treat those two values as equivalent when processing the request.

I would not change anything else. At worst, the server will send the  error WT_REQUIREMENTS_NOT_MET that the client does not understand. So what...

picohttp/h3zero_common.c
```diff
@@ -280,7 +293,7 @@ int h3zero_protocol_init(picoquic_cnx_t* cnx)
 	 */
 	if (cnx->local_parameters.max_datagram_frame_size > 0) {
 		settings.h3_datagram = 1;
-		settings.webtransport_max_sessions = 1;
+		settings.webtransport_enabled = 1;
 	}
```
huitema, line 299: To support negotiating down to old versions, we should probably set both `settings.webtransport_max_sessions` and `settings.webtransport_enabled` to 1.

picohttp/h3zero_common.c
```diff
@@ -328,6 +341,58 @@ int h3zero_protocol_init_safe(picoquic_cnx_t* cnx, h3zero_callback_ctx_t* ctx)
...
+	return ctx != NULL &&
+		ctx->settings.settings_received &&
+		ctx->settings.enable_connect_protocol &&
+		ctx->settings.h3_datagram &&
+		ctx->settings.webtransport_enabled > 0 &&
+		remote_tp != NULL &&
+		remote_tp->max_datagram_frame_size > 0 &&
+		remote_tp->is_reset_stream_at_enabled;
+}
+
```
huitema: To support negotiation, we should probably replace `ctx->settings.enable_connect_protocol &&` by
`(ctx->settings.enable_connect_protocol || (ctx->settings.webtransport_max_sessions > 0) &&`

doc/pico_webtransport.md
```diff
@@ -67,8 +77,12 @@ requirements are:
...
  * Configure the HTTP3 server to accept and process Web Transport requests by
    configuring a "path table". Each path specifies a local URL, a path callback
-   function, and a path callback context, per `picohttp_server_path_item_t`
-   in `h3zero_common.h`.
+   function, a path callback context, and the required Extended CONNECT
+   protocol `webtransport-h3`, per `picohttp_server_path_item_t` in
+   `h3zero_common.h`. WebTransport paths also need an Origin policy callback.
+   The demo baton app uses `h3zero_origin_validator_allow_all` as an explicit
+   test/demo policy; production applications should replace it with an
+   allowlist validator.
```
huitema, line 86: Probably need to add a note about supporting backward compatibility, max_sessions, and setting the protocol to the old value, `webtransport`

picohttp/h3zero.h
```diff
@@ -50,9 +50,13 @@ extern "C" {
...
 #define H3ZERO_WEBTRANSPORT_BUFFERED_STREAM_REJECTED 0x3994bd84 /* Stream arrived before webtransport session established */
 #define H3ZERO_WEBTRANSPORT_SESSION_GONE 0x170d7b68 /* Stream arrived after web transport session closed */
-#define H3ZERO_WEBTRANSPORT_APPLICATION_ERROR(code) (0x52e4a40fa8dbull + code) /* see spec for skipping grease points when mapping codes */
+#define H3ZERO_WEBTRANSPORT_APPLICATION_ERROR_FIRST 0x52e4a40fa8dbull
+#define H3ZERO_WEBTRANSPORT_APPLICATION_ERROR_LAST 0x52e5ac983162ull
+#define H3ZERO_WEBTRANSPORT_APPLICATION_ERROR(code) (H3ZERO_WEBTRANSPORT_APPLICATION_ERROR_FIRST + (uint64_t)(code) + ((uint64_t)(code) / 0x1eull))
 #define H3ZERO_USER_AGENT_STRING "H3Zero/1.0"
 
+#define H3ZERO_WEBTRANSPORT_H3_PROTOCOL "webtransport-h3"
+
```
huitema, line 60: Add:
```
#define H3ZERO_WEBTRANSPORT_H3_PROTOCOL_OLD "webtransport"
```

picohttp/webtransport.c
```diff
@@ -337,6 +337,63 @@ const char* picowt_get_authority(h3zero_stream_ctx_t* stream_ctx)
...
+    uint8_t* bytes = stream_ctx->frame;
+    uint8_t* bytes_max = stream_ctx->frame + sizeof(stream_ctx->frame);
+
+    *bytes++ = h3zero_frame_header;
+    bytes += 2; /* reserve two bytes for frame length */
+
+    bytes = h3zero_create_connect_header_frame(bytes, bytes_max, authority,
+        (const uint8_t*)path, strlen(path), "webtransport", NULL,
+        H3ZERO_USER_AGENT_STRING, wt_available_protocols);
+
```
huitema, line 448: "webtransport" seems wrong here. There should be a test of the settings. If the peer has negotiated 
`ctx->settings.webtransport_enabled`, then use H3ZERO_WEBTRANSPORT_H3_PROTOCOL (the definition of "webtransport-h3), else use H3ZERO_WEBTRANSPORT_H3_PROTOCOL_OLD.

huitema (commented):
Suggestions inline for being able to negotiate down. 

I did not find the code where we parse the incoming settings. We probably need to add or restore the parsing of the old `max_session` setting. we will end up with the conditions:

* peer has sent `enable_webtransport`: life is good. Use argument `webtransport-h3` in the connect request.
* peer has not sent `enable_webtransport` but has sent `max_session`: attempt downgrade interop. Use argument `webtransport` in the connect request.

Should not be too hard.

picohttp/h3zero_common.c
```diff
@@ -54,6 +54,115 @@ int h3zero_origin_validator_allow_all(
...
+			else {
+				memcpy(stream_ctx->frame + *connect_length, extra, extra_length);
+				*connect_length += extra_length;
+			}
+		}
+	}
+
+	return ret;
+}
+
```
huitema: Why do we need that function? We already have `h3zero_create_connect_header_frame` which does exactly that!

picohttp/h3zero_common.c
huitema: OK, done.

huitema: OK.

picohttp/h3zero_common.c
```diff
@@ -54,6 +54,115 @@ int h3zero_origin_validator_allow_all(
...
+	}
+	else if (h3zero_path_item_uses_webtransport_protocol(item)) {
+		return h3zero_is_webtransport_connect_protocol(header->protocol, header->protocol_length);
+	}
+	else {
+		return header->protocol != NULL &&
+			header->protocol_length == item->connect_protocol_length &&
+			memcmp(header->protocol, item->connect_protocol, item->connect_protocol_length) == 0;
+	}
+}
```
huitema: I think a lot of these functions are already present in `webtransport.c`. we are getting a lot of duplication.

picohttp/h3zero_common.c
huitema: "connect" is an H3 verb, used by more than web transport. It is also used, for example, by the "masque" family of protocols. We want a generic formatting function in either h3zero.c or h3zero_common.c, and then simply using it with the proper parameters for web transport.

huitema (commented):
I have bad feelings about this PR. We only needed to tighten the negotiation of draft 15, but we end up with an extensive rewrite of the web transport support, including duplication of functions.

A lot of the protocol tightening is present already in Pr #2117, including formatting of the connect requests and parsing of incoming connect, as well as parsing of the various settings. the only real changes that we need are:

1- deferring the sending of the "connect" frame until the web transport version has been negotiated

2- having code to set the protocol to the draft-15 value (webtransport-h3) or to the legacy value.

I think we could do with a much shorter PR, with the constraint to change teh existing code as little as possible.

victorstewart:
i didn't mean to push that earlier. was some experimental branch. i rolled back the history then added what you requested. see what you think now.

victorstewart:
made a few more fixes:

- Replaced per-stream pending WebTransport CONNECT state and H3 stream-tree scan with one connection-level pending CONNECT.
- Made WT protocol fallback matching comma-aware, so supported lists like `foo,bar` can match `bar`.
- Made CONNECT frame formatting fail when `extra` would overflow instead of silently omitting bytes.
- Fixed `wt_baton_accept()` cleanup so allocated baton state is freed on init/parse/relay error paths.
- Enforced server-side CONNECT `:protocol`, parsed `Origin`, and applied configured Origin validation/error status before accepting.
- Fixed `wt_baton` WT-Protocol string termination at `wt_protocol_len`.

huitema (approved):
OK, my previous review comments have been addressed. I think we should call this done and merge it.

I am worried about compatibility with PR #2117. I hope we are not going to have too many merge conflicts. I think we should merge this PR into the master branch, then I will merge the resulting master into the PR #2117 branch and solve the conflicts before merging PR #2117.

I kinda regret having insisted on being able to queue a connect request before having completed the H3 connection. It does add a fair bit of complexity. We may want to revisit that in a following PR. Maybe add an HTTP callback for "settings received" so the client will know when starting the HTTP_CONNECT. Maybe add some restrictions on pre-setting transactions, such as using the preferred version and failing if the peer has a different version. Or maybe leaving it as is. We shall see.

But, now, time to merge this.

victorstewart:
okay i'm good to merge, and that merge strategy with #2117 sounds good. when you want to fix the connect event API i'm ready to.

PR #2117, opened 2026-06-05
H3zero robustness issues

huitema:
Addressing the H3Zero robustness issues outlined in PR #2114:

* QPACK indexed-static and static-name-reference accepting out of bounds index 99: same fix as PR 2114.
* Accepting a duplicate control stream: same fix as PR 2114.
* SETTINGS resource pin: instead of attempting to control memory as in PR 2114, just perform enforce a the same allocation limit for setting streams as for data streams.
* HEADERS resource pin: simply enforce an allocation limit of 64KB per stream. The number of streams is limited by the QUIC settings, which means the allocation of resource is bounded.

We could do better, by enforcing a limit on the number of streams "in progress" simultaneously, and only relaxing that limit when the stream header is fully . But it is very hard to find a limit that works for every application. If a limit like that is enforced, it should be by application, and the application itself is in a good position to do it.


victorstewart:
A) when a second control stream is detected the code is treating it as a stream level error instead of a connection level error, we need to close the connection instead of issuing STOP_SENDING

B) we need to set `ctx->settings.settings_received` for a zero-length SETTINGS frame, which is currently skipped in here https://github.com/private-octopus/picoquic/blob/af48c33cec73597ceac94d57546e99d9cde5596a/picohttp/h3zero_common.c#L458 because of the `stream_state->current_frame != NULL` check

C) we're currently advertising our own `SETTINGS_MAX_FIELD_SECTION_SIZE` but ignoring the peer's `h3zero_setting_max_field_section_size`... is that what we want to be doing?

D) we're limiting the encoded QPACK block size everywhere with checks like `if (stream_state->current_frame_length > H3ZERO_MAX_FIELD_SECTION_SIZE) then fail`, but we're supposed to be checking the uncompressed size https://www.rfc-editor.org/rfc/rfc9114.html#:~:text=The%20size%20of%20a%20field%20list%20is%20calculated%20based%20on%20the%20uncompressed%20size%20of%20fields

E) `h3zero_post_data_or_fin()` casts away the const of `stream_ctx->path_callback`. just highlighting this

suhasHere (approved):
This looks good with the checks

victorstewart (changes requested):
see my reply

huitema:
> A) when a second control stream is detected the code is treating it as a stream level error instead of a connection level error, we need to close the connection instead of issuing STOP_SENDING

Yes, I guess that we could close the connection. This is really an H3 level error, not a web transport error..

> 
> B) we need to set `ctx->settings.settings_received` for a zero-length SETTINGS frame, which is currently skipped in here
> 
> https://github.com/private-octopus/picoquic/blob/af48c33cec73597ceac94d57546e99d9cde5596a/picohttp/h3zero_common.c#L458
> because of the `stream_state->current_frame != NULL` check

OK. That would mean "just use the defaults". 

> C) we're currently advertising our own `SETTINGS_MAX_FIELD_SECTION_SIZE` but ignoring the peer's `h3zero_setting_max_field_section_size`... is that what we want to be doing?

Not sure. Picoquic is pretty conservative when it comes to sending headers, using at most a few hundred bytes. We could refuse the connection if the setting claims too small a size, say lower than 1KB.

> D) we're limiting the encoded QPACK block size everywhere with checks like `if (stream_state->current_frame_length > H3ZERO_MAX_FIELD_SECTION_SIZE) then fail`, but we're supposed to be checking the uncompressed size https://www.rfc-editor.org/rfc/rfc9114.html#:~:text=The%20size%20of%20a%20field%20list%20is%20calculated%20based%20on%20the%20uncompressed%20size%20of%20fields

Yes. But we are almost never actually decompressing these headers. So we have no idea of the total decompressed size. The test makes the assumption that compressed size is lower than uncompressed size, which seems fair. There might be some vicious corner cases, but we can probably wait until somebody yells at us...

> E) `h3zero_post_data_or_fin()` casts away the const of `stream_ctx->path_callback`. just highlighting this

That's an unfortunate aspect of using the same callback for receiving data and for asking to send. The solution would be to change the signature of the callback function to cast its "byte" parameter as const. That maybe a good idea, but I am concerned that it will break a lot of things...

victorstewart (changes requested):
A) after picoquic_close_ex() succeeds the function needs to set bytes = NULL or return so it doesn't fall through into the normal switch and continue parsing that duplicate control stream

some new possible bugs:

B) shouldn't this be checking stream_type not current_frame_type? 

https://github.com/private-octopus/picoquic/blob/e0003f2a48f02772a0c14ce284f6cac5897fb74c/picohttp/h3zero_common.c#L552 

C) we're silently ignoring all unknown settings but technically it's a spec violation to not be returning H3_SETTINGS_ERROR for any HTTP/2 settings that have no HTTP/3 equivalent so 0x02, 0x03, 0x04, and 0x05

https://www.rfc-editor.org/rfc/rfc9114.html#name-settings:~:text=Setting%20identifiers%20that%20were%20defined%20in%20%5BHTTP/2%5D%20where%20there%20is%20no%20corresponding%20HTTP/3%20setting%20have%20also%20been%20reserved%20(Section%2011.2.2).%20These%20reserved%20settings%20MUST%20NOT%20be%20sent%2C%20and%20their%20receipt%20MUST%20be%20treated%20as%20a%20connection%20error%20of%20type%20H3_SETTINGS_ERROR.

and the same goes for HTTP/2 frame types without HTTP/3 equivalents so PRIORITY, PING, WINDOW_UPDATE, and CONTINUATION

huitema:
> A) after picoquic_close_ex() succeeds the function needs to set bytes = NULL or return so it doesn't fall through into the normal switch and continue parsing that duplicate control stream

Yes indeed.

> some new possible bugs:
> 
> B) shouldn't this be checking stream_type not current_frame_type?
> 
> https://github.com/private-octopus/picoquic/blob/e0003f2a48f02772a0c14ce284f6cac5897fb74c/picohttp/h3zero_common.c#L552

I get that the code is a bit confusing because it choses to repurpose existing fields to parse a stream header, but this has been extensively tested, I would rather leave it alone.

> 
> C) we're silently ignoring all unknown settings but technically it's a spec violation to not be returning H3_SETTINGS_ERROR for any HTTP/2 settings that have no HTTP/3 equivalent so 0x02, 0x03, 0x04, and 0x05
> 
> https://www.rfc-editor.org/rfc/rfc9114.html#name-settings:~:text=Setting%20identifiers%20that%20were%20defined%20in%20%5BHTTP/2%5D%20where%20there%20is%20no%20corresponding%20HTTP/3%20setting%20have%20also%20been%20reserved%20(Section%2011.2.2).%20These%20reserved%20settings%20MUST%20NOT%20be%20sent%2C%20and%20their%20receipt%20MUST%20be%20treated%20as%20a%20connection%20error%20of%20type%20H3_SETTINGS_ERROR.

Yes. That's also issue #2126. Let's fix it.

> and the same goes for HTTP/2 frame types without HTTP/3 equivalents so PRIORITY, PING, WINDOW_UPDATE, and CONTINUATION

OK, adding frame unexpected errors for frame types 2, 6, 8 and 9 per section 11.2.1. of RFC 9114.

huitema:
This PR will fix issue #2126

victorstewart (changes requested):
A) i think we should match the error handling in `h3zero_parse_remote_unidir_stream` to that in `h3zero_parse_remote_bidir_stream`, and change this:

```
if (opt_cnx != NULL) {
   (void)picoquic_close_ex(opt_cnx, H3ZERO_STREAM_CREATION_ERROR, "Attempt to create second control stream");
   bytes = NULL;
}
else {
   *error_found = H3ZERO_STREAM_CREATION_ERROR;
   bytes = NULL;
}
```

to just this 

```
*error_found = H3ZERO_STREAM_CREATION_ERROR;
bytes = NULL;
```

and then it'll just get closed once and cleanly here https://github.com/private-octopus/picoquic/blob/03735aabc5456c9339c6f21319c83b29b149cfcf/picohttp/h3zero_common.c#L1015 ? 

B) 
we're not calling `picowt_process_pending_connect` for zero length settings

https://github.com/private-octopus/picoquic/blob/03735aabc5456c9339c6f21319c83b29b149cfcf/picohttp/h3zero_common.c#L479

victorstewart (approved):

PR #2119, opened 2026-06-07
Reduce CPU load of processing ACK of Frames and fix Ack or Stream Reset At frames

huitema:
This PR started as an effort to reduce the CPU cost of the function `picoquic_decode_ack_frame`, which is a hot spot in some scenarios. This app does that by picking a low hanging fruit, processing frame types lower than 64 as single bytes instead of parsing them as varint, which seems to reduce the code of ack of ack processing by a bit less than 10%.

The performance improvement are measured by running a simulation, with the node sending 4 2Mbps video streams and 4 64kbps audio streams. Testing that scenario exposed a bug in the simulator, which is easy enough to fix.

The work to reorganize ack-of-ack processing exposed an issue in the processing of acks of "RESET STREAM AT" frames. The function `picoquic_process_ack_of_reset_stream_at_frame` was not called. It was also not properly debugged. The debugging exposed an issue in tracking the closing of streams that are "reset at" -- the sender should only close them when all the promised "reliable size" bytes have been acked. We should also only consider a "RESET STREAM AT" frame acknowledged if the ACK acknowledge the frame with the last "reliable size" values that was sent.


victorstewart:
A) there's an off by 1 error in all of these `(stream->reliable_size == 0 || picoquic_check_sack_list(&stream->sack_list, 0, stream->reliable_size) != 0)` SACK checks

we should make it a function and do it like:

```
static int picoquic_reliable_prefix_is_acked(picoquic_stream_head_t* stream)
{
   return stream->reliable_size == 0 || picoquic_check_sack_list(&stream->sack_list, 0, stream->reliable_size - 1) != 0;
}
```

B) isn't there a bug here https://github.com/private-octopus/picoquic/blob/1ce80e61983ce4465681d7f66c91ce097d7a3af4/picoquic/frames.c#L545 we’re gating deletion on `stream->consumed_offset >= stream->reliable_size` but the relevant condition is whether the peer ACKed our reliable prefix not how much incoming data the app consumed?

C) we added the reliable-prefix ACK requirement only for bidirectional streams but not for local unidirectional streams, shouldn't we add it for those also?

D) in here https://github.com/private-octopus/picoquic/blob/1ce80e61983ce4465681d7f66c91ce097d7a3af4/picoquic/frames.c#L3295 and in here https://github.com/private-octopus/picoquic/blob/1ce80e61983ce4465681d7f66c91ce097d7a3af4/picoquic/frames.c#L2160 shouldn't these be `offset >= stream->reliable_size`?

E) in https://github.com/private-octopus/picoquic/blob/1ce80e61983ce4465681d7f66c91ce097d7a3af4/picoquic/frames.c#L553 shouldn't we use the `reliable_size` value we decode to suppress stale reset-at frames whose reliable size no longer matches `stream->reliable_size`?

F) shouldn't this https://github.com/private-octopus/picoquic/blob/1ce80e61983ce4465681d7f66c91ce097d7a3af4/picoquic/frames.c#L422 be skipping the first / frame type byte when decoding the stream id?

huitema:
> A) there's an off by 1 error in all of these `(stream->reliable_size == 0 || picoquic_check_sack_list(&stream->sack_list, 0, stream->reliable_size) != 0)` SACK checks
> 
> we should make it a function and do it like:
> 
> ```
> static int picoquic_reliable_prefix_is_acked(picoquic_stream_head_t* stream)
> {
>    return stream->reliable_size == 0 || picoquic_check_sack_list(&stream->sack_list, 0, stream->reliable_size - 1) != 0;
> }
> ```

Yes! Good catch. Fixing.

> B) isn't there a bug here
> 
> https://github.com/private-octopus/picoquic/blob/1ce80e61983ce4465681d7f66c91ce097d7a3af4/picoquic/frames.c#L545
> we’re gating deletion on `stream->consumed_offset >= stream->reliable_size` but the relevant condition is whether the peer ACKed our reliable prefix not how much incoming data the app consumed?

No. This is testing whether more bytes need to be sent. Repetition of the lost packets, etc., is handled elsewhere. 

> C) we added the reliable-prefix ACK requirement only for bidirectional streams but not for local unidirectional streams, shouldn't we add it for those also?

Yes! Missed that one! Fixing.

> D) in here
> 
> https://github.com/private-octopus/picoquic/blob/1ce80e61983ce4465681d7f66c91ce097d7a3af4/picoquic/frames.c#L3295
> and in here
> 
> https://github.com/private-octopus/picoquic/blob/1ce80e61983ce4465681d7f66c91ce097d7a3af4/picoquic/frames.c#L2160
> shouldn't these be `offset >= stream->reliable_size`?

No. `offset > X` means bytes `[ 0, X-1 ]` have been sent.

> E) in
> 
> https://github.com/private-octopus/picoquic/blob/1ce80e61983ce4465681d7f66c91ce097d7a3af4/picoquic/frames.c#L553
> shouldn't we use the `reliable_size` value we decode to suppress stale reset-at frames whose reliable size no longer matches `stream->reliable_size`?

Yes, absolutely. Fixing.

> F) shouldn't this
> 
> https://github.com/private-octopus/picoquic/blob/1ce80e61983ce4465681d7f66c91ce097d7a3af4/picoquic/frames.c#L422
> be skipping the first / frame type byte when decoding the stream id?

Yes. Fixing. In fact, it would be better to use a table of functions rather than these big switch statements. That would force unifying all these signatures -- in some case the functions want the "bytes" pointer before the type, in other case after, which causes confusion. But in some case the code needs to check some bits in the value of the frame type ID, so unifying the signatures to point "after the frame ID" would require also passing the frame ID as a parameter. And of course, there are different ID lengths, 1 byte for the most common frames, 2, 4 or even 8 for some other frames.

picoquic/streams.c
```diff
@@ -225,7 +224,7 @@ int picoquic_mark_active_stream_internal(picoquic_cnx_t* cnx,
         if (is_active) {
             /* The call only fails if the stream was closed or reset */
             if (!stream->fin_requested &&
-                (!stream->reset_requested || picoquic_check_sack_list(&stream->sack_list, 0, stream->reliable_size) == 0) &&
+                (!stream->reset_requested || picoquic_reliable_prefix_is_acked(stream)) &&
```
victorstewart: this should be `!picoquic_reliable_prefix_is_acked(stream)`

victorstewart (changes requested):
just that one typo

picoquic/streams.c
huitema: Oops! Old and tired, am I?

suhasHere (approved):

victorstewart (approved):

huitema:
Merging master into this PR before the merging the PR, to flush out potential merge issues.

PR #2120, opened 2026-06-08
Fix memory leaks in mbedtls certificate verifier

cooldazdingo:
## Summary

The mbedtls backend (`picoquic_mbedtls/ptls_mbedtls_sign.c`) leaks the x509 certificate chain head node on a couple of paths.

`ptls_mbedtls_get_certificate_verifier()` allocates the chain head node with `malloc()`, but `mbedtls_x509_crt_free()` only releases the linked-list contents and tail nodes, not the caller-allocated head node itself. As a result the head struct is leaked:

1. In `ptls_mbedtls_get_certificate_verifier()`, when certificate parsing fails (including the `pem_fname == NULL` path, which is hit whenever no root cert file is provided), `chain_head` was released with `mbedtls_x509_crt_free()` but the malloc'd node itself was never `free()`d. The same leak occurred if the verifier allocation failed.
2. In `ptls_mbedtls_dispose_verify_certificate()`, `verifier->trust_ca` (the same malloc'd head node) was released with `mbedtls_x509_crt_free()` but never `free()`d.

## Fix

Add the missing `free()` calls so the head node is released on all paths.

## Notes

The change is limited to memory management; no functional behavior is altered.

### DIFF
--- a/picoquic_mbedtls/ptls_mbedtls_sign.c
+++ b/picoquic_mbedtls/ptls_mbedtls_sign.c
@@ -1325,10 +1325,20 @@ ptls_verify_certificate_t* ptls_mbedtls_get_certificate_verifier(char const* pem
         if (psa_ret == 0) {
             *is_cert_store_not_empty = 1;
             verifier = ptls_mbedssl_init_verify_certificate_complete(chain_head, NULL, NULL, NULL);
+            if (verifier == NULL) {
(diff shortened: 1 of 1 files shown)


huitema (approved):
Thanks for this fix, @cooldazdingo . I have opened issue #2121 to fix the reason why it was not detected earlier. We need to run leak tests using ASAN/UBSAN on the MBEDTLS builds.

PR #2128, opened 2026-06-10
tls credentials rollover

victorstewart:
i rollover containers' certificates at various different lifetime rates, so i need to be able to support rolling over the tls credentials of a live server when there are active QUIC connections by using the latest credentials for new connections but keeping around the old credentials as long as connections are using them


picoquic/quicctx.c
```diff
@@ -1153,16 +1154,15 @@ void picoquic_free(picoquic_quic_t* quic)
         /* Delete the picotls context */
         if (quic->tls_master_ctx != NULL) {
             picoquic_master_tlscontext_free(quic);
-
-            free(quic->tls_master_ctx);
-            quic->tls_master_ctx = NULL;
         }
```
huitema, line 1151: Don't we need to add these free/NULL statements somewhere else?

victorstewart: The free/NULL moved into `picoquic_master_tlscontext_free()`. In the generation-managed case, that helper clears `quic->tls_master_ctx`, releases the active master generation, and the release path frees the underlying `ptls_context_t `when its refcount reaches zero. The helper also keeps the legacy fallback that directly frees and clears `tls_master_ctx` if no generation wrapper exists.

huitema (approved):
Thanks for this contribution. I like the design, the code appears straightforward, and the tests provide good coverage.

PR #2130, opened 2026-06-13
tls session resumption key rotation

victorstewart:
good security hygiene to cycle these keys on some regular basis, and might not want to force expire every existing ticket that has been issued.

RFCs talk about lifetimes of 7 days: https://www.rfc-editor.org/rfc/rfc9325.html#section-3.4-4.2
cloudflare cycles keys every hour: https://blog.cloudflare.com/tls-1-3-overview-and-q-and-a/

### DIFF
--- a/picoquic/picoquic_internal.h
+++ b/picoquic/picoquic_internal.h
@@ -573,11 +573,10 @@ typedef int (*picoquic_performance_log_fn)(picoquic_quic_t* quic, picoquic_cnx_t
 /* QUIC context, defining the tables of connections,
  * open sockets, etc.
  */
-typedef struct st_picoquic_server_state_key_t {
+typedef struct st_picoquic_ticket_key_state_t {
     void* aead_encrypt_ctx;
     void* aead_decrypt_ctx;
-    size_t checksum_length;
-} picoquic_server_state_key_t;
+} picoquic_ticket_key_state_t;
 
 typedef struct st_picoquic_quic_t {
     void* tls_master_ctx;
@@ -683,8 +682,8 @@ typedef struct st_picoquic_quic_t {
     picoquic_connection_id_cb_fn cnx_id_callback_fn;
     void* cnx_id_callback_ctx;
 
-    picoquic_server_state_key_t server_state_key[2];
-    unsigned int server_state_key_active_slot : 1;
+    picoquic_ticket_key_state_t ticket_key_state[2];
+    unsigned int ticket_key_state_active_slot : 1;
     void ** retry_integrity_sign_ctx;
     void ** retry_integrity_verify_ctx;
 
--- a/picoquic/picoquic_internal.h
+++ b/picoquic/picoquic_internal.h
@@ -684,7 +684,7 @@ typedef struct st_picoquic_quic_t {
     void* cnx_id_callback_ctx;
 
     picoquic_server_state_key_t server_state_key[2];
-    uint8_t server_state_key_active_slot;
+    uint8_t server_state_key_active_slot : 1;
     void ** retry_integrity_sign_ctx;
     void ** retry_integrity_verify_ctx;
 
--- a/picoquic/picoquic_internal.h
+++ b/picoquic/picoquic_internal.h
@@ -684,7 +684,7 @@ typedef struct st_picoquic_quic_t {
     void* cnx_id_callback_ctx;
 
     picoquic_server_state_key_t server_state_key[2];
-    uint8_t server_state_key_active_slot : 1;
+    unsigned int server_state_key_active_slot : 1;
     void ** retry_integrity_sign_ctx;
     void ** retry_integrity_verify_ctx;
 
--- a/picoquic/picoquic_internal.h
+++ b/picoquic/picoquic_internal.h
@@ -573,6 +573,12 @@ typedef int (*picoquic_performance_log_fn)(picoquic_quic_t* quic, picoquic_cnx_t
 /* QUIC context, defining the tables of connections,
  * open sockets, etc.
  */
+typedef struct st_picoquic_server_state_key_t {
+    void* aead_encrypt_ctx;
+    void* aead_decrypt_ctx;
+    size_t checksum_length;
+} picoquic_server_state_key_t;
+
 typedef struct st_picoquic_quic_t {
     void* tls_master_ctx;
     picoquic_stream_data_cb_fn default_callback_fn;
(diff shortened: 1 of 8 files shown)


huitema (commented):
Yes, we do need something like that. Couple of questions:

1. Have you considered reserving a bit or two in the ticket to identify the key? That would allow overlap between old and new key while reserving space for only two keys, current and previous.

2. If I remember correctly (travelling, difficult to browse large code on laptop), we use the same key for tickets and tokens. Would need to reserve bits in tokens as well.

3. Maybe not related to this PR, but we need at some point to let developers pass a key *handle* instead of a key value, so the actual key can be kept in whatever safe way OpenSSL or MbedTLS supports.

victorstewart:
oh good idea because it really should be designed such that you never need more than 2 active. that's how we did it with the CID keys. i'll switch the design to that.

i'll do the same for tokens also.

picoquic/picoquic_internal.h
```diff
@@ -677,8 +683,8 @@ typedef struct st_picoquic_quic_t {
     picoquic_connection_id_cb_fn cnx_id_callback_fn;
     void* cnx_id_callback_ctx;
 
-    void* aead_encrypt_ticket_ctx;
-    void* aead_decrypt_ticket_ctx;
+    picoquic_server_state_key_t server_state_key[2];
+    uint8_t server_state_key_active_slot : 1;
```
huitema: This is not portable. The specification should be:

``` suggestion
    unsigned int server_state_key_active_slot : 1;
```

picoquic/tls_api.c
```diff
@@ -1265,14 +1295,14 @@ int picoquic_server_encrypt_ticket_call_back(ptls_encrypt_ticket_t* encrypt_tick
...
 
     if (is_encrypt != 0) {
-        ptls_aead_context_t* aead_enc = (ptls_aead_context_t*)quic->aead_encrypt_ticket_ctx;
+        picoquic_server_state_key_t* key = &quic->server_state_key[quic->server_state_key_active_slot];
         /* Encoding*/
-        if (aead_enc == NULL) {
+        if (!picoquic_server_state_key_is_set(key)) {
             ret = -1;
-        } else if ((ret = ptls_buffer_reserve(dst, 8 + 4 + src.len + aead_enc->algo->tag_size)) == 0) {
+        } else if ((ret = ptls_buffer_reserve(dst, 8 + 4 + src.len + key->checksum_length)) == 0) {
```
huitema: I suppose that these two formulations are equivalent, but the old one is more precise: what we call "checksum" is defined in AEAD as the "tag_size". It is a property of the selected AEAD algorithm.

picoquic/tls_api.c
```diff
@@ -1229,6 +1229,36 @@ int picoquic_client_hello_call_back(ptls_on_client_hello_t* on_hello_cb_ctx,
...
+}
+
+static uint64_t picoquic_server_state_random_sequence(uint8_t key_slot)
+{
+    uint64_t sequence;
+
+    do {
+        sequence = picoquic_public_random_64();
+    } while (PICOQUIC_SERVER_STATE_KEY_SLOT(sequence) != key_slot);
+
```
huitema: That's one way to do it. Another is:

static uint64_t picoquic_server_state_random_sequence(int key_slot) {
return (picoquic_public_random_64()&0x7ffffffffffffff)|(((uint64_t)key_slot&1)<<63);
}

picoquic/quicctx.c
```diff
@@ -1077,14 +1077,8 @@ void picoquic_free(picoquic_quic_t* quic)
...
-        if (quic->aead_encrypt_ticket_ctx != NULL) {
-            picoquic_aead_free(quic->aead_encrypt_ticket_ctx);
-            quic->aead_encrypt_ticket_ctx = NULL;
-        }
-
-        if (quic->aead_decrypt_ticket_ctx != NULL) {
-            picoquic_aead_free(quic->aead_decrypt_ticket_ctx);
-            quic->aead_decrypt_ticket_ctx = NULL;
+        for (int i = 0; i < 2; i++) {
+            picoquic_dispose_server_state_key(&quic->server_state_key[i]);
```
huitema: The name is a bit weird. "server_state" could be anything. I would have preferred "ticket_key_state".

picoquic/tls_api.c
```diff
@@ -2593,10 +2633,23 @@ int picoquic_server_setup_ticket_aead_contexts(picoquic_quic_t* quic,
...
+        }
+        if (ret == 0) {
+            ret = picoquic_set_aead_from_secret(&new_key.aead_decrypt_ctx, cipher, 0, temp_secret, "random label");
+        }
+
+        if (ret != 0) {
+            picoquic_dispose_server_state_key(&new_key);
+        }
+        else {
+            new_key.checksum_length = picoquic_aead_get_checksum_length(new_key.aead_encrypt_ctx);
```
huitema: Not a big deal, but we probably do not need to squirrel away the checksum length. The definition of `picoquic_aead_get_checksum_length` is simple enough, we can just call it when we need it. Better still, we could fix `picoquic_aead_get_checksum_length`, remove the temporary hack, and turn it into a Macro.

huitema (commented):
We are almost there. I fixed an issue that was crashing the compiling on Windows: when specifying a size integer (as in x:1) the base type needs to be "int" or preferably "unsigned int", because a single signed bit has thevalue -1 in complement to 2 logic. using types derived from uint8 does not work on some compilers, because "uint8_t" may be defined as "unsigned int : 8"...

Just a few comments.

picoquic/tls_api.c
```diff
@@ -1229,6 +1229,36 @@ int picoquic_client_hello_call_back(ptls_on_client_hello_t* on_hello_cb_ctx,
...
 
+#define PICOQUIC_TOKEN_TYPE_NEW 0x4000000000000000ull
+#define PICOQUIC_SERVER_STATE_KEY_SLOT(sequence) ((uint8_t)((sequence) >> 63))
+
+static int picoquic_server_state_key_is_set(picoquic_server_state_key_t* key)
+{
+    return key->aead_encrypt_ctx != NULL && key->aead_decrypt_ctx != NULL;
+}
+
+static uint64_t picoquic_server_state_random_sequence(uint8_t key_slot)
```
huitema:
```suggestion
static uint64_t picoquic_server_state_random_sequence(unsigned int key_slot)
```

huitema (approved):
Great. This was indeed a missing piece for server management!

PR #2131, opened 2026-06-15
Using liburing on the receive path

huitema:
The first step in the implementation of liburing is to use it on the receive path, replacing the calls to "recvmsg" and "poll" by io_uring equivalents. Early tests using loopback on ubuntu show an average performance gain of about 20%.

| test | poll | io_uring | gain | 
| ---- | ---- | ---- | ---- |
| 1 | 3.48 | 4.20 | 21% | 
| 2 | 3.51 | 4.27 | 22% |
| 3 | 3.57 | 4.28 | 20% |
| 4 | 3.70 | 4.30 | 16% |
| 5 | 3.72 | 4.44 | 19% | 
| Average | 3.60 | 4.30 | 20% |

These tests were conducted on an old Dell XPS15 laptop running Ubuntu. Each test represents a 10GB transfer using QUIC perf (data from memory, not from the disk). The transfer rates are measured in Gbps. The main explanation for the performance difference is that the "ring" structure in io_uring introduces less overhead when processing asynchronous calls.

The implementation of io_uring is not yet complete. ~The tests such as NAT emulation or Thread operation appear to not pass, and will need to be debugged.~ _(Fixed: needed to update the io_uring ring after the cancellation of pending SQE.)_

We will also need to complete the support of QMUX, and to extend the use of io_uring to the sending path. This might require a rewrite of the "sockloop" code, which was designed for the wait-receive-send sequence typical of a "select" loop.


gmarzot:
@huitema This looks exciting. Any news on plan to merge? I will be trying to activate this in [aiopquic](https://pypi.org/project/aiopquic/) `v0.4.0`

huitema:
Early tests showed 2 potential issues:

- That io_uring (and asynch IO) offloads some processing to a kernel thread. The main thread CPU usage is lower, but that may not be such a win when one consider the whole CPU.
- That we somehow had a high amount of "zero length read" on the wake up pipe, probably a bug of some kind.

We need to at least fix the wakeup bug before merging this. There will be two other tasks to do: support the TCP sockets if the app uses QMux; and, use io_uring on the send path as well.

PR #2135, opened 2026-06-22
Update the logging API and make them generic

huitema:
This work is inspired by the PR #2134 submitted by @gmarzot. That PR demonstrated the need for a generic API to add an application specific logging module, typically integrated with the application log.

It also pushed us to take a deep look at the current support for 3 different logging modes: text, binary and qlog. We know that the text and binary format are largely obsolete now, and that we want to ultimately removed them. They were added over time, and each time we added specialized variables in the QUIC context.

The proposed unification implemented in this PR has the following features:

* In the quic context `picoquic_quic_t`, we replace the existing ad hoc lists of logging parameters by a table of three "unified logging" function pointers and a table of three "void" parameters associated with these functions.
* In the connection context `picoquic_cnx_t` we replace the existing ad hoc lists of logging parameters by a table of three "unified logging" context parameters.
* We introduce a generic registration API that provides the function pointer and the associated parameter for the logging method.
* We update the signature of all unified logging functions to add either a per QUIC context parameter to the functions using a `quic` argument or a per connection logging context parameter for the function using a `cnx` argument.
* The method `log_new_connection` is augmented to pass the per quic context parameter and to set the value of the per connection logging context. If that value is NULL, the connection will not be logged, even if the logging method is registered for the QUIC context.
* We add a `log_flush` method to force flushing if the log file, and replace ad hoc implementations that were used for the text and binary methods. These methods are called in the code after sensitive events like creation or destruction of paths, to avid having only partial logs when issues occur.

The unified log header includes the definition of the registration function:
```
int picoquic_register_log_functions(picoquic_quic_t* quic, picoquic_unified_logging_t* fns, void* params);
```
This function can be used to register an application specific logging method.

We added a method to retrieve the per context parameter for a specific method, so applications can retrieve them whne useful, without having to use non public API, and without having make assumptions about the number of the logging slot in which the method is registered:
```
void* picoquic_get_log_params(picoquic_quic_t* quic, picoquic_unified_logging_t* fns);
```

The existing logging functions have been updated as follow:

* qlog: the registration in the QUIC context records a logging directory. The directory name is passed in the "per QUIC context" parameter. The per connection parameter includes per connection data and the pointer to the log file opened in that directory.
* binary: the registration in the QUIC context records a logging directory. The directory name is passed in the "per QUIC context" parameter. The per connection FILE pointer is passed as a per connection context parameter.
* text log: the registration in the QUIC context opens a log file. The FILE pointer is passed as the "per QUIC context" parameter, and also used as the "per connection" context parameter.

The existing tests have been updated to use the modified API, and are all passing.


gmarzot:
looks great.  Could `picoquic_packet_type_enum` move to picoquic.h, and `st_picoquic_packet_header_t` either move or be treated as an opaque pointer from picoquic_unified_log.h? Without this, `picoquic_register_log_functions` is usable in name only from outside the library — same gap we hit on https://github.com/openmoq/moxygen/pull/256.

gmarzot:
PICOQUIC_MAX_LOG_FUNCTIONS=3 matches the existing text/bin/qlog slots. If an application wants to register a 4th, it would have to first nullify a builtin — is that the intended pattern, or would bumping the cap be reasonable? Happy with either; just want to make sure the lifecycle expectation is documented.

gmarzot (approved):

huitema:
The number of slots should not be a problem, because the slots are only occupied if the application actively set the login mode -- calling picoquic_set_qlog, for example. There is almost no reason for application to set both qlog and textlog, or qlog and binlog.

suhasHere (approved):

Issue #2142, opened 2026-06-27
after install picoquic, compile error


alex-cheung92:
my application include `autoqlog.h`, but  `picoquic_qlog.h` doest exist in install dir 

`In file included from /home/alex/Desktop/dev/yunxiao/lag/src/mp/lag_picoquic_client.c:12:
/home/alex/Desktop/dev/yunxiao/lag/sysroot/include/autoqlog.h:29:10: fatal error: picoquic_qlog.h: No such file or directory
   29 | #include "picoquic_qlog.h"
      |          ^~~~~~~~~~~~~~~~~
compilation terminated.`

here is file list
<img width="362" height="813" alt="Image" src="https://github.com/user-attachments/assets/71343011-58ec-44f5-be68-bf593d8a2130" />

huitema:
You should be able to find the file "./picoquic/picoquic_qlog.h". In fact, you should be able to include that directly in your application, instead of "autoqlog.h". We have changed the implementation of qlog. The old way was to create a binary file, and at the end of the connection to convert that binary file into a qlog file (that's what autoqlog did). We replaced that by directly writing the qlog file, because if the application crashed the old way was leaving a residual binary file in the logging directory, and also because doing binary write followed by convert is in fact less efficient than directly writing the qlog.

Please check whether PR #2143 addresses your issue.

huitema:
The separation between "picoquic-core" and "loglib" looked like a good idea at the time, but it keeps generating issues like this one. It is maybe time to revisit that.

alex-cheung92:
> The separation between "picoquic-core" and "loglib" looked like a good idea at the time, but it keeps generating issues like this one. It is maybe time to revisit that.

That's a good idea.  What's is the best practices of using picoquic log? should I  use `autolog.h` or `picoquic_qlog.h`?

I checkout branch `check-qlog-include` still not works.  Here is logs, the latest three lines says I am using the right branch .


```
make[3]: Leaving directory '/home/alex/Desktop/dev/yunxiao/lag/third_party/picoquic/lag_build'
make[3]: Leaving directory '/home/alex/Desktop/dev/yunxiao/lag/third_party/picoquic/lag_build'
make[3]: Entering directory '/home/alex/Desktop/dev/yunxiao/lag/third_party/picoquic/lag_build'
make[3]: Entering directory '/home/alex/Desktop/dev/yunxiao/lag/third_party/picoquic/lag_build'
[...]

PR #2143, opened 2026-06-27
Add picoquic_qlog.h to core header private

huitema:
This should address issue #2142, making sure that the make file properly installs the header "picoquic_qlog.h"


huitema:
@alex-cheung92 Please check whether this simple fix addresses your issue, or whether we need to work a bit more.

suhasHere (approved):

PR #2147, opened 2026-06-30
Add resuming state to handle careful resume.

huitema:
The previous implementation of careful resume was inefficient, because it was too tightly enmeshed with the "slow start" logic. It is simpler to treat "resuming" as a state by itself, rather than a complication of startup.

The resuming state is entered when the stack sets the seed CWIN, from which we deduce the the seed rate. During that state, pacing rate and CWIN are pegged to the seed values. The state lasts for 2 eras, giving enough time for the rate measurement to stabilize. The eras are expanded if the connection is app limited, to avoid exiting too early. After two eras, or at any time if congestion is detected, the state transitions to recovery.

The performance of the satellite test are significantly improved. They are better than BBR, almost on par with Cubic, and within the 7 second performance target set for that test:

|  average time for simple tests| c4 | bbr | cubic | c4_pushing | this PR |
| --------- | ---:| ---:| ---:| ---:| ---:|
| satellite |  7660817 | 7431946 | 6704246 | 7660783 | 6807146 |

|  top 90% time for simple tests| c4 | bbr | cubic | c4_pushing | This PR|
| --------- | ---:| ---:| ---:| ---:| ---:|
| satellite |  7661569 | 7432276 | 6704247 | 7661405 | 6807183 |

It might be possible to improve a bit more by some further tuning, for example exiting directly to cruising instead of recovery if the nominal rate matches the seed rate. This might be something for further work.

### DIFF
--- a/picoquic/c4.c
+++ b/picoquic/c4.c
@@ -76,24 +76,8 @@
 * - RTT min
  */
 
- /* The probe level is used to set the probing rate during the "push" phase.
+ /* The probe level was used to set the probing rate during the "push" phase.
  *
- * The probe level is set to 1 when entering the initial phase.
- *
- * It is computed by an adaptive mechanism:
- *
- * - increase the level by 1 if the previous push was successful
- * - decrease the level by 1 if the previous push was not successful, down to level 1.
- * - set the level to zero if the push generated heavy packet loss or CE marks.
- *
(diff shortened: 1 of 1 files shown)


suhasHere (approved):
seems reasonable to me. Also I wonder why we have cap on 2 eras ? A comment on that might be handy

Issue #2146, opened 2026-06-29
Possible issue with picoquic_callback_stream_fin with multiple chunks of STREAM in same QUIC packet


lminiero:
Hi all,

I've been trying to debug a weird issue I'm encountering from time to time, but that I've managed to more or less consistently replicate with a few attempts in a row.

The problem is that I have STREAM data coming in in only two chunks, both of which happen to be in the same QUIC packet. In the vast majority of cases, this is not a problem: my callback fires with a `picoquic_callback_stream_data` for the first chunk, and when I'm notified about the second one I see a `picoquic_callback_stream_fin`. Sometimes, though, I get `picoquic_callback_stream_fin` for the very first chunk, even though I get another callback being invoked right after that with the second chunk and, again, `picoquic_callback_stream_fin`. My application can't know more data is coming in, and so the stream is closed in my local resources when I get the first `picoquic_callback_stream_fin`.

I'm still trying to figure out why it only happens sometimes (like once every 14-15 attempts, usually) and not always: looking at QLOG files it looks like the two STREAM frames are always part of the same QUIC packet, but for some reason in some cases that triggers the early FIN callback. Please do let me know if there's more debugging info I can provide: in case it might help, I do have the QLOG files for both working and not working use cases. A snippet of the STREAM data in the non-working case is below:

```
[617049, "transport", "packet_received", { "packet_type": "1RTT", "header": { "packet_size": 311, "packet_number": 6, "dcid": "463a49dfc49484cb" }, "frames": [{ 
    "frame_type": "stream", "id": 6, "offset": 0, "length": 5, "fin": false , "begins_with": "1d00000000"}, { 
    "frame_type": "stream", "id": 6, "offset": 5, "length": 216, "fin": true , "begins_with": "000080d47b227665"}, { 
    "frame_type": "padding"}]}],
```

which shows how `fin` is `true` only on the second STREAM frame, and yet I'm getting it for the first as well in the application callback.

huitema:
@lminiero that's an "interesting" issue. Do you have the whole qlog? Whatever is happening may well depend on the state of the stream, so looking at the previous packets would help.

lminiero:
@huitema I do! You can find it here: https://lminiero.it/69106dab215f8541.server.qlog

[...]

PR #2148, opened 2026-07-01
Add test of singleton packet

huitema:
This is an attempt at reproducing issue #2146 .

The test formats "singleton" packets, which contain an entire unidir stream encoded as one or 2 frames, and that may be repeated or not. For each test, the code verifies the number of callbacks 'picoquic_callback_stream_data' and 'picoquic_callback_stream_fin', and verifies that the received stream length and data matches what was find. We find that:

* if the stream is sent as a single frame, the app gets just one 'picoquic_callback_stream_fin' callback,
* if the stream is sent as a single frame, the app gets just gets a 'picoquic_callback_stream_data' callback followed by a  'picoquic_callback_stream_fin' callback,
* the callbacks are not repeated if the packet is sent twice,
* the received content matches the sent value.

In short, we cannot reproduce issue #2146. @lminiero can you take a look at the test? Maybe we are missing something?

### DIFF
--- a/UnitTest1/unittest1.cpp
+++ b/UnitTest1/unittest1.cpp
@@ -832,6 +832,13 @@ namespace UnitTest1
             Assert::AreEqual(ret, 0);
         }
 
+        TEST_METHOD(stream_singleton)
+        {
+            int ret = stream_singleton_test();
+
+            Assert::AreEqual(ret, 0);
+        }
+
         TEST_METHOD(provide_stream_buffer)
         {
             int ret = provide_stream_buffer_test();
--- a/picoquic_t/picoquic_t.c
+++ b/picoquic_t/picoquic_t.c
@@ -165,6 +165,7 @@ static const picoquic_test_def_t test_table[] = {
     { "vn_compat", vn_compat_test },
     { "transport_param_default", transport_param_default_test },
     { "stream_rank", stream_rank_test },
+    { "stream_singleton", stream_singleton_test },
     { "provide_stream_buffer", provide_stream_buffer_test },
     { "transport_param", transport_param_test },
     { "tls_api_sni", tls_api_sni_test },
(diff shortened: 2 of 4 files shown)


lminiero:
> In short, we cannot reproduce issue #2146. @lminiero can you take a look at the test? Maybe we are missing something?

Thanks @huitema, I'll have a look. I've been trying to figure out the issue, as it may be something that's only triggered under specific circumstances, as in my case that always only happens for that stream ID 6, and only after the same scenario worked perfectly before many times. The scenario, which is MoQ based and where my endpoint is the server, usually is:

1. client and server both create a unidirectional stream to send each other some setup data
2. the client creates a bidirectional stream for a first request (to advertise their presence), which the server acknowledges
3. the server creates a bidirectional stream for a different request (to subscribe to a client's resource), which the client acknowledges
4. the client originates a unidirectional stream to send the server some data from the subscription

More things happen after that, but 4. is where it breaks for me when it does, and I'm trying to figure out if there are some race conditions that can lead to that occasional problem. Step 1, for instance, has client and server both originate those unidirectional streams concurrently: I see that in the log I shared with you, stream ID `7` was chosen by the server for that, when it could have used `3` since it was the first stream of that kind (which is what I see in a working scenario instead). The client then uses stream `2` for its own setup (step 1), and `0` for the request of step 2. For step 3., the server picks `9`, which again is a bit surprising since in another working scenario I see it picking stream `1` (again, there were no other server originated bidirectional streams before that).

It's a bit of a convoluted pattern, so I guess harder to mimic in your test, but since it's where I managed to trigger the issue, I can try building a basic test application that replicates this same behaviour (without it needing to be actual MoQ data, although I have captures of actual traffic so I could do that) and see if I can trigger it that way.

lminiero:
@huitema nevermind, after further debugging I found out it was actually an issue on my side, where a new picoquic connection would be mapped to an old connection in my library (missing cleanup of the mapping, and re-use of a connection pointer on the picoquic side). When that happens, I encounter that problem, which also explains the higher stream IDs being used (the state for those is kept in my structures, and an older one was in use).

Apologies for wasting your time :see_no_evil:  Please feel free to close this PR, as it's not really needed, at least not for my (non-existing issue), unless you think it might still be useful as a test to have.

huitema:
Confirmed that this is not need, see issue #2146

PR #2149, opened 2026-07-04
Make packet loop send batch configurable

rwl4:
Make the packet loop send batch configurable.

Today `PICOQUIC_PACKET_LOOP_SEND_MAX` is hardcoded to 10 packets per loop iteration. This keeps the
default unchanged, but adds two opt-in ways for users with send-heavy workloads to tune it:

- allow compile-time override with `#ifndef PICOQUIC_PACKET_LOOP_SEND_MAX`
- add `send_batch_max` to `picoquic_packet_loop_param_t`, where `0` preserves the existing default

The sockloop.c packet loop uses `send_batch_max` when set. Windows remains covered by the compile-time macro.

Some server workloads have many packets ready to send every loop turn. In those cases, a hardcoded 10-
packet batch can force extra packet-loop iterations and poll/syscall churn.

This PR does not change picoquic’s default behavior. Callers must opt in.

### DIFF
--- a/doc/socket_loop.md
+++ b/doc/socket_loop.md
@@ -85,7 +85,7 @@ The `param` argument contains data to parameterize the packet loop:
   `picoquic_prepare_next_packet_ex`. This is used in debugging,
   to verify that the `UDP_GSO` implementation is functional.
 
-* `send_batch_max`: if specified, the maximum number of packets sent in
+`send_batch_max`: if specified, the maximum number of calls to `sendmsg()` in
   one packet loop iteration. If left to 0, the packet loop uses
   `PICOQUIC_PACKET_LOOP_SEND_MAX`.
 
--- a/doc/socket_loop.md
+++ b/doc/socket_loop.md
@@ -85,6 +85,9 @@ The `param` argument contains data to parameterize the packet loop:
   `picoquic_prepare_next_packet_ex`. This is used in debugging,
   to verify that the `UDP_GSO` implementation is functional.
 
+* `send_batch_max`: if specified, the maximum number of packets sent in
+  one packet loop iteration. If left to 0, the packet loop uses
+  `PICOQUIC_PACKET_LOOP_SEND_MAX`.
 
 In addition, the packet loop exposes a network level callback API, to handle
 network level events that are not directly linked to the QUIC connections.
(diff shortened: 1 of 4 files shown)


doc/socket_loop.md
```diff
@@ -85,6 +85,9 @@ The `param` argument contains data to parameterize the packet loop:
   `picoquic_prepare_next_packet_ex`. This is used in debugging,
   to verify that the `UDP_GSO` implementation is functional.
 
+* `send_batch_max`: if specified, the maximum number of packets sent in
+  one packet loop iteration. If left to 0, the packet loop uses
+  `PICOQUIC_PACKET_LOOP_SEND_MAX`.
```
huitema, line 91: That comment is misleading. What you are defining here is the number of `sendmsg` calls in a packet loop iteration. If the platform supports UDP-GSO (or the Windows equivalent) each call to `sendmsg` may be sending multiple packets, up to a limit 65635 bytes and of course the limit imposed by flow control and congestion control on the selected connection. For a Linux server handling multiple connections, the default value would allow a maximum of 450 packets.

huitema (commented):
Thanks for this submission. I think it will improve the performance in some configurations, but probably not all. As mentioned in the comment, PICOQUIC_PACKET_LOOP_SEND_MAX is a limit on the number of calls to `sendmsg`. It is only a limit on the number of packets if the platform does not support UDP GSO (or equivalent). I suggest explaining that in the comments!

doc/socket_loop.md
```diff
@@ -85,6 +85,9 @@ The `param` argument contains data to parameterize the packet loop:
   `picoquic_prepare_next_packet_ex`. This is used in debugging,
   to verify that the `UDP_GSO` implementation is functional.
 
+* `send_batch_max`: if specified, the maximum number of packets sent in
```
huitema:
~~~suggestion
`send_batch_max`: if specified, the maximum number of calls to `sendmsg()` in
~~~

doc/socket_loop.md
huitema: Corrected by applying suggestion.

huitema (commented):
Thanks again for the PR. Since the only issue was in the documentation, I fixed it myself.

huitema (approved):

PR #2160, opened 2026-07-20
Also accept webtransport if webtransport-h3 expected.

### LINKED ISSUE #2159: H3ZERO_WEBTRANSPORT_H3_PROTOCOL ("webtransport-h3") does not match the :protocol token browsers send, path items using it reject browsers with 404
Summary

h3zero.h defines:

#define H3ZERO_WEBTRANSPORT_H3_PROTOCOL "webtransport-h3"
#define H3ZERO_WEBTRANSPORT_H3_PROTOCOL_OLD "webtransport"

Current release browsers (verified with Firefox 152.0.6, 2026-07; Chrome behaves the same per the draft history) still send :protocol: webtransport in the extended CONNECT. A server that configures a path item with the non-_OLD constant, as both picoquicdemo.c and picoquictest/webtransport_test.c do:

.connect_protocol = H3ZERO_WEBTRANSPORT_H3_PROTOCOL,
.connect_protocol_length = sizeof(H3ZERO_WEBTRANSPORT_H3_PROTOCOL) - 1,

rejects every browser CONNECT, because the check in h3zero_common.c ("Unsupported CONNECT protocol on stream…") requires an exact match and answers 404. Browser-side this surfaces only as a generic WebTransportError: WebTransport connection rejected, which makes it a rough first-contact experience for anyone starting a server from the demo sources. (baton_app.c avoids it by leaving connect_protocol unset, which skips the check entirely — that's also the workaround we used.)

The asymmetry is that picoquic's own client side already negotiates the token adaptively: picowt_connect_protocol_from_settings() sends webtransport-h3 only when the peer advertised the newer settings, falling back to webtransport otherwise. The server-side path-item check has no equivalent flexibility, it's a single exact string.

Observed at commit 7e0765c

Suggested resolutions (any of these would do):

1. Let the server-side check accept either known WebTransport token when the configured connect_protocol is one of them (mirroring the client's adaptive behaviour); or
2. accept a NULL-terminated list / add a dedicated "any webtransport" sentinel for connect_protocol; or
3. at minimum, switch picoquicdemo.c / the test reference sources to H3ZERO_WEBTRANSPORT_H3_PROTOCOL_OLD (or unset) and document that H3ZERO_WEBTRANSPORT_H3_PROTOCOL does not yet interoperate with release browsers.

Related: this became reachable by browsers only after #2158 (the draft-02 SETTINGS fix) — before that, release Firefox closed during SETTINGS negotiation and never reached the CONNECT, so demo-derived servers now hit this 404 where they previously failed earlier.

huitema:
Close #2159

The correct specification of the web transport protocol over HTTP3 is "webtransport-h3", but many implementations use "webtransport". Change the header parsing to also accept "webtransport".

### DIFF
--- a/picohttp/h3zero.h
+++ b/picohttp/h3zero.h
@@ -249,6 +249,7 @@ typedef struct st_h3zero_settings_t {
     unsigned int enable_connect_protocol : 1;
     unsigned int h3_datagram : 1;
     unsigned int settings_received : 1;
+    unsigned int web_transport_legacy : 1;
 } h3zero_settings_t;
 
 extern uint8_t const * h3zero_default_setting_frame;
--- a/picohttp/h3zero_common.c
+++ b/picohttp/h3zero_common.c
@@ -1150,12 +1150,12 @@ int h3zero_check_connect_protocol(const picohttp_server_path_item_t* item, h3zer
 			memcmp(stream_ctx->ps.stream_state.header.protocol,
 				item->connect_protocol, item->connect_protocol_length) != 0)) {
 		size_t old_length = strlen(H3ZERO_WEBTRANSPORT_H3_PROTOCOL_OLD);
-		if (stream_ctx->ps.stream_state.header.protocol_length != old_length ||
+		if (stream_ctx->ps.stream_state.header.protocol_length < old_length ||
 			item->connect_protocol_length < old_length ||
 			memcmp(H3ZERO_WEBTRANSPORT_H3_PROTOCOL_OLD,
-				item->connect_protocol, item->connect_protocol_length) != 0 ||
+				item->connect_protocol, old_length) != 0 ||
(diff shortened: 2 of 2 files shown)


LockstepArcade:
Reading through this I think the fallback won't fire for the reported case: the two memcmps in h3zero_check_connect_protocol use item->connect_protocol_length (15 for "webtransport-h3") rather than old_length (12), so memcmp("webtransport", "webtransport-h3", 15) differs at byte 12 and still returns 404. With old_length in both memcmps (comparing over the OLD token, which is a prefix of "webtransport-h3") it should accept the mismatch.

huitema:
Thanks for the quick review, @LockstepArcade . I think this is fixed now.

suhasHere (approved):

PR #2172, opened 2026-07-27
Add support for SCONE in picoquic

huitema:
This PR adds support for the "Standard Communication with Network Elements Protocol",
[SCONE](https://datatracker.ietf.org/doc/draft-ietf-scone-protocol/) to picoquic, as explained in the new document "scone.md".


suhasHere (commented):
This looks neat. Thanks for adding the support.

suhasHere (approved):

PR #2178, opened 2026-08-12
Fix Header Protection interop failure with mbedtls backend

cooldazdingo:
## Summary

The Header Protection (HP) mask derivation uses `ctr_cipher` (AES-CTR) to compute `mask = AES(hp_key, sample)`. This relies on the assumption that `AES-CTR(key, IV=sample, zeros)` produces the same output as `AES-ECB(key, sample)`.

This assumption holds for OpenSSL (which treats the full 16-byte IV as the counter block), but **fails with mbedtls** which partitions the IV differently. The result is that a picoquic client using the mbedtls backend cannot complete the QUIC handshake with non-picoquic servers (e.g. ngtcp2), because:

1. The HP mask is computed incorrectly (CTR vs ECB mismatch)
2. The receive-direction HP context is created with `is_enc=0`, but AES-ECB decrypt ≠ encrypt, so inbound header deprotection silently fails

## Fix

1. Use `ecb_cipher` instead of `ctr_cipher` in `picoquic_get_header_protection_cipher()` — this directly implements RFC 9001 Section 5.4.3 which defines HP as a single AES-ECB block encrypt of the sample
2. Always pass `is_enc=1` when creating the HP cipher context (HP is always encrypt, regardless of packet direction)
3. Adapt `picoquic_pn_encrypt()` to directly ECB-encrypt the 16-byte sample
4. Return constant 16 from `picoquic_pn_iv_size()` (AES block size, since ECB has no IV)

## Notes

- This change is functionally equivalent for OpenSSL backend (AES-ECB and AES-CTR-with-full-IV produce the same output)
- Fixes interoperability with ngtcp2 server when picoquic uses mbedtls
- Tested: picoquic client (ESP32-C3, mbedtls) ↔ ngtcp2 server (picotls+mbedtls) — handshake completes and 1-RTT application data exchanges successfully

### DIFF
--- a/picoquic/tls_api.c
+++ b/picoquic/tls_api.c
@@ -518,7 +518,7 @@ static ptls_cipher_algorithm_t* picoquic_get_header_protection_cipher(ptls_ciphe
         else
 #endif
         {
-            hp_cipher = cipher->aead->ctr_cipher;
+            hp_cipher = cipher->aead->ecb_cipher;
         }
     }
 
@@ -1536,7 +1536,9 @@ static int picoquic_set_pn_enc_from_secret(void ** v_pn_enc, ptls_cipher_suite_t
     else if ((ret = ptls_hkdf_expand_label(cipher->hash, pnekey,
         hp_cipher->key_size, ptls_iovec_init(secret, cipher->hash->digest_size),
         PICOQUIC_LABEL_HP, ptls_iovec_init(NULL, 0), prefix_label)) == 0) {
-        if ((*v_pn_enc = ptls_cipher_new(hp_cipher, is_enc, pnekey)) == NULL) {
+        /* RFC 9001 Section 5.4.3: Header Protection always uses AES-ECB encrypt,
+         * regardless of the packet direction (send or receive). */
+        if ((*v_pn_enc = ptls_cipher_new(hp_cipher, 1, pnekey)) == NULL) {
             ret = PTLS_ERROR_NO_MEMORY;
         }
     }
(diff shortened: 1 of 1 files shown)


huitema:
@cooldazdingo can you comment on why you closed this PR?

I don't think that we have an actual interop problem between MBedTLS and OpenSSL builds, but if we do have one, I would like to fix it!

cooldazdingo:
Hi @huitema, thanks for asking!

We do have a real interop problem. We're running a picoquic client with the mbedtls backend on ESP32-C3 (NuttX RTOS) against an ngtcp2 server (picotls+mbedtls). The QUIC handshake fails because the server cannot decrypt the client's Initial packets — the Header Protection masks don't match.

The root cause involves two issues in `picoquic_set_pn_enc_from_secret()`:

1. **`is_enc` direction flag**: When creating `pn_dec` (receive path), `is_enc=0` is passed to `ptls_cipher_new()`. For the mbedtls PSA backend, this creates the key with `PSA_KEY_USAGE_DECRYPT` and uses `psa_cipher_decrypt_setup()`. While AES-CTR is mathematically symmetric, the PSA operation state differs between encrypt/decrypt setup on some platforms, leading to incorrect HP masks.

2. **PSA cipher operation lifecycle**: The mbedtls PSA cipher implementation requires `psa_cipher_encrypt_setup()` (via `ptls_cipher_init()`) to be called before each `psa_cipher_update()` (via `ptls_cipher_encrypt()`). Without this reset, `psa_cipher_update` operates on an uninitialized/completed operation and produces wrong output.

I closed PR #2179 (the follow-up to this one) because:
- The initial fix was too broad — switching entirely from `ctr_cipher` to `ecb_cipher` broke ChaCha20 suites and caused assertion failures in the OpenSSL backend
- The minimal `is_enc=1`-only fix passed CI (since picoquic-to-picoquic self-tests work fine with CTR regardless of direction), but didn't fully resolve the issue on our target platform
- I wanted to do more local testing before proposing a proper upstream fix

The interop failure is **not** between MbedTLS and OpenSSL builds of picoquic — it's between picoquic (any backend) and non-picoquic implementations (like ngtcp2) that use AES-ECB directly for HP as specified in RFC 9001 §5.4.3. Picoquic's CTR-based approach works in self-test because both sides produce the same (potentially wrong) mask, so they cancel out.

I'll reopen with a proper fix once we have it fully validated. Would you prefer a separate issue to track this?

cooldazdingo:
Hi @huitema, follow-up after further investigation:

It turns out this is **not a picoquic bug** — it was entirely our own configuration issue. I apologize for the confusion and noise.

**Root cause:** Our ESP32-C3/NuttX mbedtls build had `MBEDTLS_CIPHER_MODE_CTR` disabled (commented out in `mbedtls_config.h`) to save code size. Without CTR mode compiled in, `psa_cipher_encrypt_setup(op, key, PSA_ALG_CTR)` silently returns `PSA_ERROR_INVALID_ARGUMENT` (-134), and since picotls's `ptls_mbedtls_cipher_init` ignores all PSA return values with `(void)` casts, the HP cipher operation produces all-zeros output.

Once we enabled `#define MBEDTLS_CIPHER_MODE_CTR` in our mbedtls config, everything works correctly with **unmodified upstream picoquic code** — including the `is_enc` direction flag, which is fine because AES-CTR encrypt == decrypt on a properly configured mbedtls.

We've verified all four auth modes (no-verify, PSK, one-way cert, mTLS) working against our ngtcp2 test server with the stock picoquic code.

Sorry again for the false alarm. The earlier PR #2179 was also closed for the same reason. No changes needed in picoquic.

PR #2180, opened 2026-08-12
Verify that our buffer changes set via setsockopt() actually succeeded

zi0r:
Under FreeBSD, a setsockopt() request for an increased buffer will return an error if it exceeds kernel limits--add a hint/warning if the call fails and we're under FreeBSD.

Under Linux, the same call will silently succeed, but be clamped at the kernel limit.

This can cause irritating failures that are annoying to troubleshoot.

### DIFF
--- a/picoquic/sockloop.c
+++ b/picoquic/sockloop.c
@@ -447,6 +447,9 @@ int picoquic_packet_loop_open_socket(picoquic_packet_loop_param_t* param,
                 opt_ret = getsockopt(s_ctx->fd, SOL_SOCKET, last_op, (char*)&so_errbuf, &opt_len);
                 DBG_PRINTF("Cannot set %s to %d, err=%d, so_sndbuf=%d (%d)",
                     last_op_name, param->socket_buffer_size, sock_error, so_errbuf, opt_ret);
+#ifdef __FreeBSD__
+                DBG_PRINTF("- If increasing buffer size, verify requested size fits within kern.ipc.maxsockbuf");
+#endif
                 ret = -1;
             }
 #ifdef __linux__
--- a/picoquic/sockloop.c
+++ b/picoquic/sockloop.c
@@ -449,6 +449,33 @@ int picoquic_packet_loop_open_socket(picoquic_packet_loop_param_t* param,
                     last_op_name, param->socket_buffer_size, sock_error, so_errbuf, opt_ret);
                 ret = -1;
             }
+#ifdef __linux__
+            else {
+                /* On Linux, setsockopt(SO_SNDBUF/SO_RCVBUF) will silently succeed
+                 * even when the requested buffer exceeds snet.core.wmem_max or
+                 * net.core.rmem_max, with the kernel silently clamping the value.
+                 *
+                 * Read back the set value to verify what we asked for was what we got.
+                 *
+                 * The requested value is doubled in the kernel for overhead, so account
+                 * for that during our check as well. */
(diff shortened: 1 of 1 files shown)


picoquic/sockloop.c
```diff
@@ -447,8 +447,38 @@ int picoquic_packet_loop_open_socket(picoquic_packet_loop_param_t* param,
...
+                    DBG_PRINTF("SO_SNDBUF silently capped by kernel: requested %d, effective %d (check net.core.wmem_max)",
+                        param->socket_buffer_size, so_verify / 2);
+                }
+
+                opt_len = sizeof(int);
+                if (getsockopt(s_ctx->fd, SOL_SOCKET, SO_RCVBUF, (char*)&so_verify, &opt_len) == 0 &&
+                    so_verify / 2 < param->socket_buffer_size) {
+                    DBG_PRINTF("SO_RCVBUF silently capped by kernel: requested %d, effective %d (check net.core.rmem_max)",
+                        param->socket_buffer_size, so_verify / 2);
+                }
```
huitema, line 479: Why are these 5 lines repeated twice?

huitema (commented):
Thanks for the PR. There is a small issue, see below.

Also, if I understand correctly, the only effect of the PR is to do a debug print. Is that sufficient to debug the issue?

picoquic/sockloop.c
zi0r: Two checks: 1 for the receive buffer and 1 for the send buffer.

The information (about the 2 relevant sysctls) should be sufficient to debug.  I don't have a strong opinion on the log level though, so if you think something outside of debug-level would be more appropriate, I'm all for it.

zi0r: (Note the SO_SNDBUF vs SO_RCVBUF differences in the getsockopt() call as well as the DBG_PRINTF as well as the different sysctl mentioned: net.core.rmem_max /  net.core.wmem_max)

huitema: I was wondering whether the error should surface through the API. The application sets the "params" value, but the code is doing something different. Maybe have a confirmation parameter, "this is what was actually reserved"?

zi0r: One option is that we could implement the FreeBSD way of handling this on the Linux side.  Basically make the listen/bind (or whatever it is) API call fail with an error if it doesn't reserve the requested buffer size.

In this scenario, we would probably need to close the socket we just opened and then maybe set ret = -1 and allow that to get returned to the calling application.

I'm sure Linux kernel-types might have a different opinion, but I feel like succeeding 100% or generating an error and failing is more appropriate than silently doing a third option.

zi0r: One annoying thing about the FreeBSD pass/fail approach is that the application doesn't generate a meaningful error.  When I first ran into this, I just received a (mostly) generic error error saying something to the effect of "listen failed."  I needed to ktrace the application to see the setsockopt() call fail and infer why it was unhappy.

huitema: I don't like the pass/fail option because it is hard for the application to handle that in real time. "Change something and retry the socket" is not the best pattern. Also, it is a bit disruptive. So I was considering adding some parameter to report the allocated size. But thinking it over, the simplest solution is what you did. If someone complains in the future, then we can always revisit the issue.

huitema (approved):
Sorry for overlooking the sendbuf/recbuf difference.

And thanks for fixing this.

PR #2181, opened 2026-08-13
Correct parity for stream IDs

RichLogan:
Hi @huitema, I've been looking into some `STREAM_LIMIT_ERROR` issues we were hitting, and it led me to find something I wanted to get some verification on - I am not a QUIC expert by any means. I made a PR just so the code is easier to reference. 

I'm not sure if the local/peer parity of the stream ID from rank call is backwards when setting the `max_stream_id_unidir_local`, potentially causing an off by one error and rejecting the very last stream that should be accepted.

My understanding is that `max_stream_id_unidir_local` is our limit as sent to the peer, and thus is the max stream ID we will accept from the peer. However, we calculate that passing `cnx->client_mode` to `STREAM_ID_FROM_RANK` (what is our stream ID for that rank), when I think we should be using `!cnx->client_ mode` (what is their stream ID for that rank)? 

The actual check (which is correct) looks like:

```
else if (is_remote && stream_id > (IS_BIDIR_STREAM_ID(stream_id) ? cnx->max_stream_id_bidir_local : cnx->max_stream_id_unidir_local)){
        /* Protocol error, stream ID too high */
        picoquic_connection_error(cnx, PICOQUIC_TRANSPORT_STREAM_LIMIT_ERROR, 0);
    }    
```

i.e compare the incoming stream ID to `max_stream_id_bidir_local`.

I think the result here is that we may reject a valid stream ID at the very top end of the limit. For example, a limit of 512 would mean the acceptable peer initiated unidirectional stream ID `2047` would be rejected because the check would compare it to the largest locally initiated stream ID: `2046`?

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -4713,14 +4713,14 @@ int picoquic_check_max_streams_frame_needs_repeat(picoquic_cnx_t* cnx, const uin
     else {
         if (bytes[0] == picoquic_frame_type_max_streams_bidir) {
             if (max_stream_rank <= cnx->max_stream_id_bidir_rank_acked ||
-                cnx->max_stream_id_bidir_local > STREAM_ID_FROM_RANK(max_stream_rank, cnx->client_mode, 0)) {
+                cnx->max_stream_id_bidir_local > STREAM_ID_FROM_RANK(max_stream_rank, !cnx->client_mode, 0)) {
                 /* Streams bidir already increased or already acked  */
                 *no_need_to_repeat = 1;
             }
         }
         else {
             if (max_stream_rank <= cnx->max_stream_id_unidir_rank_acked ||
-                cnx->max_stream_id_unidir_local > STREAM_ID_FROM_RANK(max_stream_rank, cnx->client_mode, 1)) {
+                cnx->max_stream_id_unidir_local > STREAM_ID_FROM_RANK(max_stream_rank, !cnx->client_mode, 1)) {
                 /* Streams unidir already increased or acked */
                 *no_need_to_repeat = 1;
             }
--- a/picoquic/quicctx.c
+++ b/picoquic/quicctx.c
@@ -3797,10 +3797,10 @@ picoquic_cnx_t* picoquic_create_cnx_internal(picoquic_quic_t* quic,
         /* Initialize local flow control variables to advertised values */
         cnx->maxdata_local = ((uint64_t)cnx->local_parameters.initial_max_data);
         cnx->max_stream_id_bidir_local = STREAM_ID_FROM_RANK(
-            cnx->local_parameters.initial_max_stream_id_bidir, cnx->client_mode, 0);
+            cnx->local_parameters.initial_max_stream_id_bidir, !cnx->client_mode, 0);
         cnx->max_stream_id_bidir_local_computed = STREAM_TYPE_FROM_ID(cnx->max_stream_id_bidir_local);
         cnx->max_stream_id_unidir_local = STREAM_ID_FROM_RANK(
-            cnx->local_parameters.initial_max_stream_id_unidir, cnx->client_mode, 1);
+            cnx->local_parameters.initial_max_stream_id_unidir, !cnx->client_mode, 1);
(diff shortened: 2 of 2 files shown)


huitema:
Thanks for the proposed fix! 

I think the root cause is a bit of confusion in the definition of `STREAM_ID_FROM_RANK(rank, client_mode, is_unidir)`, and also either not enough testing or not enough factorization a complex code. I would like to fix that too.

The macro `STREAM_ID_FROM_RANK(rank, client_mode, is_unidir)` is doing a simple job: compute the stream ID corresponding to a stream. The stream ID are parsed according to RFC 9000 as: 

| Bits | Stream Type
| -- | --
| 0x00 | Client-Initiated, Bidirectional
| 0x01 | Server-Initiated, Bidirectional
| 0x02 | Client-Initiated, Unidirectional
| 0x03 | Server-Initiated, Unidirectional

I think that the root cause of the issue is a confusion between "client_initiated" and "client_mode". The "initiated by" status is independent of the mode of the connection. For example, the stream number 0 should always be treated as "client initiated" and "bi directional", regardless of whether the test happens on the server or on the client. So the first fix ought to be, change the macro to be defined as  `STREAM_ID_FROM_RANK(rank, client_initiated, is_unidir)`, to make it clear that this is not a direct use of `cnx->client_mode`.

This leads to the second issue. You are finding implementation errors in `picoquic_check_max_streams_frame_needs_repeat`
and in `picoquic_create_cnx_internal`. That points to a test gap -- I don't think we have tests verifying that the function  `picoquic_check_max_streams_frame_needs_repeat` handles stream IDs correctly, or that `picoquic_create_cnx_internal` correctly computes the maximum stream ID parameters.

The first idea is to create narrow tests for these two functions, verify that they reproduce the issue that you describe, apply the proposed fix, and verify that the tests now pass. But the test gap itself points to a code issue. We have repeated code at many places that test these parameters. But that's a much more intensive process.

What's your suggestion? Any appetite for writing tests, or is that something that I should do as part of general maintenance?

huitema:
OK. Looking at this a bit further. The root cause is a discrepancy between the code and the standard, because of the way the code was first written. RFC 9000 says "Only streams with a stream ID less than (max_streams * 4 + first_stream_id_of_type) can be opened; see [Table 1](https://www.rfc-editor.org/rfc/rfc9000.html#stream-id-types). Initial limits are set in the transport parameters; see [Section 18.2](https://www.rfc-editor.org/rfc/rfc9000.html#transport-parameter-definitions). Subsequent limits are advertised using MAX_STREAMS frames; see [Section 19.11](https://www.rfc-editor.org/rfc/rfc9000.html#frame-max-streams). Separate limits apply to unidirectional and bidirectional streams."

That means the limits set in transport parameters and in MAX_STREAMS frames are expressed as what we call "rank" in the code. That's a change from the early drafts, in which the limits were expressed as stream IDs (i.e., 4* rank + first). But to keep using stream ids, we need code such as:

~~~
       if (bytes[0] == picoquic_frame_type_max_streams_bidir) {
            if (max_stream_rank <= cnx->max_stream_id_bidir_rank_acked ||
                cnx->max_stream_id_bidir_local > STREAM_ID_FROM_RANK(max_stream_rank, cnx->client_mode, 0)) {
                /* Streams bidir already increased or already acked  */
                *no_need_to_repeat = 1;
            }
        }
~~~

... converting the `max_stream_rank` into a max_stream_id value, with a conversion depending on whether this is a client stream ID or a server stream ID. Suppose now that instead of `max_stream_id_bidir_local` we redefine the parameter as `max_stream_rank_bidir_local`. Then the code would simply be:

~~~
       if (bytes[0] == picoquic_frame_type_max_streams_bidir) {
            if (max_stream_rank <= cnx->max_stream_id_bidir_rank_acked ||
                cnx->max_stream_id_bidir_rank_local > max_stream_rank) {
                /* Streams bidir already increased or already acked  */
                *no_need_to_repeat = 1;
            }
        }
~~~

Simpler code, less error prone. I think that would be the right fix.

How about I do a PR with that change?

huitema:
@RichLogan please take a look at PR #2182. It implements the suggestion to "just simplify the code", which I believe will make future maintenance easier.

RichLogan:
Obsolete by #2182. Thanks for the help @huitema !

PR #2182, opened 2026-08-14
Simplify the code managing stream ID limits

huitema:
This is an alternative to PR #2181, "correct parity for stream ID". The problem underlying that PR is that the code managing stream ID limits has to juggle variations of macros like STREAM_RANK_FROM_ID or STREAM_ID_FROM_RANK, and that the last one had a misleading signature, so we end up with mistakes. The root cause is a discrepancy between the code and the standard, because of the way the code was first written. RFC 9000 says "Only streams with a stream ID less than (max_streams * 4 + first_stream_id_of_type) can be opened; see [Table 1](https://www.rfc-editor.org/rfc/rfc9000.html#stream-id-types). Initial limits are set in the transport parameters; see [Section 18.2](https://www.rfc-editor.org/rfc/rfc9000.html#transport-parameter-definitions). Subsequent limits are advertised using MAX_STREAMS frames; see [Section 19.11](https://www.rfc-editor.org/rfc/rfc9000.html#frame-max-streams). Separate limits apply to unidirectional and bidirectional streams."

That means the limits set in transport parameters and in MAX_STREAMS frames are expressed as what we call "rank" in the code. That's a change from the early drafts, in which the limits were expressed as stream IDs (i.e., 4* rank + first). But to keep using stream ids, which requires adding complex macros and tests. The PR reverses that. It replaces old parameters like "cnx->max_stream_id_bidir_local", expressed as stream ID, by the new parameters like "cnx->max_streams_bidir_local", expressed as number of streams. As a result, the code becomes easier to read, and less error prone.

The bugs targeted by PR #2181 were due to misuse of the "client_mode" argument when computing stream limits using STREAM_ID_FROM_RANK. We fix that macro to make clear that the second argument is a stream ID type (created by client vs. server) and not a connection type (client vs server). We fix the issue by eliminating the use of the macro altogether.

### DIFF
--- a/picoquic/frames.c
+++ b/picoquic/frames.c
@@ -41,7 +41,7 @@ picoquic_stream_head_t* picoquic_create_missing_streams(picoquic_cnx_t* cnx, uin
         /* TODO: not an error if lower than next stream, would be just an old stream. */
         picoquic_connection_error(cnx, PICOQUIC_TRANSPORT_STREAM_LIMIT_ERROR, 0);
     }
-    else if (is_remote && stream_id > (IS_BIDIR_STREAM_ID(stream_id) ? cnx->max_stream_id_bidir_local : cnx->max_stream_id_unidir_local)){
+    else if (is_remote && STREAM_RANK_FROM_ID(stream_id) > (IS_BIDIR_STREAM_ID(stream_id) ? cnx->max_streams_bidir_local : cnx->max_streams_unidir_local)){
         /* Protocol error, stream ID too high */
         picoquic_connection_error(cnx, PICOQUIC_TRANSPORT_STREAM_LIMIT_ERROR, 0);
     } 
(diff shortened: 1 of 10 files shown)


suhasHere (approved):

picoquic/frames.c
```diff
@@ -4638,33 +4638,29 @@ const uint8_t* picoquic_decode_max_streams_frame(picoquic_cnx_t* cnx, const uint
     if ((bytes = picoquic_frames_varint_decode(bytes + 1, bytes_max, &max_stream_rank)) == NULL) {
         picoquic_connection_error(cnx, PICOQUIC_TRANSPORT_FRAME_FORMAT_ERROR, max_streams_frame_type);
     }
+    else if (max_stream_rank >= (1ull << 60)) {
```
RichLogan: I'd like to see someone get a high enough value that this is an issue, but while we're here, should this be `>` not `>=`?

Ref: S4.6

> If a max_streams transport parameter or a MAX_STREAMS frame is received with a value greater than 2^60, this would allow a maximum stream ID that cannot be expressed as a variable-length integer

RichLogan (approved):
Thanks @huitema, this looks good to me! I had considered the possibility altering what uses rank vs. ID would make it simpler, but didn't want to start changing a bunch.

One thing that might be worth adding is a specific max peer unidir stream ID test? I added one that might work here as a PR into this branch: https://github.com/private-octopus/picoquic/pull/2183 That tests the specific issue that led me here, and would fail prior to this fix.

picoquic/frames.c
huitema: You are right. The issue is not so much "someone get a high enough value" as "someone write a nosy compliance test and raises an issue..." The formula for rank to stream is (4*(rank -1)+initial), so a rank 2^60 results in at most 2^62 -1, which is valid.

There was a similar issue in the checking of the transport parameter.

huitema: Fixed.

PR #2183, opened 2026-08-14
Add max limit stream ID test

RichLogan:


huitema (commented):
Thanks for doing this! Just one nit: could you add a declaration of this new test in "unitest1.cpp", so it can be run directly from the Visual Studio UI?

huitema (approved):
I will add the declaration of the test for VS later.

PR #2186, opened 2026-08-18
Declare mlkem variants if available and requested

huitema:
Add support for PQC algorithms based on MLKEM if supported by the crypto provider -- which for now is only some versions of OpenSSL..

The support is described in doc/post-quantum.md.

The testing is minimal. The algorithm itself is tested in picotls, we only verify that it is correctly made available.


huitema:
Damn. It took me 6 attempts to correctly disable the test on Apple platforms. All that because "brew install openssl" does not export the path of the openssl headers. Because Apple, I guess.

huitema:
Please take a look at issue #2187 for a discussion of key exchange selection. We can keep that issue separate, and produce a PR once we agree on what the right solution is.

huitema:
Replace the key exchange test. The first version was essentially trying to predict the result of the configuration, which merely replicates the configuration code. Instead, add a test to set the preferred exchange to x25519mlkem on the client, and verify that the connection succeeds. Also set a similar test for x25519, to enhance the test coverage.

huitema:
The last commits revisited the API `picoquic_set_key_exchange(quic, key_exchange_id)`. Previously, this APi only supported the value 0 and PTLS_GROUP_SECP256R1. It now supports 0 and any key_exchange_id value, as long as there is code available to support the specified key exchange. The implementation also remove the treatment of PTLS_GROUP_SECP256R1 as a special case.

This partially addresses issue #2187, but not completely. We probably need a parallel API that allows setting multiple supported key exchanges. That requires agreeing on a preferred design.

huitema:
I think that we now have a solution to issue #2187 . The PR introduces a new API:

~~~
void picoquic_sort_key_exchange_algorithms(uint16_t* ordered_key_exchange, size_t nb_ordered_key_exchange);
~~~

That API sorts the key exchange algorithms that were discovered according to the preferred order specified in the list `ordered_key_exchange`. The default priority will set the hybrid `x25519mlkem768` first in list, following the IETF recommendation, and setting `x25519` second for best performance. Application developers may want to change that.

In fact, I found that I needed to use that API for our tests. Switching the default from `secp256r1` to `x25519mlkem768` changes some of the timings, because the initial packets are longer. That implies changes in a variety of log files, or changes in the execution times of some tests. We want tests to be consistent across platforms, and thus I called to API in the test setup to set the preference to `[ secp256r1, x25519 ]`. We have individual tests that verify the behavior of `x25519` and `x25519mlkem768` when available.

There are residual differences between platforms based on the list of available algorithms. That list is copied in the "Client Hello" in the Initial packet. Discovering more algorithms changes the timing by a few microseconds. That was enough to fail the `performance log` test, and that's very annoying. I solved the issue by using floating point representation for duration and throughput in the performance log, and limiting the floating point precision to 3 decimals.

picoquic/tls_api.c
```diff
@@ -126,6 +127,28 @@ picoquic_crypto_random_provider_t picoquic_crypto_random_provider_fn = NULL;
...
+    PTLS_GROUP_X25519MLKEM768, /* Preferred hybrid PQC + classic group. */
+    PTLS_GROUP_X25519, /* Preferred classic group */
+    PTLS_GROUP_MLKEM1024, /* CNSA 2.0 compliance */
+    PTLS_GROUP_SECP256R1MLKEM768, /* legacy hybrid PQC + classic group. */
+    PTLS_GROUP_SECP256R1 /* default classic group. */
+#if 0
+    /* rarely used groups, not very useful */
+    ,PTLS_GROUP_SECP384R1 /* CNSA 1.0 compliance */
+    ,PTLS_GROUP_SECP384R1MLKEM1024
+    ,PTLS_GROUP_NAME_SECP384R1MLKEM1024
```
suhasHere:
```suggestion
    ,PTLS_GROUP_SECP384R1MLKEM1024
```

suhasHere (approved):
This is a right place and does address PQ-protect confidentiality, but not authenticity direction and hybrid recommended variants.

huitema:
TODO: verify the list of exchnages and the correct data for mlkem 1024 sha384.

PR #2193, opened 2026-08-28
C4: do not trim or drain if rtt < 1ms

huitema:
The recent changes in C4 implementation added functions to "trim" the nominal rate if it remains constantly over the measured values, and to "drain" the standing queues after the Initial phase, or if the nominal max RTT is decreasing. These functions make C4 less aggressive, reduce the amount of queueing and the RTT values, but generally maintain performance. However, they also had the effect of lowering the performance of C4 in high speed tests over low latency connections such as loopback. C4 used to be on par or slightly better than Cubic; after the changes, it became 24% slower.

Trial | Cubic | C4-current | C4-no-trim
------|-------|------|---------
1 | 2643 | 1983 | 2943
2 | 2678 | 2030 | 3009
3 | 2680 | 2040 | 3020
4 | 2705 | 2046 | 3022
5 | 2723 | 2088 | 3035
Average | 2686 | 2037 | 3006

In the table above, we see the data rates (in Mbit/s) achieved over 5 trials with Cubic, the current version of C4, and the version in this PR. All measurements were done on my laptop,  with software compiled in Windows 64 bit release mode. The change is to not perform the "drain" and "trim" functions if the RTT is lower than 1 ms. In such conditions, the bandwidth is only reduced if congestion is noticed, or once the queues cause the RTT to exceed 1ms. After the changes, C4 is 12% faster than Cubic.

### DIFF
--- a/picoquic/c4.c
+++ b/picoquic/c4.c
@@ -545,13 +545,17 @@ static void c4_exit_initial(picoquic_path_t* path_x, c4_state_t* c4_state)
         /* We assume that any required correction is done prior to calling this */
         uint64_t ssthresh = c4_state->initial_cwnd / 2;
         c4_state->nominal_max_rtt = ssthresh * 1000000 / c4_state->nominal_rate;
+        c4_state->is_drain_required = 1;
         if (c4_state->nominal_max_rtt < C4_MAX_RTT_MIN) {
             c4_state->nominal_max_rtt = C4_MAX_RTT_MIN;
+            c4_state->is_drain_required = 0;
+        }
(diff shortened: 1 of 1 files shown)


suhasHere (approved):

huitema:
Thanks for reviewing, Suhas. That exercise motivates me to look a bit further. Getting rid of the trim improves perf by about 25% in this scenario -- a very constrained scenario, with RTT < 1ms, a condition when time measurements may have a lot of relative imprecision, and data rate measurements as well. Is it possible that the current trim is excessive and slows down other scenarios too much?

PR #2195, opened 2026-09-01
Fix offset calculation in `picoquic_is_pkt_ctx_backlog_empty`

RichLogan:
I was looking through some code and came across this offset calculation that I think might not be theoretically correct (although I don't think this actually causes any incorrect behaviour in practise). 

The max bytes arg to skip_frames was always `p->length - p->offset` when I think it should be negated by incrementing `bytes_offset` (offset += frame_length), like the other skip_frame loops?


huitema (approved):
You are right. Thanks for the fix.

You are correct that this should not matter in practice: the `bytes_maxsize` argument is there for safety, to make sure that the code does not attempt to parse data beyond the end of the packet. That safety is not needed if all frames within the packet are well formed. Since the code is only parsing packets formatted locally, there should never be an issue. Well, almost never, because there is in fact an exception. If we apply fuzzing, the stored packet may have been fuzzed, and in that case the safety will matter!

PR #2196, opened 2026-09-01
Do not use OPENSSL_INCLUDE_DIR when WITH_OPENSSL is OFF

krocard:
`WITH_OPENSSL` guards `find_package(OpenSSL REQUIRED)`, but `picoquic-core` and `picohttp-core` add `${OPENSSL_INCLUDE_DIR}` to their include directories unconditionally.

With `WITH_OPENSSL=OFF` that is usually just empty. It is not when another project in the same build already ran CMake's `FindOpenSSL` and failed — picotls does exactly that — which leaves `OPENSSL_INCLUDE_DIR-NOTFOUND` in the cache and fails the generate step:

```
CMake Error: The following variables are used in this project, but they are
set to NOTFOUND: .../OPENSSL_INCLUDE_DIR
```

Hit cross-compiling for Android with `WITH_OPENSSL=OFF` and picotls-minicrypto, where there is no OpenSSL to find. The error names neither OpenSSL nor the cause, which made it slow to track down.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -386,6 +386,7 @@ endif()
 
 if(WITH_OPENSSL)
     find_package(OpenSSL REQUIRED)
+    set(PICOQUIC_OPENSSL_INCLUDE_DIRS ${OPENSSL_INCLUDE_DIR})
     message(STATUS "root: ${OPENSSL_ROOT_DIR}")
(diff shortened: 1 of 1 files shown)


huitema (commented):
Thanks for the fix! Yes, this is indeed a real issue and a good fix. Sorry for not having that fixed that before.

huitema (approved):

PR #2198, opened 2026-09-02
Move full payload in fuzz

RichLogan:
(On the topic of fuzzing!) Something that came up while I was testing another change with memcheck - I think the memmove count here should be the length of the payload (`length - header_length`), rather than the length of the injected frame (`len`).


huitema (approved):
Thanks for that fix too! That was doing the wrong kind of fuzzing...

PR #2199, opened 2026-09-03
Reliability improvements in the picoquic stack

huitema:
This PR incorporates a number of reliability and security fixes in the stack.

The first commit fixes reliability issues in some HTTP tests on the x86 platforms. We had multiple tests verifying that they could properly write a file, then deleting that file as part of test clean-up. Critically, multiple tests were often using the same file name. The tests were failing on Windows 86. It seems that some kind of background process was keeping the files open -- no idea what process. This caused a race: if the next test tries to write the file before the background process closes it, the text failed. Fixed by ensuring that different tests either write different file names or use different folders.

The second commit closes a potential DoS avenue. Packets that arrive before the endpoint has the key to decrypt them are stashed. That stash was not limited, so external parties could grow the queue to a large size and thus consume the server's memory. The commit limits the number of such packets to 10, i.e., the initial congestion window.

The third commit protects a DoS attack in which the peer does not send packets in sequence, e.g., only sends odd numbered packets. This would cause the list of received packet number ranges to grow potentially very large -- there is no flow control parameter limiting the number of packets sent, and congestion control is a local matter. The solution is to cap the number of PN ranges to 128, and to evict the oldest range if a new range-creating packet is received.

The fourth commit fixes a mistake in the code checking that an incoming initial that is not padded does not contain ack eliciting frames. The previous code had a subtraction resulting in an integer underflow, i.e., a very large "size_t". But it turns out that when adding that value to the beginning of the packet, we obtained a pointer before the current one and the parsing stopped, without negative consequence. So we just fix the error, no need to add a new test.

The fifth commit ensures that the queue of stateless packets such as version negotiation or retry is bounded. It was not, and an adversary could have used that to try overwhelm the server.

The sixth commit rewrites a length check in h3zero_parse_qpack_header_frame. Written as `bytes + n_length > bytes_max` it could overflow if n_length was very large. Instead, write `n_length > (uint64_t)(bytes_max - bytes)`.

The seventh commit fixes the `bytes + length > bytes_max` pattern at various places in qlog_frames.c.

The eighth commit fixes the `bytes + length > bytes_max` pattern at various places in log_writer.c.

The 9th commit fix a similar issue in parsing the length field of datagrams.


huitema:
@victorstewart this PR addresses all the issues flagged in your review, except 3:

- Not fixing the NAT validation issue for reasons explained multiple times.
- Not addressing the "token reuse tree" yet because it is medium-low priority and would require a complex rewrite
- Not addressing the "implicit stream opening issue" yet because applications can protect themselves by limiting the number of stream credits that they give to the peer.

huitema:
There was a small issue in the original fix to the [Prevention of DoS attack on Sack List of Packet Numbers.](https://github.com/private-octopus/picoquic/pull/2199/commits/db118f6a151219b518a8f78f2705785807a2fb10) If the endpoint received a packet with a sequence number lower than all the packet in the sack list, the ejection of the last range could cause the "ack horizon" to regress, and also open a hole after that horizon causing some duplicates to not be detected. This was found during the code review by @victorstewart, and is now corrected.

huitema:
Subsequent review by @victorstewart  found two issues: the fix of the fix for [Prevention of DoS attack on Sack List of Packet Numbers](https://github.com/private-octopus/picoquic/pull/2199/commits/db118f6a151219b518a8f78f2705785807a2fb10) was wrong, allowed for unchecked duplicates; and the fix of the max_rtt_remote issue was incomplete. Both issues have now been fixed.

victorstewart (approved):

PR #2203, opened 2026-09-10
Add optional bind address to the packet loop parameters

mondain:
## Problem

`picoquic_packet_loop_param_t` only carries `local_port` and `local_af`, and `picoquic_bind_to_port` zeroes the sockaddr before `bind()`, so every socket opened by the packet loop (v2/v3, network thread) listens on the wildcard address. An application that needs to serve on a single interface of a multi-homed host has no way to express that.

## Change

- Add `struct sockaddr_storage local_addr` at the end of `picoquic_packet_loop_param_t`. Zero-initialized params behave exactly as before.
- When `local_addr.ss_family` is set, `picoquic_packet_loop_open_sockets` opens a single socket of that family and binds it to that address and to `local_port` (the port inside `local_addr` is ignored). A conflicting non-zero `local_af` is rejected.
- Add `picoquic_bind_to_address()` in picosocks; `picoquic_bind_to_port()` is now a wrapper around it, so existing callers are unchanged.
- The legacy `picoquic_packet_loop_win` path (`picoquic_packet_loop_open_sockets_win`) is untouched.

## Tests

New `sockloop_bind_addr` test:
- opens sockets with `local_addr` set to 127.0.0.1 and ::1 with `local_af` left unspecified, and verifies via `getsockname` that exactly one socket of the right family is created and bound to that address;
- runs a full client/server loop with the server bound to 127.0.0.1 only.

Registered in `picoquic_t`, `picoquictest.h`, and `UnitTest1`. All `sockloop_*` tests pass locally on Linux.

## Docs

`doc/socket_loop.md` and `doc/parallel.md` updated with the new parameter.

### DIFF
--- a/UnitTest1/unittest1.cpp
+++ b/UnitTest1/unittest1.cpp
@@ -181,6 +181,13 @@ namespace UnitTest1
             Assert::AreEqual(ret, 0);
         }
 
+        TEST_METHOD(sockloop_bind_addr)
+        {
+            int ret = sockloop_bind_addr_test();
+
+            Assert::AreEqual(ret, 0);
+        }
+
         TEST_METHOD(sockloop_migration)
         {
             int ret = sockloop_migration_test();
--- a/doc/parallel.md
+++ b/doc/parallel.md
@@ -80,6 +80,7 @@ typedef struct st_picoquic_packet_loop_param_t {
     int simulate_eio;
     size_t send_length_max;
     size_t send_batch_max;
+    struct sockaddr_storage local_addr; /* Optional bind address */
 } picoquic_packet_loop_param_t;
 ```
 
--- a/doc/parallel.md
+++ b/doc/parallel.md
@@ -80,7 +80,7 @@ typedef struct st_picoquic_packet_loop_param_t {
     int simulate_eio;
     size_t send_length_max;
     size_t send_batch_max;
-    struct sockaddr_storage local_addr; /* Optional bind address */
+    struct sockaddr_storage local_addr[PICOQUIC_PACKET_LOOP_LOCAL_ADDR_MAX]; /* Optional bind addresses, one per family */
 } picoquic_packet_loop_param_t;
 ```
 
--- a/doc/socket_loop.md
+++ b/doc/socket_loop.md
@@ -57,6 +57,14 @@ The `param` argument contains data to parameterize the packet loop:
   the local socket. If the value is left to `AF_UNSPEC`, two sockets
   will be created, one for `AF_INET` (IPv4), and one for `AF_INET6` (IPv6).
 
+* `local_addr`: optional local address to which the socket will be bound,
+  as a `sockaddr_storage`. If the address family is left to `AF_UNSPEC`
+  (the default after zero initialization), sockets are bound to the
+  wildcard address and listen on all interfaces. If an address is set,
+  a single socket of that address family is created and bound to that
+  address and to `local_port`; the port inside `local_addr` is ignored.
+  Setting `local_af` to a different family than `local_addr` is an error.
+
 * `dest_if`: the interface identifier that should be associated with the local
   socket, or 0.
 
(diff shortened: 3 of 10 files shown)


huitema (commented):
This looks like a good PR, and a nice functionality. I just have a couple of concerns.

1- The current multipath functionality depends on having a socket capable of sending through multiple interfaces. The default path setup specifies the destination address and lets the local routing function decide which source address and interface is appropriate, but the API also allows the application to set the source address and the interface.

2- In Windows, we have a strong host model and binding to a source address binds to the corresponding interface. In Linux, we have a weak host model, and a packet can be sent from the IP address of one interface through another interface. I have had to help developers debug the consequences.

3- We support dual stack IPv4-IPv6. Should we specify 2 IP addresses, not just one?

4- If we want to restrict an application to use a specific interface, should we not specify the interface ID and enforce it somehow?

5- We have the concept of "extra socket", i.e., allow the socket loop to open two ports, one bound for server operation, and one random for client connections. What kind of restriction shall we place on the extra socket?

mondain:
@huitema I'm addressing some of the points via code; longer response coming shortly. Thanks for the quic review and commentary.

mondain:
1. Multipath and 2. Weak host model - Considered as one issue here; We found a gap in the original PR, which we  addressed in the latest push. The send path hands each path's local address to `picoquic_socks_cmsg_format` as `addr_from`, which becomes `IP_PKTINFO.ipi_spec_dst` / `IPV6_PKTINFO.ipi6_addr`. On Linux the bind restricted what the socket received, but a multipath probe from a different local address would still have gone out with that source, and its replies could never reach the bound socket. On Windows the bound address governs the send, so the two platforms behaved differently. The intent of this parameter is the single-homed service case: "this process speaks on this address", the same thing a `listen` directive gives every other server. Multipath across several local interfaces needs one unbound socket, or one socket per local address, and is out of scope here.

3. Dual stack - Agreed, and done. picoquic already sets `IPV6_V6ONLY` on non-Windows; the generalization is one socket per configured address. `local_addr` is now `struct sockaddr_storage local_addr[PICOQUIC_PACKET_LOOP_LOCAL_ADDR_MAX]` (2 entries, at most one per family). Each entry with a non-zero `ss_family` yields one socket bound to that address and `local_port`; zero-initialized params keep today's behavior; two entries of the same family, or an entry conflicting with a non-zero `local_af`, are refused. A family with no entry gets no socket. The test covers 127.0.0.1 alone, ::1 alone, both together, and both refusal cases.

4. Interface ID - `dest_if` is already in `picoquic_packet_loop_param_t` and documented in `socket_loop.md` as "the interface identifier that should be associated with the local socket", but as far as I can see nothing applies it when the socket is opened; only the per-packet receive-side `dest_if` from pktinfo is used. Enforcing it at open time would be `SO_BINDTODEVICE` on Linux (unprivileged since 5.7), `IP_BOUND_IF` / `IPV6_BOUND_IF` on macOS, and `IP_UNICAST_IF` / `IPV6_UNICAST_IF` on Windows. That is complementary to address binding, not a replacement: operators configure addresses, and on Linux an address bind alone does not stop the kernel routing out of another interface, so the two compose (address for what to listen on, interface to enforce the path). I would keep that as a separate PR since it touches platform-specific options and needs its own tests, but I am willing to do it as a follow-up.

5. Extra socket - The extra socket, and any socket the loop opens later for a new destination family or port, goes through the same `picoquic_packet_loop_open_socket` and so is bound to the same addresses with its own port. When bind addresses are configured, a socket for a family that has no entry is now refused rather than opened on the wildcard, since an unrestricted socket would reintroduce the point-1 problem. I think that is the right restriction: if the application asked to speak only from those addresses, its client connections should too. This is stated in the doc. If there is a use for an unrestricted extra socket, a separate parameter could be added later without changing this behavior.

huitema (approved):
Thanks. The changes address my immediate concerns.

PR #2204, opened 2026-09-11
Fix DBG_PRINTF issue with __VA_ARGS__

zi0r:
This resolves a FreeBSD-specific build failure that I introduced with a previous PR.

DBG_PRINTF needs more than one argument.


huitema (approved):
Thanks for fixing this!

PR #2207, opened 2026-09-13
Do not send IP_SENDSRCADDR on sockets bound to an IPv4 address

mondain:
## Problem

Follow-up to #2203. With `local_addr` configured, every IPv4 send from a bound socket fails on FreeBSD with `EINVAL`.

On platforms without `IP_PKTINFO`, `picoquic_socks_cmsg_format` sets the IPv4 source with `IP_SENDSRCADDR`. FreeBSD refuses that option on a socket bound to a specific address: `udp_output()` passes the address to `in_pcbbind_setup()`, which starts with

```c
laddr.s_addr = *laddrp;          /* the socket's bound address */
if (sin != NULL && laddr.s_addr != INADDR_ANY)
    return (EINVAL);
```

and ip(4) documents that a socket bound to a local address may only pass `INADDR_ANY` with `IP_SENDSRCADDR`. `picoquic_packet_loop_set_send_source()` always filled the path's local address with the bound address, so the control message was added to every packet. Server replies would hit it even without that function, since their source comes from the received destination address.

Reproduced on FreeBSD 14.3: on master, `sockloop_bind_addr` fails and the server qlog shows `Could not send message to AF_to=2, AF_from=2, if=0, ret=-1, err=22` for every send.

IPv6 is not affected: `in6_selectsrc` accepts `IPV6_PKTINFO` with a local address on a bound socket. Linux, macOS and Windows use `IP_PKTINFO` and are not affected. Default configurations (no `local_addr`) are not affected.

## Change

- `picoquic_packet_loop_set_send_source()`: when `IP_PKTINFO` is not defined and the socket is bound to a specific IPv4 address, clear the local address instead of setting it. No control message is added, and the kernel sends from the bound address, which is the intended source. On BSD the bind pins the source, so the weak host concern that motivated the substitution does not apply.
- `sockloop_bind_addr`: expect the cleared IPv4 source on those platforms, and add a full loop run with the server bound to `::1`.
- Header comment and `doc/socket_loop.md` updated.
- New `CITestsFreeBSD` workflow (FreeBSD 14.3 VM via `vmactions/freebsd-vm`) running the same `picoquic_ct` and `picohttp_ct` suites as `CITests`. Nothing in CI built or tested on FreeBSD, which is how this got through.

## Testing

- FreeBSD 14.3 (qemu/KVM): master fails `sockloop_bind_addr` with `err=22`; this branch passes it and the other sockloop tests. With this change and #2208 together, full `picoquic_ct` (574 tests) and `picohttp_ct` (104 tests) pass, 0 build warnings. The workflow commands were run as written in that VM.
- Linux gcc: full `picoquic_ct` (573) and `picohttp_ct` (104) pass, 0 warnings.
- Linux, on this change plus #2208: ASAN/UBSAN full suites with no sanitizer reports; `WITH_SELECT` sockloop tests; `WITH_THREAD_CHECK` full `picoquic_ct`; clang `-Wall -Wextra -Werror -Wpedantic` with no diagnostics in the changed files; `WITH_OPENSSL=OFF` builds, and `minicrypto` passes (`minicrypto_is_last` fails identically on master with my local picotls build, so that one is environmental); embedding test (standalone install and FetchContent re-export) passes.

@zi0r, you build on FreeBSD; could you review this and confirm it works on your setup?


zi0r:
This now works for me under FreeBSD.  Thanks!

picoquic/sockloop.c
```diff
@@ -403,11 +403,20 @@ void picoquic_packet_loop_set_send_source(const picoquic_socket_ctx_t* s_ctx, st
     }
     if (s_ctx->bound_addr.ss_family == AF_INET) {
         struct sockaddr_in* s4 = (struct sockaddr_in*)local_addr;
+#ifndef IP_PKTINFO
+        /* Without IP_PKTINFO (BSD), the source is set with IP_SENDSRCADDR, and
+         * the kernel refuses that option with EINVAL on a socket bound to a
+         * specific address. Leave the source unspecified so that no control
+         * message is added: the kernel then sends from the bound address. */
+        memset(s4, 0, sizeof(struct sockaddr_in));
+        return;
```
huitema: That return statement is not necessary. In general, I avoid doing return (or goto) from the middle of a block, because that might unvoluntarily skip the final clean-up statements.

picoquictest/sockloop_test.c
```diff
@@ -812,6 +812,19 @@ static int sockloop_bind_addr_refused(int local_af, int af0, int af1)
...
+/* Without IP_PKTINFO (BSD), the source of an IPv4 packet sent on a bound
+ * socket is left unspecified, because IP_SENDSRCADDR is refused there. */
+static void sockloop_bind_addr_expected_source(struct sockaddr_storage* expected)
+{
+#ifndef IP_PKTINFO
+    if (expected->ss_family == AF_INET) {
+        memset(expected, 0, sizeof(struct sockaddr_storage));
+    }
+#else
+    (void)expected;
```
huitema: That's ugly! Is there another way? Maybe define a macro bind_addr_expected_source that's a noop if IP_PKTINFO is not defined?

picoquictest/sockloop_test.c
```diff
@@ -852,6 +865,7 @@ static int sockloop_bind_addr_send_source(int af)
         picoquic_store_addr(&local_addr, (struct sockaddr*)&other);
         picoquic_packet_loop_set_send_source(&s_ctx[0], &local_addr);
         (void)sockloop_test_addr_config(&expected, af, 1234);
+        sockloop_bind_addr_expected_source(&expected);
```
huitema: Or maybe bracket *that* statement by the ifndef IP_PKTINFO, so there is no side effect on other platfrms?

huitema (commented):
You are solving a real problem. Having a CI action for FreeBSD is great for reliability. I have some issues on details of the solution. I also have a more general issue: the "GSO" data path is more efficient in software than preparing individual packets one by one. I would like to evolve the socket loop to always prepare packets the same way, i.e., in batches. Then, depending on support for GSO or not, push that batch as a single sendmsg using GSO, or as a series of individual calls. Even in that last case, there are efficiencies to be gained because all these individual calls will have almost the same sendmsg header. 

You don't have do do the "generalize GSO" work in this PR -- but please take a look at the other issues.

Also, please take a look at PR #2205 -- it includes lots of improvements in testing, including testing of sockloop.c

picoquic/sockloop.c
mondain: Agreed, removed. It was the last statement in the IPv4 branch, so nothing changes without it, but I take the point about not skipping clean-up.

picoquictest/sockloop_test.c
mondain: Agreed, the helper is gone. It's now `SOCKLOOP_BIND_ADDR_EXPECTED_SOURCE`, defined once and a no-op when `IP_PKTINFO` is defined; I went with the macro over bracketing each call so the two call sites stay readable.

I also added `sockloop_bind_addr_send_bound`, following the pattern of `sockloop_send_err_unreachable` from #2205: it opens a socket bound to loopback, takes the source from `picoquic_packet_loop_set_send_source`, sends with `picoquic_sendmsg`, and checks that the send succeeds and the receiver sees the bound address. That is the send that returned `EINVAL` on FreeBSD, and it needs no platform conditional.

picoquictest/sockloop_test.c
mondain: Handled with the macro from your other comment: it expands to nothing when `IP_PKTINFO` is defined, so there is no side effect on the other platforms.

The branch is also rebased on master with #2205, so the FreeBSD job now runs the new sockloop tests from that PR too.

mondain:
@huitema The GSO batching change makes sense, but I'd do it as a separate PR; I'm happy to assist

huitema (approved):
Looks good. Thanks for the quick turnaround! And yes, I agree that GSO batching should be tackled in another PR.

PR #2208, opened 2026-09-13
Fix send error handling in picoquic_packet_loop_do_udp_send

mondain:
## Problem

Two defects in the error path of `picoquic_packet_loop_do_udp_send` (sockloop.c), found while reviewing the send path for #2203:

1. **Unreachable errors were ignored.** On an error implying that the destination is unreachable (`EHOSTUNREACH`, `ENETUNREACH`, `EAFNOSUPPORT`, ...), the function called

   ```c
   picoquic_notify_destination_unreachable(last_cnx, current_time,
       (struct sockaddr*)&peer_addr, (struct sockaddr*)&local_addr, ...);
   ```

   but `peer_addr` and `local_addr` are already `struct sockaddr_storage*`. The notification received the address of the pointer variables, so `picoquic_find_path_by_address` compared stack bytes, never matched a path, and no path challenge or demotion happened. The `winsockloop.c` call sites take the address of struct members and are correct.

2. **GSO was never disabled after EIO.** After resending a GSO batch packet by packet, the function set its by-value `send_msg_ptr` argument to `NULL` and logged "UDP GSO was disabled", but the caller's pointer and segment size were unchanged, so the loop kept preparing and sending GSO batches on an interface that had just rejected them.

## Change

- Pass `peer_addr` and `local_addr` directly.
- `send_msg_ptr` becomes `size_t**`. On EIO the function clears the caller's pointer, so `picoquic_prepare_next_packet_ex` stops producing batches, and the segment size it points to, so later sends do not use a stale GSO size. The loop passes `&send_msg_ptr`.
- Declare `picoquic_packet_loop_do_udp_send` in `picoquic_packet_loop.h`. It had no prototype and no other callers, and the test needs it.

## Testing

New `sockloop_send_err` test, registered in `picoquic_t`, `picoquictest.h` and `UnitTest1`. It runs on a connection established in simulation:

- EIO: `simulate_eio` on a 3 x 1440 byte batch; checks that GSO is disabled for the caller.
- Unreachable: sends to an IPv6 address on an IPv4 socket (`EAFNOSUPPORT` on Linux and FreeBSD) using path 0's addresses, and checks that a path challenge is set on that path. The test first probes the error; if a stack reports something that is not an unreachable error, this check is skipped rather than failing.

Reverting either fix, one at a time, makes the test fail. On FreeBSD 14.3 the unreachable check runs (errno 47) rather than being skipped.

- Linux gcc: full `picoquic_ct` (574 tests) and `picohttp_ct` (104 tests) pass, 0 warnings.
- FreeBSD 14.3 (qemu/KVM): `sockloop_send_err` and the sockloop tests pass. With #2207, full `picoquic_ct` and `picohttp_ct` pass.
- Linux, on this change plus #2207: ASAN/UBSAN full suites with no sanitizer reports; `WITH_SELECT`; `WITH_THREAD_CHECK` full `picoquic_ct`; clang `-Wall -Wextra -Werror -Wpedantic` with no diagnostics in the changed files; `WITH_OPENSSL=OFF` builds, and `minicrypto` passes (`minicrypto_is_last` fails identically on master with my local picotls build, so that one is environmental); embedding test (standalone install and FetchContent re-export, which installs the changed header) passes.


huitema (approved):
Thanks for finding and correcting this issue.

The socket loop is harder to test than the protocol code, because of operating system dependencies. This set of test and fixes is a nice complement.

PR #2213, opened 2026-09-15
Fix sockloop_bind_addr on Windows and re-enable send_bound there

### LINKED ISSUE #2210: sockloop_bind_addr flaky on Windows (non-deterministic failures, added by #2207)
`sockloop_bind_addr` (added by #2207) fails non-deterministically on Windows: two full-suite runs of `picoquic_t.exe` on the same build, back to back, failed at two different points inside the test, with two different symptoms.

Environment: Windows 11, MSVC 2022, Debug/x64, local dev build (not yet observed on CI).

**Run 1** — failed immediately in `sockloop_bind_addr_one`, on the very first socket:

```
sockloop.c:483 [picoquic_packet_loop_open_socket]: Cannot set socket (fd=-1, af=2, port = 0, shared: 0)
sockloop.c:642 [picoquic_packet_loop_open_sockets]: Cannot set socket (af=2, port = 0)
sockloop_test.c:947 [sockloop_bind_addr_one]: Expected 1 sockets, got 0
```

`fd=-1` means `WSASocket(AF_INET, SOCK_DGRAM, IPPROTO_UDP, ...)` itself returned `INVALID_SOCKET` — this is before any `setsockopt`/bind call, for a plain loopback (127.0.0.1) IPv4 socket with an ephemeral port. This reproduced when running `sockloop_bind_addr` in isolation too, not just as part of the full suite.

**Run 2** — got past the point where run 1 failed (including the deliberately-invalid-af "refused" sub-cases, `af=200` and `af=23`, which were correctly rejected), then failed later in a different sub-test:

```
sockloop.c:587 [picoquic_packet_loop_open_sockets]: Unsupported local address family af=200
sockloop.c:443 [picoquic_packet_loop_open_socket]: No local address configured for af=23
sockloop.c:591 [picoquic_packet_loop_open_sockets]: Cannot bind af=23 socket to address of af=2
sockloop.c:595 [picoquic_packet_loop_open_sockets]: Duplicate local address for af=2
sockloop_test.c:1223 [sockloop_bind_addr_send_bound]: Expected 64 bytes from bound socket, got 0
```

(The `af=200`/`af=23` lines above look like intentional negative-test output, not the failure itself — the actual failure is the last line, "Expected 64 bytes from bound socket, got 0" in `sockloop_bind_addr_send_bound`.)

Both runs used the exact same binary; nothing else on the machine changed between them as far as I could tell. The fact that the failure point itself moves between runs suggests either:
- something timing/scheduling-sensitive in how the test binds and immediately exercises real loopback UDP sockets on Windows, or
- a genuine bug in the new `local_addr[]`-array socket-opening path that only manifests intermittently.

I have not root-caused this further — flagging it since #2207 was developed/tested against FreeBSD networking issues and may not have been validated on Windows. Given the non-determinism, this risks making CI red on unrelated PRs. Worth considering a temporary Windows-specific skip/exclusion for this test while it's investigated, if it turns out to reproduce on Windows CI too.

mondain:
Fixes #2210. @huitema

The two failures in #2210 have separate causes, both in the test code rather than in the library.

**Run 1: `WSASocket` returns `INVALID_SOCKET` (fd=-1).** The bind_addr sub-tests call `picoquic_packet_loop_open_sockets` directly. Winsock is only started inside the packet loop itself or by `picoquic_open_client_socket`, so `WSASocket` fails whenever no earlier test has started Winsock. This is also why it reproduced when the test ran in isolation. `sockloop_bind_addr_test` now calls `WSA_START` first on Windows.

**Run 2: "Expected 64 bytes from bound socket, got 0".** On Windows, `picoquic_packet_set_windows_socket` posts an overlapped `WSARecvMsg` as soon as a packet loop socket is opened. The receiving socket in `sockloop_bind_addr_send_bound` was a packet loop socket, so that pending receive raced `picoquic_select` for the datagram. The receiver is now a plain socket (`picoquic_open_client_socket` bound to the loopback address), so `picoquic_select` behaves the same on all platforms. The sending socket is still a packet loop socket, which is the path under test.

With both fixed, this removes the `#ifndef _WINDOWS` skip added in #2211.

**Testing:** on Linux, `sockloop_bind_addr` passed 20 of 20 runs and all 16 `sockloop_*` tests pass. I have not been able to run it on Windows, and there is no Windows CI job, so a run on the machine where #2210 reproduced would be appreciated.


huitema (approved):
Thanks for fixing the test!

PR #2217, opened 2026-09-16
Add .gitattributes and extend .gitignore for build and test output

mondain:
Two small repository-hygiene changes, no functional code touched.

### `.gitattributes` (new)

Working on Windows, a checkout picks up CRLF endings and any subsequent commit turns into a whole-file diff — I had ~49k lines of pure line-ending churn across 28 files in my tree before adding this. The rules normalize the common text types so that stops happening.

One wrinkle worth calling out: the ECH config files under `certs/ech/` use a `.txt` extension but hold raw binary — they end in `00 0d 0a` (NUL + CRLF). An explicit `text` attribute overrides git's NUL-byte auto-detection, so `*.txt text` alone silently rewrites them. They are marked `binary` to prevent that.

I verified the rules are inert against the existing tree: `git add --renormalize .` over the full checkout stages no tracked file. So this should not produce a normalization churn commit for anyone.

### `.gitignore`

- `/build-*/` — covers out-of-tree build directories alongside the existing `/build/` and `/coverage-report/` entries.
- `*.log`, `*.qlog`, `*.svg`, `*.csv` — test runs leave these scattered around the tree.

Since the repo tracks reference files matching those patterns (`picoquictest/binlog_ref.log`, `binlog_ref.qlog`, `svglog_ref.svg`, `memlog_test_ref.csv`, `app_msg_overflow_ref.qlog`, `picolog_test_input.log`, `loglib/template.svg`) and new ones get added over time, `picoquictest/` and `loglib/` are negated so a future reference file cannot be silently dropped from a commit.

Happy to drop either half if you would rather not carry it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.gitignore
```diff
@@ -31,3 +32,12 @@ Release
 /coverage-report/
 *.gcda
 *.gcno
+*.log
+*.qlog
```
huitema, line 36: Not sure about that one. We do have .log and .qlog files in the test library -- they are used as references for tests. Will that rule imply that updates won't be accepted?

huitema (commented):
The git attributes part is very nice, but the git ignore part is dangerous. I would definitely not apply the "ignore" in the picoquictest folder, because the presence of qlog or log files, and perhaps csv, is just legit.

I think the bulk of the issue is because we are tempted to run tests directly from the top line in the depot. On Linux, the following would work;

mkdir test_results
cd test_results
../picoquic_ct -S .. -n
../picohttp_ct -S .. -n

Leaving the test results in a test result folder, without polluting the main depot.

.gitignore
mondain: I defer to your judgement, but this was all in response to building on windows and it barfing up a whole bunch of qlog, svg, etc files in the project root.

mondain: Not sure what I may have done to build "wrong" on windows; I didn't use VS Code or Studio.

huitema: On Windows, I run the tests in Visual Studio, setting the current directory to the $Outdir picked by VS, which I can exclude as a whole.

Did you try using git clean?

mondain: I'll have to check that and see why my windows build picked the wrong directory.

PR #2220, opened 2026-09-19
document handling of multipath.

huitema:
Add documentation of the handling of multipath.

This documentation reflects what is implemented. There are known feature requests, such as the ability to declare affinity between datagrams and paths, or the possibility to declare a path as used for redundancy, sending duplicated versions of some streams. We might also want to discuss whether marking path affinity for a standby path moves that path to a "restricted availability" status, in which only streams with affinity are sent. The documentation will be updated if and when these features are implemented.


mondain:
The doc generally looks okay to me, but I found a few things I’d want corrected or clarified before merging.

1. **The path creation retry instructions don’t match the API.** The application-facing function is `picoquic_subscribe_new_path_allowed(cnx, &is_already_allowed)`. It returns an error code and reports immediate readiness through the output argument. The internal `picoquic_check_new_path_allowed` returns zero when allowed, and doesn’t subscribe for notifications. The callback is `picoquic_callback_next_path_allowed`; `picoquic_callback_next_path_ready` doesn’t exist. Following the current instructions could leave an application waiting for a notification it never requested. [Implementation](https://github.com/private-octopus/picoquic/blob/8c6bbf82b6b6d98a413a69e6bf2265e7f6a10161/picoquic/quicctx.c#L2423-L2503)

2. **Please distinguish the draft’s zero-value behavior from what picoquic currently supports.** The draft allows advertising `initial_max_path_id = 0` to negotiate multipath and permit additional paths later. Picoquic omits the parameter when zero and only enables multipath when both values are positive. Someone using zero as a starting limit would disable the capability instead. [Draft §2.1](https://www.ietf.org/archive/id/draft-ietf-quic-multipath-21.html#section-2.1), [implementation](https://github.com/private-octopus/picoquic/blob/8c6bbf82b6b6d98a413a69e6bf2265e7f6a10161/picoquic/transport.c#L284-L303)

3. **The affinity fallback needs a more precise explanation.** The doc implies that stream data can move to another path when its preferred path becomes unavailable. The stream selection code still rejects other paths while the affinity pointer is set. It clears on path deletion or an application change, rather than simply because the path is temporarily unusable. For our media use, that means unsent data could stall despite another healthy path. Please document when fallback actually happens, including clearing affinity with `UINT64_MAX`. [Stream selection](https://github.com/private-octopus/picoquic/blob/8c6bbf82b6b6d98a413a69e6bf2265e7f6a10161/picoquic/streams.c#L1017-L1031), [affinity cleanup](https://github.com/private-octopus/picoquic/blob/8c6bbf82b6b6d98a413a69e6bf2265e7f6a10161/picoquic/quicctx.c#L2021-L2028)

4. **Please separate draft requirements from picoquic’s scheduling policy.** The draft’s available/backup signals express independent peer preferences; they aren’t a mandatory, shared scheduling state. It also explicitly supports concurrent use, while leaving scheduling algorithms to implementations. The path creation section should mention that only clients initiate additional paths under this draft. [Draft §3](https://www.ietf.org/archive/id/draft-ietf-quic-multipath-21.html#section-3), [§3.3](https://www.ietf.org/archive/id/draft-ietf-quic-multipath-21.html#section-3.3)

For our use, I’d also want a short operational caveat: distributing a stream across paths with different delays can increase delivery latency; an idle backup can lose its NAT mapping; and separate congestion controllers don’t guarantee independent capacity when paths share a bottleneck. These are deployment concerns, not new bugs introduced by this documentation. [Draft §§5.3–5.9](https://www.ietf.org/archive/id/draft-ietf-quic-multipath-21.html#section-5.3)

At a minimum, please fix these editorial issues:

- `mot` → `not`; `beaving` → `behaving`
- `folling` → `following`; `congetsion` → `congestion`
- `At any point in type` → `At any point in time`
- Remove the duplicated `local` and correct `if when`.
- Fix the RFC 9000 Markdown link and replace `{{paths-and-tuples}}` with a working anchor.
- Replace the obsolete `picoquic_select_next_path_mp` reference with the current `picoquic_select_next_path_tuple` entry point. [PR document](https://github.com/private-octopus/picoquic/blob/ff122eabacafe7c0d68fa1cdcaeb40142d5d97d4/doc/multipath.md), [current scheduler](https://github.com/private-octopus/picoquic/blob/8c6bbf82b6b6d98a413a69e6bf2265e7f6a10161/picoquic/paths.c#L517)

mondain (changes requested):
See in-line comment

huitema:
@mondain Thanks for an excellent review! I think I have applied your feedback in the document's update.

mondain:
Thanks for the quick turnaround — the substantive fixes (API names, the `initial_max_path_id=0` note, affinity fallback, client-only path creation, and the available/backup "suggestion" framing) all look correct now. A few items from the original review didn't make it into this update though:

**Editorial typos — still present:**
- Line 37: `At any point in type` → `At any point in time`
- Line 45: `mot` → `not`
- Line 46: `beaving` → `behaving`
- Line 198: `folling` → `following`
- Line 375: `congetsion` → `congestion`

**Stale API reference:**
- Lines 287, 291: still reference `picoquic_select_next_path_mp`, which doesn't exist. Should be `picoquic_select_next_path_tuple`.

**Broken links:**
- Line 58: `{{paths-and-tuples}}` is not a working anchor.
- Line 63: `[RFC 9000]{https://www.rfc-editor.org/rfc/rfc9000.html}` uses `{}` instead of `()` — invalid Markdown link syntax.

**Missing operational caveat:**
The deployment note I asked for isn't in this revision — nothing about cross-path latency variance, idle backup paths losing their NAT mapping, or independent congestion controllers not guaranteeing independent capacity when paths share a bottleneck. Could you add a short paragraph covering these (draft §§5.3–5.9 territory)?

Once these are in I think this is ready to merge.

PR #2221, opened 2026-09-21
Reset credit blocked stream

RichLogan:
Hey @huitema, my usual caveat of "I'm not sure if this is right".

However, I've been seeing some cases of hitting `STREAM_LIMIT_ERROR` and corresponding connection close under high stream churn situations. After digging into it I think what's happening here is that I may locally create a stream that is (correctly) blocked due to stream limits, and then shortly afterwards I may RESET that stream and move on. In that case, I don't think picoquic checks for credit when doing RESET, and thus it may open the blocked stream in order to send the RESET, and then get (correctly) rejected by the peer. If we keep the RESET pending until the limit supports it, I think that issue would go away.

The test there can repro the behaviour: given limit of 1, open 1, open 2, RESET 2, (STREAM_LIMIT_ERROR would bite here today), close 1, 2 now actually opens and the RESET now gets delivered. 

My understanding is that this is allowed usage - the app is allowed to RESET a locally initiated stream before any frame has been transmitted, and that the end result would be a stream consisting solely of a RESET_STREAM frame, which itself has to follow the usual rules otherwise?


huitema:
To your question: yes, sending a RESET frame as the first frame on a stream is OK. Per RFC 9000: An endpoint MAY send a RESET_STREAM as the first frame that mentions a stream; this causes the sending part of that stream to open and then immediately transition to the "Reset Sent" state.

Note the "sending part". If the client sends a RESET on a bidirectional stream, the reset only closes the client to server part of the stream, but not the server to client part. The server can still send data. The stream will only be deleted after the server closes its own part of the stream by sending FIN or RESET.

picoquic/streams.c
```diff
@@ -856,7 +867,15 @@ void picoquic_add_output_streams(picoquic_cnx_t* cnx, uint64_t old_limit, uint64
...
-                picoquic_insert_output_stream(cnx, stream);
+                if (stream->reset_requested && !stream->reset_sent) {
+                    if (picoquic_queue_reset_stream_frame(cnx, stream) != 0) {
+                        (void)picoquic_connection_error(cnx, PICOQUIC_TRANSPORT_INTERNAL_ERROR, 0);
+                        break;
+                    }
+                }
+                else {
+                    picoquic_insert_output_stream(cnx, stream);
+                }
```
huitema, line 880: What  happens if the client has queued a reset for a unidir stream?

picoquictest/flow_control_test.c
```diff
@@ -665,6 +665,176 @@ int stream_uni_blocked_test(void)
...
+		ret = -1;
+	}
+
+	if (test_ctx != NULL) {
+		tls_api_delete_ctx(test_ctx);
+	}
+
+	return ret;
+}
+
```
huitema, line 849: Nice test, but could we test both the unidir and the bidir variant?

huitema (commented):
Thanks for the PR! It looks good, but I am concerned about potential confusion between bidir and unidir streams. I am concerned in particular that the code you added in `picoquic_add_output_stream` only executes if the stream `is_bidir`. How hard would it be to have the test execute twice, once for testing unidir behavior and once for testing bidir behavior?

picoquic/streams.c
RichLogan: When the unidir reset is queued it'll take the same reset branch above, because `IS_BIDIR_STREAM_ID(stream_id) == is_bidir` will be...admittedly somewhat non-obviously...true (`0 == 0`) in the unidir case like it would be in the bidir case `(1 == 1)`. So that condition there is not only bidir, it's on matching stream ID with stream type. 

Once `reset_sent` is true, we'll still call `picoquic_insert_output_stream` but internally it bails early on that condition - `picoquic_find_ready_stream_has_data` does `if (stream->reset_sent) has_data = 0`

picoquictest/flow_control_test.c
RichLogan: Added

picoquic/streams.c
```diff
@@ -856,7 +867,15 @@ void picoquic_add_output_streams(picoquic_cnx_t* cnx, uint64_t old_limit, uint64
                 break;
             }
             if (IS_LOCAL_STREAM_ID(stream->stream_id, cnx->client_mode) && IS_BIDIR_STREAM_ID(stream->stream_id) == is_bidir) {
```
huitema, line 871:
```suggestion
            /* The is_bidir argument says whether the increase in max_streams applies to
            * unidir or bidir stream. If it does not apply to the current stream, do nothing. */
            if (IS_LOCAL_STREAM_ID(stream->stream_id, cnx->client_mode) && IS_BIDIR_STREAM_ID(stream->stream_id) == is_bidir) {
```

picoquic/streams.c
huitema: OK. That code line is a bit confusing. I will try to minimize the confusion by adding a comment.

picoquictest/flow_control_test.c
huitema: Thanks!

huitema (approved):
Thanks for the adding the unidir+bidir test!

